PR #153, opened 2016-04-18
Rename .hgignore to .gitignore

mtorpey:
This means that git will ignore any generated files and not try to track them.

We should later make the same change to `stable-2.7`.


james-d-mitchell:
I'm not sure that the format of `.hgignore` and `.gitignore` files is the same. Is renaming this file the correct thing to do?

mtorpey:
I've double-checked, and the syntax is valid for a `.gitignore` file (since we haven't done anything too advanced).  But I did remove an unnecessary line that was specific to mercurial.

PR #156, opened 2016-04-19
Travis setup for stable-2.7 branch

mtorpey:
This should run all tests in `SemigroupsTestAll` and `SemigroupsTestManualExamples`.  Travis tests should be passing at the moment.  They currently work using the `master` branch of GAP, but we should add further checks for `stable-4.8`.


james-d-mitchell:
Awesome!

PR #158, opened 2016-04-20
Run Travis tests with more setups

mtorpey:
Travis now runs the test suite 8 times, i.e. every combination of 3 variables:
 * 64-bit/32-bit compile
 * `master`/`stable-4.8` version of GAP
 * `gcc`/`clang` compiler

The version numbers of all required packages are set once each, in `.travis.yml`, for easy editing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,22 @@ language: cpp
 compiler:
   - gcc
   - clang
+env:
+  - GAP_CONFIGURE_FLAGS="" PKG_CONFIGURE_FLAGS=""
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
@mtorpey are the tests expect to fail? I.e. for the same reasons as before?

mtorpey:
Yep, we're expecting these to fail at least until your `fix-semi-iso` changes get merged into GAP.

However, if you look at the 32-bit builds you'll see that there's a separate error relating to compiled code, which occurs in `SemigroupsTestInstall`:

```
gap: src/converter.cc:89: virtual MatrixOverSemiring* MatrixOverSemiringConverter::convert(Bag, size_t): Assertion `((Int)((*(Bag**)(o))[0])) > 0' failed.
```

This should also be fixed later, if we want Semigroups to work in 32-bit mode.

PR #159, opened 2016-04-20
Implement new maximal subsemigroups ideas

wilfwilson:
I have spent this afternoon rebasing the whole `maximal` branch right from its inception in February. This included about 100 commits and 6 merges from `3.0-dev`.

The main work of this PR is in the most recent commit, which includes the work to implement the new maximal subsemigroups ideas which have arisen from the work between C. Donoven, J. D. Mitchell, and W. Wilson (me).

Included in this PR are various other small changes which I now know to make as separate pull requests in the future. However they are largely inter-dependent so I don't feel I can separate them out right now.

The Travis tests are not currently passing - I shall consult with Michael about this tomorrow.


james-d-mitchell:
@wilfwilson you don't want to merge this right now, correct? 

To me this looks like 7 separate pull requests, which is ok this time. I will write some comments on the code itself.

doc/maximal.xml
```diff
@@ -10,102 +11,189 @@
...
     The maximal subsemigroups of <A>S</A>.
   </Returns>
   <Description>
-    If <A>S</A> is a semigroup, then <C>MaximalSubsemigroups</C> returns
-    a list of the maximal subsemigroups of <A>S</A>.
+    If <A>S</A> is a finite semigroup, then <C>MaximalSubsemigroups</C> returns
+    a list of the non-empty maximal subsemigroups of <A>S</A>.  A <E>maximal
+      subsemigroup</E> of <A>S</A> is a proper subsemigroup of <A>S</A> which is
+    contained in no other proper subsemigroup of <A>S</A>.  The methods used by
+    <C>MaximalSubsemigroups</C> are based on <Cite Key="Graham1968aa"/> and ???.
```
james-d-mitchell: Why is ??? written here?

wilfwilson: @james-d-mitchell Well spotted - I put this in as a placeholder. When our work with Casey is available somewhere, I'll put a reference to it in the documentation along with the GGR reference. For now shall I just remove this?

wilfwilson:
@james-d-mitchell Correct, I don't want it merging just yet - I'll let you know when I do.

The code is 'done', but mostly I want to check with Michael about why Travis is failing. Assuming that his PR is merged soon, I'll rebase on top of that, fix anything I identify (eg the ??? in the documentation) and be done.

You're right that the commits in this PR should be split up into different PR. I wouldn't do something like this in the future (especially since git makes branching so easy).

mtorpey:
Currently, all tests on the `unstable-3.0` branch fail.  This is because they are being tested on the `master` or `stable-4.8` branch of GAP, which doesn't yet include James' `fix-semi-iso` changes.

I could always alter this temporarily - it would be a two-line commit which would only affect the one branch.  What do we reckon?

doc/maximal.xml
wilfwilson: I got rid of the ???

wilfwilson:
@james-d-mitchell I removed the question marks from the documentation. The tests pass fine on my computer so the failed tests are Travis must be failing for the reasons Michael gave.

I am happy for this to be merged if you and Michael are.

james-d-mitchell:
Thanks @wilfwilson, I'll need a bit more time to review this before merging. I'll try to do it early in the week.

doc/maximal.xml
```diff
@@ -10,102 +11,194 @@
...
+          <K>false</K> (or unspecified).</E>
+
+        <P/>
+
+        If <C><A>opts</A>.gens</C> is <K>true</K>, then
+        <C>MaximalSubsemigroups</C> will return a generating set for each
+        maximal subsemigroup that it finds.  If <C><A>opts</A>.gens</C> is
+        <K>false</K>, then the maximal subsemigroups will instead be returned as
+        semigroup objects.
+
```
james-d-mitchell: @wilfwilson the formulation above is a bit clunky, How about `If <C><A>opts</A>.gens</C> is <K>false</K> or unspecified, then the maximal subsemigroups themselves are returned and not just generating sets for these subsemigroups.' or something like that?

doc/maximal.xml
```diff
@@ -10,102 +11,194 @@
...
+        <P/>
+
+        For certain types of semigroup and maximal subsemigroup, it can be more
+        computationally expensive to find a generating set for the maximal
+        subsemigroup than to create the maximal subsemigroup as a semigroup
+        object without knowing a generating set. The default value for this
+        option is <K>false</K>.
```
james-d-mitchell: Again please remove `semigroup object` from this. `It can be more computationally expensive to return the generating sets for the maximal subsemigroups, than to return the maximal subsemigroups themselves.`

doc/maximal.xml
```diff
@@ -10,102 +11,194 @@
...
+      <Item>
+        If <C><A>opts</A>.contain</C> is a subset of <A>S</A>, then
+        <C>MaximalSubsemigroups</C> will search for those maximal subsemigroups
+        of <A>S</A> which contain the set <C><A>opts</A>.contain</C>.
+      </Item>
+
+      <Mark>D</Mark>
+      <Item>
+        For a maximal subsemigroup <M>M</M> of a finite semigroup <A>S</A>,
+        there exists a unique &D;-class which contains <M><A>S</A> \setminus
```
james-d-mitchell: Try to avoid using the `<M>` tags as they don't display well in the text version of the manual. I think it is better to write out what you mean, in this example, `which contains the complement of <C>M</C> in <A>S</A>.`

doc/attributes.xml
```diff
@@ -815,52 +816,66 @@ gap> RZMSConnectedComponents(R);
   <Attr Name="RZMSNormalization" Arg="R"/>
```
james-d-mitchell: In the man section for `RZMSNormalization` you should cite Graham's paper.

doc/maximal.xml
```diff
@@ -10,102 +11,194 @@
...
+        <E>This option is only relevant if <A>S</A> is a Rees (0-)matrix
+          semigroup.</E>
+
+        <P/>
+
+        A maximal subsemigroup of a regular Rees 0-matrix semigroup over a group
+        is one of 6 possible types: 1, it consists of the 0; 2, it consists of
+        everything except the zero; 3, it is a union of rows; 4, it is a union
+        of columns; 5, it is a union of both rows and columns; 6, it intersects
+        every H-class non-trivially. A maximal subsemigroup of a Rees matrix
```
james-d-mitchell: `H-class` should be `&H;-class`, please check there aren't more occurrences!

Also: in this view I can't see if you have defined the types anywhere prior to this. I guess you have, but please double check.

doc/maximal.xml
```diff
@@ -10,102 +11,194 @@
...
+        semigroup is either of type 3, 4, or 6. 
+
+        <P/>
+        
+        Hence if <A>S</A> is a Rees (0-)matrix semigroup and
+        <C><A>opts</A>.types</C> is a subset of <C>[1..6]</C>, then
+        <C>MaximalSubsemigroups</C> will search for those maximal subsemigroups
```
james-d-mitchell: Wrong spaces in `<C>[1..6]</C>` should be `<C>[1 .. 6]</C>`

doc/maximal.xml
```diff
@@ -10,102 +11,194 @@
...
+        of <A>S</A> of the types listed by <C><A>opts</A>.types</C>.
+
+        <P/>
+
+        The default value for this option is <C>[1..6]</C> (i.e. no
+        restriction).
```
james-d-mitchell: Same again

doc/maximal.xml
```diff
@@ -117,14 +210,17 @@ true]]></Example>
     <K>true</K> or <K>false</K>.
   </Returns>
   <Description>
-    If <A>S</A> and <A>T</A> are semigroups, then <C>IsMaximalSubsemigroup</C> 
+    If <A>S</A> and <A>T</A> are semigroups, then <C>IsMaximalSubsemigroup</C>
     returns true if and only if <A>T</A> is a maximal subsemigroup of <A>S</A>.
     <P/>
```
james-d-mitchell: `true` -> `<K>true</K>`

gap/attributes/maximal.gd
```diff
@@ -1,19 +1,16 @@
...
-# FIXME What does this do?
-DeclareOperation("MaximalSubsemigroupsNC", [IsReesMatrixSubsemigroup, IsGroup,
-                 IsList, IsAssociativeElement]);
-# FIXME What does this do?
-DeclareOperation("MaximalSubsemigroupsNC", [IsReesZeroMatrixSubsemigroup,
-                 IsGroup, IsDigraph, IsList, IsList, IsList]);
 DeclareOperation("IsMaximalSubsemigroup", [IsSemigroup, IsSemigroup]);
+DeclareAttribute("MaximalSubsemigroups", IsSemigroup);
+DeclareAttribute("NumberMaximalSubsemigroups", IsSemigroup);
+DeclareOperation("MaximalSubsemigroups", [IsSemigroup, IsRecord]);
```
james-d-mitchell: Shouldn't this be named `NrMaximalSubsemigroups` for consistency? Or would it be better to rename `NrWhatever` to `NumberWhatever`? We should do one or the other but not both.

gap/semigroups/semirms.gi
```diff
@@ -11,6 +12,28 @@
...
+function(R)
+  local P;
+  P := ReesMatrixSemigroupOfFamily(FamilyObj(R.1));
+  if HasIsFinite(P) and IsFinite(P) then
+    return true;
+  fi;
+  TryNextMethod();
+end);
+
 #############################################################################
```
james-d-mitchell: I think these immediate methods for `IsFinite` are superseded by changes I made in the GAP library. Can you double check please?

james-d-mitchell:
This looks really good. @wilfwilson As you can see, I've added some comments directly to the code. Please address these, and rebase. Beware that once you've pushed the rebased version to GitHub, the comments above will disappear (they will still be accessible from the links in any emails you've received about this, but they won't appear on this page any longer. 

Some further comments:

* I suggest removing documentation for the `number` option from `MaximalSubsemigroups`. It is unnecessary since there is the operation `NumberMaximalSubsemigroups`. This will make the doc a bit simpler for `MaximalSubsemigroups`.

* Does this PR introduce `RZMSDigraph` but not delete `RZMSGraph`? If so, then please remove `RZMSGraph` in favour of `RZMSDigraph`. It is better that they not both exist.

gap/tools/display.gi
```diff
@@ -727,7 +727,7 @@ function(S, opts)
 
   # TODO make PartialOrderOfDClasses return a digraph
   rel := OutNeighbours(DigraphReflexiveTransitiveReduction(
-                       Digraph(PartialOrderOfDClasses(S))));
+                       DigraphNC(PartialOrderOfDClasses(S))));
   for i in [1 .. Length(rel)] do
```
james-d-mitchell: Why did you make this change? I think it would be better not to introduce this change. It gives an additional check that the output of `PartialOrderOfDClasses` is correct.

wilfwilson: @james-d-mitchell I'm happy to remove this. I made it simply to speed up `DotString`, since from looking at the code it seemed almost impossible that `PartialOrderOfDClasses` could return something which was not a digraph.

gap/attributes/maximal.gd
wilfwilson: @james-d-mitchell I think `NrMaximalSubsemigroups` is the better name, I'll change it.

gap/semigroups/semirms.gi
wilfwilson: Your GAP pull request installs an `ImmediateMethod` for `IsReesMatrixSubsemigroup`, but not for `IsReesZeroMatrixSubsemigroup` (I have both here).

However that might not be a problem - the example which prompted this commit was a Rees matrix subsemigroup, and I only put the 0-matrix code in pro-actively. When I remove this code (and use your GAP changes) the tests still run, so this code is probably unnecessary.

@james-d-mitchell Can you think of a reason why the `ImmediateMethod` would only be necessary for a Rees matrix subsemigroup and not a Rees 0-matrix subsemigroup?

doc/attributes.xml
wilfwilson: @james-d-mitchell Good suggestion, although I already do.

james-d-mitchell: Argh, I see that it is probably in the line after the last line of this diff, good job @wilfwilson

gap/tools/display.gi
james-d-mitchell: Please remove it, it's just a bit unsafe is all.

gap/semigroups/semirms.gi
james-d-mitchell: There was already an immediate method for `IsFinite` for `IsReesZeroMatrixSubsemigroups` and so it just didn't show up in the diff.

wilfwilson:
Thanks for your comments @james-d-mitchell. I've made most of the changes already and I'll push the updated rebased version at some point today. I just want to update the documentation slightly.

This PR does not introduce the function `RZMSDigraph`. The commit might make it look like I added it, but for whatever reason it has been moved within its file (so there will be a corresponding place from where it has been 'deleted'). I agree that we should have only one such function, but I think it'll be best to deal with that in another PR - `RZMSGraph` is actually used in one or two places, and neither of the functions has documentation or tests. So I think dealing with this properly needs its own PR.

wilfwilson:
@james-d-mitchell I've pushed my changes, I think I've addressed everything you've raised. Handily GitHub hasn't removed the comments on my old (overwritten) commits - they're still on the page, you just have to click to see them.

Issue #164, opened 2016-04-22
Bug in MatrixEntries


wilfwilson:
`MatrixEntries` returns the set of elements contained in the matrix of a Rees (0-)matrix semigroup.

This is not documented, and so the expected behaviour is not specified.

Currently `MatrixEntries` returns `Union(Matrix(R){Columns(R)}{Rows(R)});`. The function `Union` returns a sorted list, however it is not always possible to compare certain elements with `0` (which may be a matrix entry), leading to an error:
```
gap> S := Semigroup(SymmetricInverseMonoid(2));;
gap> id := Identity(S);;
gap> R := ReesZeroMatrixSemigroup(S, [[id], [0]]);;
gap> MatrixEntries(R);
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 1st choice method found for `<' on 2 arguments at /Users/Wilf/GAP/lib/methsel2.g:241 called from
Set( Concatenation( tounite )
 ) at /Users/Wilf/GAP/lib/coll.gi:2788 called from
Union( Matrix( x ){Columns( x )}{Rows( x )}
 ) at /Users/Wilf/GAP/pkg/semigroups/gap/semigroups/semirms.gi:340
```
This bug is present in `stable-2.7` and `unstable-3.0`.

I will submit a pull request to fix this problem in each branch, by calling `Unique` instead of `Union` to remove duplicates without sorting. Another option would be to always place `0` first (if present) and then sort the remaining entries (which should all be comparable, as they come from the same semigroup) and append them to the end of `[0]`.

james-d-mitchell:
@wilfwilson I think the last thing you suggest sounds like the best idea. Sort the non-zero entries, then append them to `[0]`. Can you please amend this PR to do this?

wilfwilson:
@james-d-mitchell I've amended the PRs as requested.

PR #166, opened 2016-04-22
Fix MatrixEntires in stable-2.7 (Issue #164)

wilfwilson:
`MatrixEntries` fails for some Rees 0-matrix semigroups whose
matrices contain `0`, since MatrixEntries removes duplicates and
sorts using `Union`, however `0` is not comparable with all
semigroup elements.

This issue is resolved by using `Unique` to remove duplicates,
but not sorting the list of matrix entries.

Resolves: #164

### DIFF
--- a/gap/reesmat.gi
+++ b/gap/reesmat.gi
@@ -78,7 +78,29 @@ function(R)
 end);
 
 InstallMethod(MatrixEntries, "for a Rees 0-matrix semigroup",
-[IsReesZeroMatrixSemigroup], x -> Union(Matrix(x){Columns(x)}{Rows(x)}));
+[IsReesZeroMatrixSemigroup],
+function(R)
+  local mat, elt, zero, i, j;
+
+  mat := Matrix(R);
+  elt := [];
+  zero := false;
+
+  for i in Rows(R) do
+    for j in Columns(R) do
+      if mat[j][i] = 0 then
+        zero := true;
+      else
+        AddSet(elt, mat[j][i]);
+      fi;
+    od;
+  od;
+
+  if zero then
+    return Concatenation([0], elt);
+  fi;
+  return elt;
+end);
 
 #
 
--- a/tst/reesmat.tst
+++ b/tst/reesmat.tst
@@ -1037,6 +1037,34 @@ true
 gap> UnderlyingSemigroup(T);
 <partial perm group of size 120, rank 5 with 73 generators>
 
+#T# MatrixEntries: Test for Issue #164
+gap> mat := [
+>  [Bipartition([[1, 2, 3, 4, -2, -3], [-1], [-4]]), 0, 0, 0],
+>  [0, Bipartition([[1, 3, -1], [2, 4, -2, -3], [-4]]), 0,
+>   Bipartition([[1, 4, -1], [2, 3], [-2], [-3, -4]])],
+>  [0, 0, Bipartition([[1, 2, 3, -3], [4, -1, -4], [-2]]), 0]];;
+gap> R := ReesZeroMatrixSemigroup(PartitionMonoid(4), mat);;
+gap> MatrixEntries(R);
+[ 0, <bipartition: [ 1, 2, 3, 4, -2, -3 ], [ -1 ], [ -4 ]>, 
+  <bipartition: [ 1, 2, 3, -3 ], [ 4, -1, -4 ], [ -2 ]>, 
+  <bipartition: [ 1, 3, -1 ], [ 2, 4, -2, -3 ], [ -4 ]>, 
+  <bipartition: [ 1, 4, -1 ], [ 2, 3 ], [ -2 ], [ -3, -4 ]> ]
+gap> mat := [
+>  [Bipartition([[1, 2, 4], [3, -1, -2], [-3], [-4]]),
+>   Bipartition([[1, -2, -4], [2, 3, 4, -3], [-1]])],
+>  [Bipartition([[1, 2, 4, -1, -4], [3], [-2, -3]]),
+>   Bipartition([[1, 3, -1], [2, 4, -2, -3], [-4]])],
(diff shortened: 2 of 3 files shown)


james-d-mitchell:
Looks ok, but I think we need to modify the only function calling `MatrixEntries` to remove the assumption that `MatrixEntries` is a set. 

An alternative solution would have been to install a method for `\<` for `0` and any element of a Rees 0-matrix semigroup. 

What do you think @wilfwilson ?

wilfwilson:
@james-d-mitchell  Yes, I see the problem. In `unstable-3.0` there is a call to `MatrixEntries` in the code for calculating automorphism groups of Rees 0-matrix semigroups (although in `stable-2.7` there is only code for Rees matrix semigroups, and RMS's are unaffected by this bug and fix). So this needs to be addressed.

> An alternative solution would have been to install a method for \< for 0 and any element of a Rees 0-matrix semigroup. 

The alternative solution would be to install a method for `\<` for 0 and _any element of any semigroup_, because it is the collection of entries from the matrix which are being sorted, and they are 0 along with elements from some particular semigroup.

I'm happy for `MatrixEntries` to return a sorted list. And I don't mind how we do it. Either the idea from the previous paragraph (although I'm not sure how to implement this properly), or what I suggested in Issue #164: manually place the `0` element first, and then sort the remaining elements as usual. What do you think?

By the way - this `AutomorphismGroup` code seems to assume that the underlying group of the Rees (0-)matrix semigroup is a perm group. Is that a deliberate assumption?

james-d-mitchell:
It would probably be a good idea to have a test where there is more than one non-zero entry, so that it is actually sorted. If you could add that, I'll merge.

wilfwilson:
@james-d-mitchell Agreed - I've now done this to both PRs.

james-d-mitchell:
Looks ok, but wouldn't:
    InstallMethod(MatrixEntries, "for a Rees 0-matrix semigroup",
    [IsReesZeroMatrixSemigroup], 
    function(R)
      local elt;

      elt := Concatenation(Matrix(R){Columns(R)}{Rows(R)});
      Sort(elt,
           function(x, y) 
             if x = 0 or y = 0 then 
               return x = 0; 
             else 
               return x < y;
             fi;
           end);
      return DuplicateFreeList(elt);
    end);

Be a bit more efficient, in both number of lines, and performance?

wilfwilson:
@james-d-mitchell I don't think this is better.

1. This has the same number of lines as the code I wrote.

2. Although the list `elt` is "sorted", it is sorted by your customised function. The special functions in GAP for sorted lists are only for those sorted by `\<`. Indeed the applicable method for `DuplicateFreeList` for `elt` is

```
function ( list )
    local  l, i;
    l := [  ];
    for i  in list  do
        if not i in l  then
            Add( l, i );
        fi;
    od;
    return l;
endfunction( list ) ... end
```
Which is the same function (and hence as much effort) as removing duplicates before sorting. The documentation for `DuplicateFreeList` (aka `Unique`) says that for a list sortable by `\<`, it is far faster to use `Set`. In other words, as far as I can tell, there is no special method for `DuplicateFreeList` for a sorted list - and even if there was it would be for a list sorted by `\<`.

So I would argue that your solution is slower - it runs the same `DuplicateFreeList` method on a list of the same length as mine, but it sorts a larger list first.

I think the correct method is to remove every occurrence of `0` from the original set of matrix entries, place `0` at the start of the 'out' list if present, and then sort (and remove duplicates from) the remaining elements using `Set` and append it to the 'out' list.

wilfwilson:
@james-d-mitchell I've made the changes that I just mentioned.

james-d-mitchell:
@wilfwilson Hmm, the version of the code that I commented on is not the current version of the code, which looks much better. The current version of the code also doesn't related to the comments you've made above.  

My suggestion above is indeed slower than the current version, I just tested it on some random 89 x 10 ^ 4 matrix. 

While there is no special method for `DuplicateFreeList` and a sorted list, there is a special `in` method, and if `list` is sorted in the function you quote above, then so is `l`, and so the cost of looking up `not i in l` is lower than when `list` is unsorted.

Your code could probably be further improved (for example, simply adding `0` to the end of the list rather than concatenating `[0]` and `elt` and by not doing `Concatenation(Matrix(R){Columns(R)}{Rows(R)})` but rather just looping over `Columns(R)` and `Rows(R)` instead), but I doubt it is worth it.

wilfwilson:
@james-d-mitchell  For posterity, the version that you commented on simply did `Unique(Concatenation(Matrix(R){Columns(R)}{Rows(R)}))`, removed the `0` (if present), sorted the remaining elements using `Sort`, and then put the `0` back in if necessary.

i.e. it removed duplicates and then sorted; whereas your suggestion was to sort and then remove duplicates. I argued that your suggestion would be slower - and some tests that I ran on random 250x250 RZMS's over S8 indicated that your code was indeed several times slower. But now the current code is massively faster than both.

Yes you're right that the current version of the code doesn't relate to what I said at the end of my long comment, sorry about that. When I came to implement it I realised that removing every occurrence of `0` and then sorting was worse than what I now do, calling `AddSet` on each of the non-zero entries in turn.

I like the idea of not doing `Concatenation`, but of looping over the columns and rows. So I'll make that change. But also I like the idea of `0` being the first entry (if present), so I'll leave that as it is.

I'm not sure that your comments about `DuplicateFreeList` are correct. While there is certainly a special method for `in` for a sorted list and an element, while the list `l` in the quoted function is indeed (mathematically) sorted, I don't see how there would be any way for GAP to know this. e.g.
```
l := [  ];
for i  in list  do # <list> is sorted
  if not i in l  then
    Add( l, i );
  fi;
od;
```
I don't think GAP is clever enough to realise that `l` must also be sorted, and hence be able to apply its special methods.

wilfwilson:
@james-d-mitchell I made the change so that the matrix is dealt with directly, rather than concatenating the matrix into a new list.

PR #168, opened 2016-04-29
Travis: test more configurations with stable-2.7

mtorpey:
This brings the Travis tests for `stable-2.7` in line with the tests for `unstable-3.0`.  It now runs all the tests with the `stable-4.8` and `master` branches of GAP, and it does the compilation with both gcc and clang.

For interest's sake, I also included a configuration for **GAP and its packages compiled in 32-bit mode**.  Since Semigroups 2.7 isn't compiled I assumed it would make no difference, but it looks like there are two tests that fail, which seem to be related to **orb**:

```
reading /home/travis/build/mtorpey/gap/pkg/semigroups/tst/freeband.tst . . .
########> Diff in /home/travis/build/mtorpey/gap/pkg/semigroups/tst/freeband.t\
st:185
# Input is:
HTAdd(ht, new, true);
# Expected output:
33159
# But found:
35014
########
########> Diff in /home/travis/build/mtorpey/gap/pkg/semigroups/tst/freeband.t\
st:190
# Input is:
ht;
# Expected output:
<tree hash table len=100003 used=159 colls=1 accs=160>
# But found:
<tree hash table len=100003 used=159 colls=0 accs=160>
########
Semigroups package: freeband.tst
elapsed time: 1114ms
```

Perhaps these are harmless, but I'm not sure.

We should either:
 * Merge this pull request and then fix these tests somehow.
 * Remove the 32-bit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,19 +2,33 @@ language: cpp
 compiler:
   - gcc
   - clang
+env:
+  global:
+    - SEMIGROUPSPLUSPLUS_BR=0.1
+    - DIGRAPHS_BR=0.5.1
+    - IO=io-4.4.5
+    - GAPDOC=GAPDoc-1.5.1
+    - ORB=orb-4.7.5
+    - GENSS=genss-1.6.3
+    - GRAPE=grape4r7
+  matrix:
+    - GAP_BRANCH=master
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
@mtorpey I think this is harmless, can you update the tests please?

PR #169, opened 2016-05-03
Added ham-examples (WIP)

n-ham:
I've been unable to get the profiling package configured (despite installing and running autoconf), consequently I haven't tested the tests that I've added to Semigroups/tst/examples.tst using code-coverage-test.py. However they are hopefully straight forward, and I'm reasonably confident the rest is good to go.

### DIFF
--- a/doc/chap4.xml
+++ b/doc/chap4.xml
@@ -127,7 +127,41 @@
     <C>(x^*)^*=x, xx^*x=x, x^*xx^*=x^*, (xy)^*=y^*x^*.</C>
     <P/>
   </Alt>
-  In this way, the partition monoid is a <E>regular *-semigroup</E>.
+  In this way, the partition monoid is a <E>regular *-semigroup</E>. <P/>
+
+  A bipartition &x; of degree &n; is called <E>planar</E> if there does not exist
+  distinct blocks <M>A, B \in</M> &x;, along with <M>a, a' \in A</M> and
+  <M>b, b' \in B</M> such that <M>a &lt; a'</M>, <M>b &lt; b'</M> and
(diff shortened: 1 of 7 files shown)


doc/chap4.xml
```diff
@@ -127,7 +127,41 @@
     <C>(x^*)^*=x, xx^*x=x, x^*xx^*=x^*, (xy)^*=y^*x^*.</C>
     <P/>
   </Alt>
-  In this way, the partition monoid is a <E>regular *-semigroup</E>.
+  In this way, the partition monoid is a <E>regular *-semigroup</E>. <P/>
+
+  A bipartition &x; of degree &n; is called <E>planar</E> if there does not exist
```
james-d-mitchell: Shouldn't this be `If there do not `?

james-d-mitchell:
@n-ham I've looked this over and it looks pretty good! From my point of view, there are only one or two minor things to change, and then I can release a version of the package (2.8) with these changes. 

If you are happy with what you have (i.e. that this isn't WIP anymore), then I'll just make the minor changes as part of the release process. The changes are just things like whitespace, and you missed some tests of the text in error messages. It's probably easier for me to do these than to communicate them and have you do them, but we can do that too if you want. 

What do you reckon?

n-ham:
@james-d-mitchell I'm happy for you to make any changes yourself with your own best and/or preferred judgement (including agreement with switching 'does' to 'do' where noted the other day). It will be handy to identify where I'm making these whitespace issues, I am trying to be more on the ball about that kind of thing! However I should be able to look through the history to see what you changed anyway..

PR #172, opened 2016-05-05
Introduce IsMonogenic[Inverse]Monoid and related changes

wilfwilson:
Primarily, this PR introduces the property `IsMonogenicMonoid`, whose code is very similar to the property `IsMonogenicSemigroup`. The property is documented and all new code is covered by tests.

In addition:

- `IsMonogenicSemigroup` itself is updated to add an additional verification step (involving D-classes) before the more expensive final step.
- Attributes `MinimalSemigroupGeneratingSet`, `MinimalMonoidGeneratingSet`... etc are declared, although no methods are installed and these are not documented. Their primary purpose is to allow the storing of the generator of a monogenic semigroup, and the attribute is not currently intended to be user-facing.
- Code has been updated in several places so that any time a semigroup/monoid finds out that it is monogenic, its `Minimal(Semigroup/Monoid/etc)GeneratingSet` is set to contain that single generator.
- The `Size` method for a monogenic transformation semigroup has been updated to use this minimal generating set and its filter rank increased so that it is actually executed, and a similar method has been introduced for a monogenic transformation _monoid_.


wilfwilson:
For some reason the commits are not listed in the correct order on this page. The correct order is shown [https://github.com/wilfwilson/Semigroups/commits/is-monogenic-monoid](on my GitHub page).

james-d-mitchell:
@wilfwilson I'm not sure what you are referring to in your last comment, and the link in your comment is broken.

doc/properties.xml
```diff
@@ -622,6 +622,32 @@ true]]></Example>
...
+    <Prop Name = "IsMonogenicMonoid" Arg = "S"/>
+    <Returns><K>true</K> or <K>false</K>.</Returns>
+   <Description>
+     <C>IsMonogenicMonoid</C> returns <K>true</K> if the monoid
+     <A>S</A> is a monogenic monoid and it returns <K>false</K> if
+     it is not. <P/>
+
+     A monoid is <E>monogenic</E> if it is generated as a monoid
+     by a single element.  See also <Ref Prop = "IsMonogenicSemigroup"/>
+     and <Ref Attr = "IndexPeriodOfTransformation" BookName = "ref"/>.
```
james-d-mitchell: Please remove the reference to `IndexPeriodOfTransformation`, it might be hard to understand the relevance of this reference.

gap/attributes/properties.gi
```diff
@@ -775,13 +782,23 @@ function(S)
     return false;
   fi;
 
+  if HasGreensDClasses(S) then
+    i := NrDClasses(S);
+    if not IsTrivial(I) then
+      i := i - 1;
+    fi;
+    if not i = Number(GreensDClasses(S), IsTrivial) then
+      return false;
```
james-d-mitchell: `if i <> Number ...`

james-d-mitchell:
This looks pretty good. Some comments are inline, here are some more:

1. Please add `IsMonogenicInverseMonoid` too.
2. Please document the attributes `MinimalXGeneratingSet`, it is likely that someone will notice that this exists and then think it is a bug that it doesn't work. Better to write in the doc that this is the currently intended behaviour.

wilfwilson:
Sorry (again) @james-d-mitchell - I did the markdown wrong in my second comment (I switched the URL and the link text of the link by accident) but I've corrected it now. It was just getting a bit confused by GitHub - immediately below the first main comment of this PR is the list of commits of this PR:

screen shot 2016-05-06 at 11 48 55

However they're in completely the wrong order! I was confused by this. However on the (now correct) link, you can view the commits in the correct order, since it might be confusing for you (or any other reviewer) to look at commits in the wrong order.

doc/properties.xml
wilfwilson: @james-d-mitchell Yes I agree. I copy-and-pasted (and then edited) this from `IsMonogenicSemigroup` and should've been more careful. The documentation for `IsMonogenicInverseSemigroup` also contains this reference so I'll remove it from there too.

wilfwilson:
@james-d-mitchell I believe I've addressed all of the points which you raised.

doc/properties.xml
```diff
@@ -569,7 +569,9 @@ true]]></Example>
      is monogenic and it returns <K>false</K> if it is not. <P/>
 
      A semigroup is <E>monogenic</E> if it is generated by a single element.
-     See also <Ref Prop = "IsMonogenicInverseSemigroup"/> and <Ref
+     See also <Ref Prop = "IsMonogenicMonoid"/>,
+     <Ref Prop = "IsMonogenicInverseSemigroup"/>,
+     <Ref Prop = "IsMonogenicInverseMonoid"/>, and <Ref
        Attr = "IndexPeriodOfTransformation" BookName = "ref"/>.
 <Example><![CDATA[
```
james-d-mitchell: There's still an `IndexPeriodOfTransformation` here, please remove.

james-d-mitchell:
@wilfwilson Please remove the one remaining `IndexPeriodOfTransformation` and I'll merge it.

wilfwilson:
@james-d-mitchell I've removed the reference.

PR #173, opened 2016-05-17
Make IsUniversalSemigroupCongruence a property

mtorpey:
For technical reasons, ever since Rees 0-matrix semigroup congruences by linked triple were implemented, `IsUniversalSemigroupCongruence` has been a **category** which contains only objects created explicitly by the function `UniversalSemigroupCongruence(S)`.  Therefore we had the situation where a congruence (say by generating pairs or by linked triple in an RMS) might turn out to be universal, but would not satisfy this filter.

`IsUniversalSemigroupCongruence` has now been changed to a **property**, which can be tested, and will return `true` iff a congruence has precisely one congruence class.  Methods intended for universal congruences will then be selected if possible, reducing work done.

### DIFF
--- a/gap/congruences/conguniv.gi
+++ b/gap/congruences/conguniv.gi
@@ -18,11 +18,48 @@ function(S)
   local fam, cong;
   fam := GeneralMappingsFamily(ElementsFamily(FamilyObj(S)),
                                ElementsFamily(FamilyObj(S)));
-  cong := Objectify(NewType(fam, IsUniversalSemigroupCongruence), rec());
+  cong := Objectify(NewType(fam,
+                            IsSemigroupCongruence and IsAttributeStoringRep),
+                    rec());
   SetSource(cong, S);
   SetRange(cong, S);
+  SetIsUniversalSemigroupCongruence(cong, true);
   return cong;
 end);
+
+#
+
+InstallMethod(IsUniversalSemigroupCongruence,
+"for a semigroup congruence",
+[IsSemigroupCongruence],
+function(cong)
+  return NrEquivalenceClasses(cong) = 1;
+end);
+
+#
+
+InstallImmediateMethod(IsUniversalSemigroupCongruence,
+IsSemigroupCongruence and HasNrEquivalenceClasses, 0,
+function(cong)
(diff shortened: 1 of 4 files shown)


gap/congruences/conguniv.gi
```diff
@@ -18,14 +18,43 @@ function(S)
...
+end);
+
+#
+
+InstallMethod(IsUniversalSemigroupCongruence,
+"for a semigroup congruence",
+[IsRMSCongruenceByLinkedTriple],
+function(cong)
+  return cong!.n = UnderlyingSemigroup(Range(cong)) and
+         Length(cong!.colBlocks) = 1 and
```
wilfwilson: If I was being picky, I'd recommend having the semigroup equality check last, since it is the most expensive equality check here.

gap/congruences/conguniv.gi
```diff
@@ -18,14 +18,43 @@ function(S)
...
+                    rec());
   SetSource(cong, S);
   SetRange(cong, S);
+  SetIsUniversalSemigroupCongruence(cong, true);
   return cong;
 end);
 
 #
 
+InstallMethod(IsUniversalSemigroupCongruence,
```
wilfwilson: This could also be installed as an immediate method, given the additional filter `and HasNrEquivalenceClasses`

wilfwilson:
@mtorpey This seems like a sensible change to make. One of your commits says "Complete code coverage in conguniv.tst", which I would take to mean that running `conguni.tst` achieves (almost) 100% code coverage in `conguni.gi`, however I only achieve 48% when running this test file. This should be improved now in this PR if the missing coverage is related to this PR, or otherwise soon.

wilfwilson:
Sorry, as you rightly pointed out, the code coverage is 95% rather than 48% (I was reading the figure for the the similarly-named `conginv.gi`) - although again as you point out, this 95% could still be improved.

mtorpey:
I've modified the branch to implement all Wilf's suggested changes.  `conguniv.tst` now tests 100% of `conguniv.gi` - I was testing all the `cong*.tst` files at once, and that's why I thought it was 100% before.  Thanks for spotting those things.

PR #175, opened 2016-05-20
Travis: run tests in GAP branch fix-semi-iso

mtorpey:
Currently, the Travis tests in the `unstable-3.0` branch all fail, due to us relying on GAP changes in James' `fix-semi-iso` branch which has not yet been merged into the official GAP repo.  In the long-term we should only test against official GAP versions, but for the moment it's difficult to tell whether new changes we make are causing new problems without any effective automated testing.

This pull request adds another configuration to Travis, which *additionally* runs the tests under `fix-semi-iso`.  Tests under the other branches still fail, but if you want to see whether new changes are alright, you can look at the Travis build page, and you're looking for **two green ticks** to indicate that Semigroups at least works with the altered GAP.

`semipbr.tst` was still failing due to a lack of helper functions at the top - it seems that it was passing only if executed after certain other test files which installed the functions.  I've added the functions to `semipbr.tst`, and made sure they're always unbound at the end of a file so that this problem doesn't go undetected again.

There was also an (I assume) harmless diff in `attributes.xml`, which I've fixed.

I expect this pull request to fail Travis tests (X), but pass 2 out of 12 on the actual build page.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,12 @@ env:
     - GENSS=genss-1.6.3
     - GRAPE=grape4r7
   matrix:
-    - GAP_BRANCH=master
-    - GAP_BRANCH=stable-4.8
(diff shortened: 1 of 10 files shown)


james-d-mitchell:
@mtorpey this seems like a good idea, thanks for the fixes in the test files too.

PR #181, opened 2016-05-26
Travis: add log-folding (2.8)

mtorpey:
The full log produced by Travis CI is verbose and unwieldy. If the build fails it can take a long time to figure out what's going wrong.

This pull request just adds some extra commands to the build and test scripts which "fold" the Travis output, i.e. collapse certain sections so that they can be expanded by a click. A successful test should display on just one or two screens, and a failing test should be much easier to diagnose.

Have a look at the Travis output for this PR to see how it looks in action.


wilfwilson:
What a nice idea

Issue #174, opened 2016-05-19
Semigroups fails to compile in gcc 5.3.1


mtorpey:
In a recently-updated version of `gcc` (version 5.3.1) the Semigroups package fails to compile, giving the following output after `./autogen.sh`, `./configure`, and `make`:

```
[17:01:22] ~/gap/pkg/semigroups $ make
  CXX      src/semigroups_la-semigroups.lo
  CXX      src/semigroups_la-boolean.lo
  CXX      src/semigroups_la-types.lo
  CXX      src/semigroups_la-converter.lo
  CXX      src/semigroups_la-interface.lo
  CXX      src/semigroups_la-data.lo
  CXX      src/semigroups_la-bipart.lo
src/bipart.cc: In constructor ‘NrIdempotentsFinder::NrIdempotentsFinder(Bag, Bag, Bag, unsigned int, Bag)’:
src/bipart.cc:1149:23: error: ‘pow’ is not a member of ‘std’
         total_load += std::pow(_scc.back().size(), 2);
                       ^
src/bipart.cc:1149:23: note: suggested alternative:
In file included from /usr/include/features.h:361:0,
                 from /usr/include/stdlib.h:24,
                 from /home/mtorpey/gap/src/system.h:32,
                 from /home/mtorpey/gap/src/compiled.h:17,
                 from src/bipart.h:11,
                 from src/bipart.cc:10:
/usr/include/x86_64-linux-gnu/bits/mathcalls.h:153:1: note:   ‘pow’
 __MATHCALL_VEC (pow,, (_Mdouble_ __x, _Mdouble_ __y));
 ^
Makefile:594: recipe for target 'src/semigroups_la-bipart.lo' failed
make: *** [src/semigroups_la-bipart.lo] Error 1
[17:01:38] ~/gap/pkg/semigroups $ gcc --version
gcc (Debian 5.3.1-17) 5.3.1 20160429
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

[17:01:50] ~/gap/pkg/semigroups $ g++ --version
g++ (Debian 5.3.1-17) 5.3.1 20160429
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
[...]

PR #182, opened 2016-05-26
Resolve Issue #174

mtorpey:
Issue #174 is a problem with c++11.  `std::pow` is used, but in the latest version of GCC `pow` is no longer in `std`.  This is only used in one place in the whole of Semigroups, and it is easily replaced with a straight multiplication.

I've checked on my computer, and this compiles successfully in GCC 5.3.1


james-d-mitchell:
Please don't merge this, I already fixed this issue as part of more extensive changes, which are not yet ready to be pushed.

mtorpey:
Understood!  I'll close this pull request for now.

PR #187, opened 2016-06-14
UFData: add SaveWorkspace and LoadWorkspace support

mtorpey:
This should make it possible to save `UFData` objects and reload them with `SaveWorkspace` and `LoadWorkspace`.

I haven't been able to test it properly, since I get a segfault any time I try to use `SaveWorkspace` (even before these changes).

### DIFF
--- a/src/gap.cc
+++ b/src/gap.cc
@@ -18,6 +18,12 @@
 
 #include "semigroups++/semigroups.h"
 
+#if !defined(SIZEOF_VOID_P)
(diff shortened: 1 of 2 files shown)


mtorpey:
The second commit makes this safe on 32-bit architectures (hopefully!)

james-d-mitchell:
@mtorpey please squash the commits

mtorpey:
I've tested this in the way you suggested, and it works.  Ready for merging.

PR #195, opened 2016-07-18
Fix for hashing RZMS elements over a pc group

flsmith:
This fixes a bug where hashing RZMS elements over pc groups would fail (not segfault as the commit message says).


james-d-mitchell:
This looks pretty good. Can you please reword the commit message to correctly describe the issue? I will make another inline comment too.

gap/main/setup.gi
```diff
@@ -12,8 +12,11 @@ SEMIGROUPS.HashFunctionRZMSE := function(x, data, func)
   if x![1] = 0 then
     return 1;
   fi;
-
-  return (x![1] + x![3] + func(x![2], data)) mod data + 1;
+  if IsNBitsPcWordRep(x![2]) then
+    return (x![1] + x![3] + func(x![2], data)) mod data[2] + 1;
+  else
```
james-d-mitchell: It would probably be better to do 

    elif IsPerm(x![2]) then 
       return ...
    else 
       ErrorNoReturn("Semigroups: SEMIGROUPS.HashFunctionRZMSE: error,\n",
                                 " this shouldn't happen");
    fi;

This way it will give a proper error message if this crops up again in the future.

gap/main/setup.gi
```diff
@@ -632,21 +635,23 @@ InstallMethod(FakeOne, "for a Rees 0-matrix semigroup element collection",
...
   fi;
   func := function(x, hashlen)
-    return SEMIGROUPS.HashFunctionRZMSE(x, hashlen, under);
+    return SEMIGROUPS.HashFunctionRZMSE(x, data, under);
   end;
 
   return rec(func := func,
-             data := hashlen);
+             data := data);
 end);
```
james-d-mitchell: Should probably check that `x` is a perm or pc group element here too, and return an error if it is not.

flsmith:
I've updated the message and pushed a new commit, which I think uses a more robust way of fixing the problem.

james-d-mitchell:
Looks good, but this needs some tests, and for the last commit (June 30, at the top of the list) to disappear.

flsmith:
Broken, will rerequest.

Issue #191, opened 2016-06-22
The GAP test file tst/testinstall/semicong.tst runs forever


james-d-mitchell:
Line 42 of the GAP test file `tst/testinstall/semicong.tst` runs forever in `unstable-3.0`. This is because it tries the method:

```
InstallMethod(Size,
Concatenation("for a ", _record.info_string, "congruence class"),
[_IsXCongruenceClass],
function(class)
  local p, tab;
  if not IsFinite(Parent(class)) then
    TryNextMethod();
  fi;
  p := Position(GenericSemigroupData(Parent(class)), Representative(class));
  tab := AsLookupTable(EquivalenceClassRelation(class));
  return Number(tab, n -> n = tab[p]);
end);
```

in `congpairs.gi` and the `IsFinite(Parent(class))` runs forever. The semigroup where this congruence class is defined is infinite, but does not know that it is finite. So, `IsFinite` is trying to enumerate the semigroup, which will never finish.

The library method finds the size of the congruence class in a different way (and instantaneously).

james-d-mitchell:
Just to say that this is still an issue with the newer Todd-Coxeter code.

wilfwilson:
Some discussion about the details of the test which currently fails, for anyone interested.

The semigroup created in this test is the fp semigroup `S`:
```
S = <x, y, z | xy = yx, xz = zx, yz = zy  # commutativity relations
                x^2 = x, y^3 = y,         # x and y are torsion
                xz = x, yz = y>           # z is an identity for x and y
```
The first two lines of relations give us that `S` is first constructed as a direct product of the trivial group, the cyclic group of order 2, and the natural numbers, and then the final line says it's a quotient of that direct product by those '`z` is an identity for `x` and `y`' relations.

Since no power of `z` appears as the left/right-hand side of a relation, it follows that `z` has infinite order in `S`, and so `S` is infinite.

A congruence `c` on `S` is created by the single generating pair `[x, y]`.

Lines 41-43:
```
gap> ec := EquivalenceClassOfElement(c, z);;
gap> Size(ec);
1
```
The generating pair is `[x, y]`. Thus we can only have something equivalent to `z` if we can multiply either `x` or `y` by something to reach `z`. But by looking at the relations of `S`, `z` never appears on its own as one side of a relation. Thus it is only equivalent to itself.

Lines 44-48:
```
gap> ec := EquivalenceClassOfElement(c, y);;
gap> z in ec;
false
gap> Size(ec);
5
```
[...]

PR #196, opened 2016-07-21
Fix issue #191

mtorpey:
This fixes Issue #191, [The GAP test file tst/testinstall/semicong.tst runs forever](https://github.com/gap-packages/Semigroups/issues/191), by checking whether a given semigroup is *known to be* finite.  This way, we don't get stuck on an infinite run trying to determine whether the `cong-pairs` algorithm is appropriate; we just go straight to the fallback method (which may well be replaced by our Todd-Coxeter algorithm in the future).

### DIFF
--- a/gap/congruences/congpairs.gi
+++ b/gap/congruences/congpairs.gi
@@ -761,7 +761,7 @@ function(_record)
     cong := EquivalenceClassRelation(class);
     S := Range(cong);
 
-    if not IsFinite(S) then
+    if not (HasIsFinite(S) and IsFinite(S)) then
       TryNextMethod();
     fi;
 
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
While this does fix the immediate issue, I'd prefer to understand how the library code actually performs this calculation. I don't think either of our approaches in the Semigroups package (the one in this PR or the Todd-Coxeter) would ever terminate on this example. 

It would be better to have a solution where our code also terminates on this example. It is probably not worth trying to do this with the old code, but only with the new Todd-Coxeter code.

mtorpey:
This is all getting replaced now anyway, so I'm closing this PR.

PR #199, opened 2016-07-26
SaveWorkspace: add tests to Travis

mtorpey:
This pull request adds a new test system to the automatic Travis tests, to allow objects to be saved to a workspace which is then reloaded.  There is a new `tst/workspaces` directory with two new files, `save-workspace.tst` and `load-workspace.tst`, which create objects and then verify that the objects exist after loading.

`travis-test.sh` now has an extra section where it:
 * creates a new session
 * runs `save-workspace.tst`
 * saves the workspace to a file
 * ends the session
 * reloads the workspace from the file
 * runs `load-workspace.tst`
 * deletes the workspace file

I have included tests for the UFData object as an example.  New tests should be added for any other C/C++ objects we have created, and this can be done easily by adding new test blocks to those files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,4 @@ src/pkgconfig.h
 src/semigroups++
 src/stamp-h1
 tags
+tst/workspaces/test-output.w
--- a/scripts/travis-test.sh
+++ b/scripts/travis-test.sh
@@ -1,8 +1,18 @@
 # If a command fails, exit this script with an error code
 set -e
 
+# Run the standard tests and manual examples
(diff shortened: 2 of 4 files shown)


james-d-mitchell:
This is great!

james-d-mitchell:
Should it maybe also run `testinstall.tst` in both test files just to make sure that they also work?

mtorpey:
`SemigroupsTestInstall()` is run at the start and end of the tests, and it's all rebased onto `unstable-3.0`.  I expect the Travis test to pass.

PR #200, opened 2016-07-29
Delete redundant OneImmutable code for congruence classes

mtorpey:
This is no longer used anywhere, and it's not clear what `One` should mean for a congruence class anyway.


james-d-mitchell:
Looks good!

Issue #192, opened 2016-06-23
MinimalIdeal doesn't work well for very high degree but small size transformation semigroups


james-d-mitchell:
In the following example:

    S := AsSemigroup(IsTransformationSemigroup, FullPBRMonoid(2));
    MinimalIdeal(S);

GAP runs out of memory trying to create the `GraphOfRightActionOnPairs` digraph which has _O(65536 ^ 2)_ many vertices. But we know the whole semigroup structure and so shouldn't use this method in this case.

wilfwilson:
I agree this is a big issue, but I'm not sure how to proceed.

To calculate a transformation representation of the full PBR monoid we enumerate it, but that information is not stored for the resulting transformation semigroup. That information is attached to the full PBR monoid. And as it's the transformation semigroup that you're hoping to compute with, I'm not sure what to do.

Explicitly:
```
gap> B := FullPBRMonoid(2);;
gap> S := AsSemigroup(IsTransformationSemigroup, B);;
gap> KnownAttributesOfObject(S);
[ "Size", "GeneratorsOfMagma", "GeneratorsOfMagmaWithOne",
  "MultiplicativeNeutralElement" ]
gap> KnownPropertiesOfObject(S);
[ "IsEmpty", "IsTrivial", "IsNonTrivial", "IsFinite",
  "CanEasilyCompareElements", "IsDuplicateFree", "IsAssociative",
  "IsGeneratorsOfSemigroup", "IsMonoidAsSemigroup", "IsSemigroupIdeal",
  "IsPartialPermBipartitionSemigroup", "IsPermBipartitionGroup" ]
gap> KnownPropertiesOfObject(S) = KnownPropertiesOfObject(B);
true
```
Whereas for `B`, the full PBR monoid, we know a lot more:
```
gap> KnownAttributesOfObject(B);
[ "Size", "Representative", "OneImmutable", "GeneratorsOfMagma",
  "GeneratorsOfMagmaWithOne", "MultiplicativeNeutralElement",
  "CayleyGraphSemigroup", "IsomorphismTransformationSemigroup",
  "DegreeOfPBRSemigroup", "GenericSemigroupData" ]
```
So GAP doesn't (explicitly) know the full semigroup structure of `S`, it only knows things about `B`, which it has forgotten is isomorphic to `S` by some isomorphism. How do we solve this problem?

james-d-mitchell:
[...]

PR #204, opened 2016-08-17
Make RepresentativeOfMinimalIdeal check for high degree (Issue #192)

wilfwilson:
This PR addresses Issue #192 by comparing the size of the semigroup (if known) to the degree of the transformation semigroup in `RepresentativeOfMinimalIdeal`, and doing `TryNextMethod` if appropriate.

I also check for the degree of the transformation semigroup (regardless of the size of the semigroup). If the degree is greater than some fixed number, we determine that the algorithm will take too long in any circumstance, and so we may as well try the next method. This requires a somewhat arbitrary decision for what that number should be - I went with degree  10,000 as the limit.

### DIFF
--- a/gap/semigroups/semitrans.gi
+++ b/gap/semigroups/semitrans.gi
@@ -607,17 +607,15 @@ function(S)
   local gens, nrgens, n, min_rank, rank, min_rank_index, graph, nrpairs, elts,
   marked, squashed, j, t, im, reduced, y, i, k, x;
 
-  gens := GeneratorsOfSemigroup(S);
-  nrgens := Length(gens);
   n := DegreeOfTransformationSemigroup(S); # Smallest n such that S <= T_n
                                            # We must have n >= 2.
-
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
@wilfwilson this looks good, I'm happy to merge this mod the following question. Shouldn't we make the same changes for semigroups of partial perms too?

wilfwilson:
@james-d-mitchell For partial perms, the "graph" method for `RepresentativeOfMinimalIdeal` only seems slow down when the degree is huge, in which case the other methods are slow for the same reasons. I can come up with a couple of examples where other methods are a little quicker, but nothing which makes me think it would be a good idea to make a change as we did for transformations.

Overall I don't think we should make any changes for partial perms.

wilfwilson:
@james-d-mitchell I'm ready for this to be merged.

james-d-mitchell:
I'm happy with this, merging now.

PR #205, opened 2016-08-17
Introduce IndexPeriodOfSemigroupElement, and related functions

wilfwilson:
This PR introduces a new function `IndexAndPeriod` which calculates the index and the period of an arbitrary semigroup element. Previously, functions only existed for transformations and partial perms. I have added a special method which works for elements which have a rank function, which includes bipartitions. There are tests and documentation for the new code.

Additionally, for monogenic semigroups or monoids, the size of the semigroup can be calculated from the index and the period of the generator. Therefore I have installed methods to calculate the size of monogenic semigroups/monoids (with known minimal generating sets) which use this fact.


wilfwilson:
I'll have to wait until the tests in unstable-3.0 are fixed to know fully if this PR passes the tests.

james-d-mitchell:
This look pretty good. One question that immediately comes to mind is: if `IndexAndPeriodByRank` is only new method for bipartitions, then why isn't it just a method for `IndexAndPeriodOfBipartition`?

I want to look more closely at the code, and will comment some more then

james-d-mitchell:
The code coverage of `standard/elements.tst` is not very good.

wilfwilson:
@james-d-mitchell I've got the code coverage back up to 100% - I made some changes this afternoon which had brought the coverage down.

There is currently no operation called `IndexAndPeriodOfBipartition` and I don't think we should introduce one. There exist functions `IndexPeriodOfTransformation` and `IndexPeriodOfPartialPerm` in the GAP library, so they can't be removed, but `IndexAndPeriod` should be the user-facing operation, with method selection left to work out the appropriate method. Index and period are definitions which do not depend on representation, so I don't think each representation should have a named function.

However, to get to your point: why is `IndexAndPeriodByRank` a separate function, and not just the directly installed method for `IndexAndPeriod` for `IsBipartition`? Since no other type of element currently uses `IndexAndPeriodByRank`, it certainly could be. But seeing as `IndexAndPeriodByRank` has nothing specific to do with bipartitions, I think it makes sense to keep it separate. Perhaps new elements will be introduced in the future which have a rank function (semigroups of matrices over finite fields, perhaps?) - if this happened it'd be simple to install a method for `IndexAndPeriod` for these elements which calls the separate `IndexAndPeriodByRank` function.

I don't think see any penalty with the way things are, so I propose to leave it as is.

james-d-mitchell:
We already support matrices over finite fields, so why not use the rank method there? Also I think the name of this should be `IndexPeriodOfSemigroupElement` for consistency with the existing functions of which, as you say, we cannot change the names.

wilfwilson:
@james-d-mitchell I agree with your suggestion for `IndexPeriodOfSemigroupElement`.

For matrices over finite fields, is the "rank" of a matrix (in the sense that things are D-related in the full matrix semigroup iff they have the same rank) equal to the dimension of its column space = the dimension of its row space?

So for a square matrix over a finite field I could have:
```
rank := x -> Dimension(VectorSpace(BaseDomain(x), AsList(x)));
```
Can you think of any other types of semigroup element which have a rank function?

wilfwilson:
@james-d-mitchell The operation is now called `IndexPeriodOfSemigroupElement` and the method installed for matrices over finite fields uses the rank method.

james-d-mitchell:
Great! Can you please add some tests for finite field matrices too? And can you add a method for boolean matrices too? Possibly for other types of matrices over semirings.

Also there are several other operations/methods that make use of `IndexPeriodOfSemigroupElement`:

* `SmallestIdempotentPower`
* `Order`

Could you please install methods for these too?

I've reviewed your code and tried iton some large examples (of bipartitions arising from transformations), and it works pretty well. It does exactly what I would have hoped. The bottleneck seems to be the calculation of the period, but I don't know a better way of computing it than the way you did it.

wilfwilson:
- There are tests for matrices over finite fields in `tst/standard/ffmat.tst`.
- We decided that I will work on methods for matrices over other types of semirings separately, and put them into another pull request.
- We decided not to make a generic function for `Order` at the moment, since it would conflict with the documentation for the GAP function `Order`. (There does currently exist an undocumented `Order` function in the GAP library for transformations, but I guess this will have to stay there).
- I have made a generic `SmallestIdempotentPower` method which uses `IndexPeriodOfSemigroupElement`.

james-d-mitchell:
Looks good (mod some trailing whitespace in elements.gi), I'll merge it now

PR #208, opened 2016-08-24
LatticeOfCongruences: add restriction argument

mtorpey:
`LatticeOfCongruences` can now take an optional `restriction` argument.

If `restriction` is specified and is a collection of elements from `S`, then the result will only include congruences generated by pairs of elements from `restriction`.  Otherwise, all congruences will be calculated (as before).


wilfwilson:
@mtorpey it is unclear what the numbers in the lattice correspond to, when the `restriction` option is specified.

james-d-mitchell:
@mtorpey I agree with Wilf, it is not clear what the numbers mean, and in particular, it does not seem to be possible to get our hands on the actual congruences in the lattice if the `restriction` option is used. 

Shouldn't there be a `CongruencesOfSemigroup` operation which has `restriction` parameter, so that we can recover the congruences from the lattice?

mtorpey:
I'm working on this at the moment, and I think the best solution would be to have a function which takes a congruence lattice and returns the list of congruences.  For now, I'm calling this `CongruencesOfLattice`, but if anyone can think of a better name I'd like to hear it.

mtorpey:
I've updated this with the changes I described.  How's that?

wilfwilson:
In the documentation for `LatticeOfCongruences` I think you should explicitly say that the numbering corresponds to the numbering of `CongruencesOfSemigroup` if `restriction` is not specified, and that if `restriction` is specified then the numbering corresponds to what get by doing `CongruencesOfLattice` on the lattice.

I'm not keen on the name `CongruencesOfLattice` since I would expect this give the congruences on a particular instance of the algebraic object known as a lattice. Unfortunately I don't have a nicer suggestion. `CongruencesOfLatticeOfCongruences` would avoid this issue but is horrible.

wilfwilson:
I'll be happy to merge this with your changes, as long as the tests pass.

PR #209, opened 2016-08-25
Install selective Size methods for monogenic semigroups

wilfwilson:
Previously all monogenic semigroups which knew their minimal
generating set would use IndexPeriod to calculate their size.
For generic semigroup elements, IndexPeriod would then have to
recreate the monogenic semigroup, enumerate it, and hence deduce
its size - and then do even more work. This method is now only used
for elements which have a better IndexPeriod method.


james-d-mitchell:
Looks good except that the convention used elsewhere in the package is to name a function which will be deleted/unbound after being used in an `InstallMethod` starting with an underscore. 

Also I installed a method for `IsFinite` in PR #212, so you could remove that from the tests.

james-d-mitchell:
Sorry, the convention is to call the function `_SemigroupSizeByIndexPeriod`, i.e. without the prefix `SEMIGROUPS`, please revise.

wilfwilson:
@james-d-mitchell no worries, I've sorted the naming out now.

james-d-mitchell:
Looks great!

Issue #210, opened 2016-08-25
Size of an exhaustive matrix monoid over a finite field


wilfwilson:
I have come across this problem where the size is being wrongly reported, in `unstable-3.0`:
```
gap> x := Matrix(GF(2^2),
> [[Z(2^2), 0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2)],
>  [Z(2^2), 0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2)],
>  [0*Z(2), Z(2)^0, 0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2)],
>  [0*Z(2), 0*Z(2), Z(2)^0, 0*Z(2), 0*Z(2), 0*Z(2)],
>  [0*Z(2), 0*Z(2), 0*Z(2), Z(2)^0, 0*Z(2), 0*Z(2)],
>  [0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2), 0*Z(2)]]);;
gap> S := Monoid(x, rec(generic := true));
<monoid of 6x6 matrices over GF(2^2) with 1 generator>
gap> HasIsFinite(S);
false
gap> SEMIGROUP_SIZE(GenericSemigroupData(S));
8
gap> Size(S);
8
gap> Length(Elements(S));
7
```
Some explanation:

This came up when I was coming up with an example for my "size of a monogenic semigroup/monoid" stuff. Since this is a matrix over a finite field, it should use the `IndexPeriodOfSemigroupElement` method for `Size`. However this method is not being called, since the semigroup `S` does not know that it is monogenic, which is because the immediate method for `IsMonogenicSemigroup` is not being triggered. The immediate method includes the filter `IsFinite`, and the matrix semigroup `S` does not know that it is finite at creation. I would expect `IsFinite` to be set true at creation for a semigroup of matrices over finite fields. Is this another bug, or is this as expected?

james-d-mitchell:
I think this is probably due to some wrong hash function, I'll have to check and get back to you.

Issue #211, opened 2016-08-25
Errors computing with the full matrix semigroup


wilfwilson:
I get the following errors in `unstable-3.0`. I assume this is unrelated to the other issue I posted today, since that did not involve an acting semigroup.
```
gap> S := FullMatrixSemigroup(3, 4);
<general linear monoid 3x3 over GF(2^2)>
gap> One(S) in S;
true
gap> H := GroupHClass(DClass(S, One(S)));
<Green's H-class: Matrix(GF(2^2), [[Z(2)^0, 0*Z(2), 0*Z(2)], [0*Z(2), Z(2)^0, 0*Z(2)],
   [0*Z(2), 0*Z(2), Z(2)^0]])>
gap> IsomorphismPermGroup(H);
Error, List Element: <list>[3] must have an assigned value in
  x := gens[genpos[1][i]] * trace( genpos[2][i] )
 ; at /Users/Wilf/GAP/pkg/semigroups/gap/main/lambda-rho.gi:123 called from
trace( i ); at /Users/Wilf/GAP/pkg/semigroups/gap/main/lambda-rho.gi:144 called from
LambdaOrbMult( o, m, i
 ) at /Users/Wilf/GAP/pkg/semigroups/gap/greens/greens-acting.gi:198 called from
SEMIGROUPS.RectifyLambda( H
 ); at /Users/Wilf/GAP/pkg/semigroups/gap/greens/greens-regular.gi:64 called from
SchutzenbergerGroup( H
 ) at /Users/Wilf/GAP/pkg/semigroups/gap/greens/greens-acting.gi:1614 called from
<function "unknown">( <arguments> )
brk> quit;
gap> Random(S);
Error, Semigroups: * (for matrices over a finite field): usage,
the degree or domain of the arguments do not match, at /Users/Wilf/GAP/pkg/semigroups/gap/elements/ffmat.gi:813 called from
LambdaOrbRep( o, m ) * Random( LambdaOrbSchutzGp( o, m ) )
  at /Users/Wilf/GAP/pkg/semigroups/gap/main/semigroups-acting.gi:454 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:32
```

james-d-mitchell:
@markuspf any chance you could look at this Markus, this is in code you wrote for Semigroups 2.6, which I merged into unstable-3.0 earlier in the summer.

james-d-mitchell:
No, wait, I think I fixed it

PR #212, opened 2016-08-25
Replace hash function for IsPlistMatrix/RowBasisOverFiniteFieldRep

james-d-mitchell:
Remove unused hash functions, fix hash functions for matrices and row bases,
make sure that IsMatrixSemigroup implies IsFinite (since IsMatrixSemigroup =
IsMatrixOverFiniteFieldSemigroup), add some tests.

Resolves Issues #210 and #211.

### DIFF
--- a/gap/elements/ffmat.gi
+++ b/gap/elements/ffmat.gi
@@ -59,6 +59,12 @@ function(mat, pos)
 end);
 
 InstallMethod(MatrixNC, "for a matrix over finite field and mutable list",
+[IsMatrixOverFiniteField, IsList],
+function(sample, mat)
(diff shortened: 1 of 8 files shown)


james-d-mitchell:
@wilfwilson can you please review this, when the tests are done?

wilfwilson:
@james-d-mitchell I don't know a lot about hash functions but what you've done seems reasonable and everything else is fine. The only problem is a diff in `semiffmat.tst` when it's run using the exhaustive methods.

james-d-mitchell:
I've removed the test output, if the test run, will you merge @wilfwilson?

PR #214, opened 2016-08-26
Improve inverse semigroup congruences

mtorpey:
This pull request re-enables inverse semigroup congruences by kernel and trace, with the following improvements:
-  Kernel-trace is only selected when the size of the semigroup is at least 100,000 - configurable through `inv_threshold` which can be set via:
  - the default options rec (at runtime or in `options.g`)
  - a record passed to `Semigroup(...)`
  - a record passed to `SemigroupCongruence(...)`
- Two significant bugs fixed
- Small performance improvements (re-creating classes' type object was taking a long time)
- Documentation updated


doc/conginv.xml
```diff
@@ -111,25 +111,24 @@ gap> cong2 := AsInverseSemigroupCongruenceByKernelTrace(cong);
     <Description>
       This category contains any inverse semigroup congruence <A>cong</A> which
       is represented internally by its kernel and trace.  The <Ref
-        Func = "SemigroupCongruence"/> function will almost always create an
+        Func = "SemigroupCongruence"/> function may create an
       object of this category if its first argument <A>S</A> is an inverse
-      semigroup.
+      semigroup and has sufficiently large size.  It can be treated externally
```
james-d-mitchell: I'd probably remove the word ''externally'' from this sentence.

gap/options.g
```diff
@@ -9,13 +9,14 @@
...
-      regular    := false,
-      generic    := false,
-      report     := true,
-      batch_size := 8192, 
-      nr_threads := 4);
+  rec(small         := false,
+      hashlen       := rec(S := 251, M := 6257, L := 25013),
+      regular       := false,
+      generic       := false,
+      inv_threshold := 10^5,
```
james-d-mitchell: Missing spaces around `^` in `10^5`

mtorpey:
2: The name of `inv_threshold` is going to be a tradeoff between being descriptive and being usable.  I'm going for `cong_by_ker_trace_threshold` for the moment, as this is reasonably unambiguous, short enough to use, and follows the naming formats of the other options.

Issue #218, opened 2016-09-12
Faulty InjectionPrincipalFactor method for acting D-classes


james-d-mitchell:
gap> S := AnnularJonesMonoid(6);
    <regular bipartition *-monoid of degree 6 with 2 generators>
    gap> MaximalSubsemigroups(S);
    Error, usage: the second argument <s> does not belong to the
    underlying semigroup of the first argument <R>, at /Users/jdm/gap/lib/reesmat.gi:741 called from
    RMSElement( S, lookup_rows[x![1]] ^ p.row, r[x![1]] * x![2]
      * invert( c[x![3]] ), lookup_cols[x![3]] ^ p.col
     ) at /Users/jdm/gap/pkg/semigroups/gap/semigroups/semirms.gi:745 called from
    map!.fun( elm ) at /Users/jdm/gap/lib/mapprep.gi:591 called from
    ImageElm( map, elm ) at /Users/jdm/gap/lib/mapping.gi:835 called from
    x
    ^ iso
      at /Users/jdm/gap/pkg/semigroups/gap/attributes/maximal.gi:888 called from
    func( C[i] ) at /Users/jdm/gap/lib/coll.gi:746 called from
    ...  at *stdin*:2
    you can 'quit;' to quit to outer loop, or
    you can 'return;' to continue
    brk>

james-d-mitchell:
Same thing happens for:

    S := Semigroup(PlanarPartitionMonoid(5), AsBipartition((1,2,3,4,5)));

wilfwilson:
This is a bug in `InjectionPrinicpalFactor` for a D-class. As shown below, the injection is mapping an element of the D-class to an element which isn't in the principal factor Rees 0-matrix semigroup:
```
gap> S := AnnularJonesMonoid(5);;
gap> x := Bipartition([[1, -1], [2, 3], [4, -2], [5, -5], [-3, -4]]);;
gap> x in S;
true
gap> D := DClass(S, x);;
gap> x in D;
true
gap> inj := InjectionPrincipalFactor(D);;
gap> R := Range(inj);;
<Rees 0-matrix semigroup 5x5 over Group([ (2,4,3) ])>
gap> G := UnderlyingSemigroup(R);
Group([ (2,4,3) ])
gap> x ^ inj;
(2,(3,4),4)
gap> x ^ inj in R;
true
gap> UnderlyingElementOfReesZeroMatrixSemigroupElement(x ^ inj) in G;
false
```

[...]

PR #220, opened 2016-09-13
Resolve Issue #218

james-d-mitchell:
There was an old and incorrect method for InjectionPrincipalFactor
that was earlier removed in stable-2.8 but not in unstable-3.0
(for some reason). This method is removed, tests added, and an
additional check is added to the proper method for
InjectionPrincipalFactor which checks if an element is in the
D-class in the injection and returns fail if it is not.


wilfwilson:
This PR fixes the issue whilst removing code - nice.

PR #226, opened 2016-10-12
Remove Small/SmallerDegreeTransformationRepresentation

mtorpey:
We've established that these functions currently perform badly and aren't very useful, so they shouldn't go in unstable-3.0 at the moment.  I've got a branch on my fork which readds them, if we need them later.


wilfwilson:
Good.

Issue #227, opened 2016-10-19
Some cong tests in error


naipmoro:
Hello,
I am on ubuntu 16.04, using the latest 4.8.5 gap, which I compiled. All the recent 4.8.x
versions of _semigroups-2.8.0_ give the same test errors in simple-cong and reesmat-cong:

 ```Running list 195 . . .
 Diff in 
[ 
  "/home/naip/apps/gap/gap4r8p5/pkg/semigroups-2.8.0/doc/simple-cong.xm\
l", 52, 63 ]
 **Input is:**
congs := CongruencesOfSemigroup(s);
 **Expected output:**
[ <universal semigroup congruence over 
    <Rees 0-matrix semigroup 2x2 over Sym( [ 1 .. 3 ] )>>, 
  <semigroup congruence over <Rees 0-matrix semigroup 2x2 over 
      Sym( [ 1 .. 3 ] )> with linked triple (1,2,2)>, 
  <semigroup congruence over <Rees 0-matrix semigroup 2x2 over 
      Sym( [ 1 .. 3 ] )> with linked triple (3,2,2)>, 
  <semigroup congruence over <Rees 0-matrix semigroup 2x2 over 
      Sym( [ 1 .. 3 ] )> with linked triple (S3,2,2)> ]
 **But found:**
[ <universal semigroup congruence over 
    <Rees 0-matrix semigroup 2x2 over Sym( [ 1 .. 3 ] )>>, 
  <semigroup congruence over <Rees 0-matrix semigroup 2x2 over 
      Sym( [ 1 .. 3 ] )> with linked triple (S3,2,2)>, 
  <semigroup congruence over <Rees 0-matrix semigroup 2x2 over S3>
      with linked triple (3,2,2)>, 
  <semigroup congruence over <Rees 0-matrix semigroup 2x2 over S3>
      with linked triple (1,2,2)> ]

 Running list 204 . . .
 Diff in 
[ 
  "/home/naip/apps/gap/gap4r8p5/pkg/semigroups-2.8.0/doc/reesmat-cong.x\
ml", 282, 288 ]
 **Input is:**
MeetSemigroupCongruences(congs[2], congs[3]);
 **Expected output:**
<semigroup congruence over <Rees 0-matrix semigroup 2x2 over 
  Sym( [ 1 .. 3 ] )> with linked triple (1,2,2)>
 **But found:**
<semigroup congruence over <Rees 0-matrix semigroup 2x2 over 
[...]

PR #228, opened 2016-10-21
Fix issue #227: congruence manual examples failing

mtorpey:
This issue was caused by a change in the order of the list given by `CongruencesOfSemigroup` for a Rees matrix semigroup.  As such, it was a harmless diff, and I've modified the tests so that they don't rely on ordering.

### DIFF
--- a/doc/reesmat-cong.xml
+++ b/doc/reesmat-cong.xml
@@ -282,8 +282,8 @@ gap> Size(classes);
     <Example><![CDATA[
 gap> S := ReesZeroMatrixSemigroup(SymmetricGroup(3), 
 > [[(),(1,3,2)],[(1,2),0]]);;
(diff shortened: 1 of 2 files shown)


mtorpey:
Note that there are currently other tests failing in the `stable-2.8` branch, so Travis won't be happy.

james-d-mitchell:
@mtorpey the other failing tests are related to me preparing a new version of Semigroup (2.8.1) which will work against the changes in the fix-semi-iso branch of my fork of the GAP repo. The cause of the failures is that the fix-semi-iso branch PR was not accepted, and I haven't had time to rework it fully yet. So, this is might take a while.

mtorpey:
My understanding is that we're not planning to make another 2.x.x release of Semigroups, so I'm closing this on the basis that it's fixed in the `master` branch and won't be a problem in 3.0

PR #231, opened 2016-11-17
Fix broken conginv.tst

mtorpey:
The problem was related to the removal of `GenericSemigroupData`.  We used to have `idsdata := GenericSemigroupData(idsmgp)`, and then we'd look up positions in `idsdata`.  Sensibly, we've removed the line where we set `idsdata`, but it was still trying to look things up in it.

I've now removed the variable, and we just look things up **in the semigroup itself**, using `Position(idsmgp, x)` and so on, which I understand is now the way we do things.

### DIFF
--- a/gap/congruences/conginv.gi
+++ b/gap/congruences/conginv.gi
@@ -472,7 +472,7 @@ SEMIGROUPS.KernelTraceClosure := function(S, kernel, traceBlocks, pairstoapply)
   # "traceBlocks" in its trace, and containing all the given pairs
   # TODO Review this JDM for use of Elements, AsList etc. Could iterators work
   # better?
-  local idsmgp, idsdata, idslist, slist, kernelgenstoapply, gen, nrk, nr,
+  local idsmgp, idslist, slist, kernelgenstoapply, gen, nrk, nr,
         traceUF, i, pos1, j, pos, hashlen, ht, treehashsize, right, genstoapply,
         NormalClosureInverseSemigroup, enumerate_trace, enforce_conditions,
         compute_kernel, oldLookup, oldKernel, trace_unchanged, kernel_unchanged;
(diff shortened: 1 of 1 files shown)


gap/congruences/conginv.gi
```diff
@@ -490,16 +490,16 @@ SEMIGROUPS.KernelTraceClosure := function(S, kernel, traceBlocks, pairstoapply)
   od;
   nrk := Length(kernelgenstoapply);
   Elements(kernel);
```
james-d-mitchell: What's this `Elements(kernel);` here for? I realise that this is not a line that you changed here, but this puzzled me when I looked at this function earlier too (see my comment in line 473 of this file).

gap/congruences/conginv.gi
```diff
@@ -490,16 +490,16 @@ SEMIGROUPS.KernelTraceClosure := function(S, kernel, traceBlocks, pairstoapply)
...
+  pairstoapply := List(pairstoapply, x -> [Position(idsmgp, RightOne(x[1])),
+                                           Position(idsmgp, RightOne(x[2]))]);
   nr := Length(pairstoapply);
 
   # Calculate traceUF from traceBlocks
   traceUF := UF_NEW(Length(idslist));
   for i in [1 .. Length(traceBlocks)] do
-    pos1 := Position(idsdata, traceBlocks[i][1]);
+    pos1 := Position(idsmgp, traceBlocks[i][1]);
     for j in [2 .. Length(traceBlocks[i])] do
```
james-d-mitchell: `Position(S, x)` only checks if `x` is known to belong to `S` or not, it does not perform any enumeration of the semigroup. For example,

    gap> S := RandomSemigroup();
    <semigroup of 4x4 max-plus matrices with 8 generators>
    gap> Position(S, S.1);
    fail

It would be safer to use `PositionCanonical` here which will enumerate the semigroup until the position of the element is determined or the semigroup is fully enumerated and the element is known to not belong to it.

mtorpey:
Merged by accident, but after addressing the changes.  `Elements(kernel)` to be replaced shortly.

Issue #230, opened 2016-11-17
Incorrect results in MaximalSubsemigroups


wilfwilson:
Whilst investigating the maximal subsemigroups of the inverse monoid PORI_n, I have discovered a problem with my code.

There are maximal subsemigroups of PORI_n which arise from the group of units (no problem there), and some which arise from the D-classes of elements of rank n-1 (these are the problem ones). The `MaximalSubsemigroups` code is missing some maximal subsemigroups of this type which intersect every H-class. Since it does not find them, it believes that the subsemigroup formed by removing this entire D-class is maximal, which is not the case.

I currently do not know why this is occurring. I hope it's nothing disastrous.
```
gap> S := Monoid( [ PartialPerm( [ 1, 2, 3, 4, 5, 6 ], [ 2, 3, 4, 5, 6, 1 ] ),
>  PartialPerm( [ 1, 2, 3, 4, 5 ], [ 1, 2, 3, 4, 6 ] ),
>  PartialPerm( [ 1, 2, 3, 4, 6 ], [ 1, 2, 3, 4, 5 ] ),
>  PartialPerm( [ 1, 2, 3, 4, 5, 6 ], [ 6, 5, 4, 3, 2, 1 ] ) ] );; # PORI_6
gap> U := InverseMonoid( [ PartialPerm( [ 1, 2, 3, 4, 5, 6 ], [ 6, 1, 2, 3, 4, 5 ] ),
>  PartialPerm( [ 1, 2, 3, 4, 5, 6 ], [ 3, 2, 1, 6, 5, 4 ] ),
>  PartialPerm( [ 2, 3, 4, 5, 6 ], [ 2, 3, 4, 5, 6 ] ),
>  PartialPerm( [ 2, 3, 5, 6 ], [ 6, 5, 4, 3 ] ),
>  PartialPerm( [ 2, 3, 4, 6 ], [ 5, 4, 1, 6 ] ) ] );;
gap> IsMaximalSubsemigroup(S, U);
true
gap> U in MaximalSubsemigroups(S);
false
gap> V := InverseMonoid( [ PartialPerm( [ 1, 2, 3, 4, 5, 6 ], [ 2, 3, 4, 5, 6, 1 ] ),
>  PartialPerm( [ 1, 2, 3, 4, 5, 6 ], [ 6, 5, 4, 3, 2, 1 ] ),
>  PartialPerm( [ 1, 2, 3, 4 ], [ 1, 4, 3, 2 ] ),
[...]

PR #232, opened 2016-11-18
Fix issue #230: MaximalSubsemigroups

wilfwilson:
After carefully stepping through the code with a failing example, I identified a single incorrect line in the code for `MaximalSubsemigroups` for a Rees 0-matrix semigroup which intersect every H-class and contain a given set of elements. Essentially I was creating an element as `x * y` rather than `y * x`.

I also took this opportunity to use the variable name `zero` instead of `Z`, since apparently `Z` is already defined in GAP and is (should be?) read only.


wilfwilson:
I'm just going to merge this because it's a one-line fix, a variable-name change, and tests.

PR #234, opened 2016-12-07
UF_BLOCKS: improve performance

mtorpey:
I've implemented several of James's suggestions here.  Note that the output of `UF_BLOCKS` is supposed to have holes in it, so `T_PLIST_TAB` is not appropriate.  `nullptr` can be included, and indicates where a hole should be.

### DIFF
--- a/src/ufdata.cc
+++ b/src/ufdata.cc
@@ -73,34 +73,24 @@ Obj UF_TABLE(Obj self, Obj ufdata) {
   return gap_table;
 }
 
-// FIXME it is not valid to store Obj's inside a C++ container such as a vector
-
 Obj UF_BLOCKS(Obj self, Obj ufdata) {
-  UFData::blocks_t* blocks = CLASS_OBJ<UFData*>(ufdata)->get_blocks();
-  size_t            size   = blocks->size();
-  size_t            i, j;
+  UFData::blocks_t const* blocks = CLASS_OBJ<UFData*>(ufdata)->get_blocks();
+  size_t                  size   = blocks->size();
+  size_t                  i, j;
 
-  // Rewrite each block as a PLIST object
-  std::vector<Obj> obj_list;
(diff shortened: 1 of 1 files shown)


src/ufdata.cc
```diff
@@ -73,35 +73,24 @@ Obj UF_TABLE(Obj self, Obj ufdata) {
...
-      SET_LEN_PLIST(obj_list[i], blocks->at(i)->size());
-      for (j = 0; j < blocks->at(i)->size(); j++) {
-        SET_ELM_PLIST(obj_list[i], j + 1, INTOBJ_INT(blocks->at(i)->at(j) + 1));
+    if ((*blocks)[i] != nullptr) { // nullptr represents a hole in the list
+      Obj block = NEW_PLIST(T_PLIST_CYC, (*blocks)[i]->size());
+      SET_LEN_PLIST(block, (*blocks)[i]->size());
+      for (j = 0; j < (*blocks)[i]->size(); j++) {
+        SET_ELM_PLIST(block, j + 1, INTOBJ_INT((*(*blocks)[i])[j] + 1));
       }
+      SET_ELM_PLIST(gap_blocks, i + 1, block);
```
james-d-mitchell: Think you need a `CHANGED_BAG(gap_blocks);` here.

src/ufdata.cc
```diff
@@ -73,35 +73,24 @@ Obj UF_TABLE(Obj self, Obj ufdata) {
   return gap_table;
 }
 
-// FIXME it is not valid to store Obj's inside a C++ container such as a vector
-
 Obj UF_BLOCKS(Obj self, Obj ufdata) {
   UFData::blocks_t* blocks = CLASS_OBJ<UFData*>(ufdata)->get_blocks();
```
james-d-mitchell: This could be `UFData::block_t const*` since  `blocks` is not modified by this function.

PR #235, opened 2016-12-07
Convert ReadPackage single argument calls to two arguments

fingolfin:


fingolfin:
The test failure seems to be an issue with your test setup?!

james-d-mitchell:
@fingolfin yes, indeed, it does! Thanks for the PR.

james-d-mitchell:
I've added these changes to the `unstable-3.0` branch, which has diverged a lot from `stable-2.8`. The next release will be made from a descendant of `unstable-3.0` so I am going to close this PR.

PR #236, opened 2016-12-07
Include RandomIDTest

ffloresbrito:
Input a semigroup S and an identity in the form of two lists and determines whether it holds in S.


ffloresbrito:
After some feedback, I have decided to close this pull request.

PR #239, opened 2016-12-12
Add Lookup and CanonicalLookup methods for generic congruences

mtorpey:
`EquivalenceRelationLookup` and `EquivalenceRelationCanonicalLookup` now have methods for generic congruences.  This allows us to compare congruences of different types, and to compare their classes (a test which is readded here, having been unwisely removed a few commits ago).

Doc and tests are adjusted, and a few other little features are added.

### DIFF
--- a/gap/congruences/cong.gi
+++ b/gap/congruences/cong.gi
@@ -21,9 +21,32 @@
 ## cong.gd contains declarations for many of these.
 ##
 
+InstallMethod(\=, "for two semigroup congruences",
+[IsSemigroupCongruence, IsSemigroupCongruence],
+function(c1, c2)
+  if not IsFinite(Range(c1)) then
+    TryNextMethod();
+  fi;
+  return Range(c1) = Range(c2)
+         and EquivalenceRelationCanonicalLookup(c1) =
+             EquivalenceRelationCanonicalLookup(c2);
+end);
+
+InstallMethod(\=, "for two semigroup congruences with generating pairs",
+[IsSemigroupCongruence and HasGeneratingPairsOfMagmaCongruence,
+ IsSemigroupCongruence and HasGeneratingPairsOfMagmaCongruence],
+function(c1, c2)
+  return Range(c1) = Range(c2)
+         and ForAll(GeneratingPairsOfSemigroupCongruence(c1), p -> p in c2)
+         and ForAll(GeneratingPairsOfSemigroupCongruence(c2), p -> p in c1);
+end);
+
 InstallMethod(\=, "for a left and a right semigroup congruence",
 [IsLeftSemigroupCongruence, IsRightSemigroupCongruence],
 function(c1, c2)
+  if not IsFinite(Range(c1)) then
+    TryNextMethod();
+  fi;
   return Range(c1) = Range(c2)
          and EquivalenceRelationCanonicalLookup(c1) =
              EquivalenceRelationCanonicalLookup(c2);
@@ -32,6 +55,9 @@ end);
 InstallMethod(\=, "for a right and a left semigroup congruence",
 [IsRightSemigroupCongruence, IsLeftSemigroupCongruence],
 function(c1, c2)
+  if not IsFinite(Range(c1)) then
+    TryNextMethod();
+  fi;
   return Range(c1) = Range(c2)
          and EquivalenceRelationCanonicalLookup(c1) =
              EquivalenceRelationCanonicalLookup(c2);
@@ -50,6 +76,14 @@ function(class1, class2)
   return EquivalenceClassOfElementNC(EquivalenceClassRelation(class1),
                                      Representative(class1) *
                                      Representative(class2));
+end);
+
+InstallMethod(\=,
+"for two congruence classes",
+[IsCongruenceClass, IsCongruenceClass],
+function(class1, class2)
+  return EquivalenceClassRelation(class1) = EquivalenceClassRelation(class2)
+         and Representative(class1) in class2;
 end);
 
 InstallMethod(\<,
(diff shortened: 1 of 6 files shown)


gap/congruences/cong.gi
```diff
@@ -390,17 +424,69 @@ function(class, elm)
...
-EquivalenceRelationCanonicalLookup);
+SEMIGROUPS.GenericCongLookup);
 
 InstallMethod(EquivalenceRelationLookup,
 "for a right semigroup congruence",
 [IsRightSemigroupCongruence],
-EquivalenceRelationCanonicalLookup);
+SEMIGROUPS.GenericCongLookup);
+
+SEMIGROUPS.GenericCongCanonicalLookup := function(cong)
```
james-d-mitchell: Same comment for this as for `GenericCongLookup` I don't think this needs to hang around in `SEMIGROUPS`

tst/standard/cong.tst
```diff
@@ -324,10 +324,98 @@ gap> cong2 := SemigroupCongruence(S, pair2);;
...
+  4, 4 ]
+
+# EquivalenceRelationCanonicalLookup with an RMS cong
+gap> S := ReesMatrixSemigroup(SymmetricGroup(3), [[(1,2), ()], [(), (1,3)]]);
+<Rees matrix semigroup 2x2 over Sym( [ 1 .. 3 ] )>
+gap> cong := RMSCongruenceByLinkedTriple(S, Group((1,2,3)),
+>                                        [[1],[2]], [[1,2]]);;
+gap> ccong := AsSemigroupCongruenceByGeneratingPairs(cong);;
+gap> cong := RMSCongruenceByLinkedTriple(S, Group((1,2,3)),
+>                                        [[1],[2]], [[1,2]]);;
```
james-d-mitchell: Duplicate definition of `cong`?

gap/congruences/cong.gi
```diff
@@ -390,17 +424,69 @@ function(class, elm)
...
+SEMIGROUPS.GenericCongLookup := function(cong)
+  local S, lookup, class, nr, elm;
+  S := Range(cong);
+  if not IsFinite(S) then
+    ErrorNoReturn("Semigroups: SEMIGROUPS.GenericCongLookup: usage,\n",
+                  "<cong> must be over a finite semigroup,");
+  fi;
+  lookup := [1 .. Size(S)];
+  for class in NonTrivialEquivalenceClasses(cong) do
+    nr := PositionCanonical(S, Representative(class));
```
james-d-mitchell: And it would be better to do `TryNextMethod();` instead of returning an error from a cryptic hidden function.

gap/congruences/cong.gi
```diff
@@ -21,9 +21,32 @@
...
+ IsSemigroupCongruence and HasGeneratingPairsOfMagmaCongruence],
+function(c1, c2)
+  return Range(c1) = Range(c2)
+         and ForAll(GeneratingPairsOfSemigroupCongruence(c1), p -> p in c2)
+         and ForAll(GeneratingPairsOfSemigroupCongruence(c2), p -> p in c1);
+end);
+
 InstallMethod(\=, "for a left and a right semigroup congruence",
 [IsLeftSemigroupCongruence, IsRightSemigroupCongruence],
 function(c1, c2)
```
james-d-mitchell: Shouldn't these methods just return `false`? Are they equal if one is a right congruence and the other is not?

tst/standard/cong.tst
```diff
@@ -324,10 +324,98 @@ gap> cong2 := SemigroupCongruence(S, pair2);;
...
+gap> EquivalenceRelationCanonicalLookup(cong);
+[ 1, 2, 3, 1, 2, 2, 2, 1, 1, 3, 4, 4, 2, 1, 1, 2, 3, 4, 4, 4, 3, 3, 3, 4 ]
+gap> EquivalenceRelationCanonicalLookup(ccong);
+[ 1, 2, 3, 1, 2, 2, 2, 1, 1, 3, 4, 4, 2, 1, 1, 2, 3, 4, 4, 4, 3, 3, 3, 4 ]
+gap> cong = ccong;
+true
```
james-d-mitchell: Which method is used here? The one for checking if the generating pairs belong or the lookup table one?

gap/congruences/cong.gi
mtorpey: It could be that they are both two-sided congruences, but specified using different sets of generating pairs (one using `LeftSemigroupCongruence` and the other using `RightSemigroupCongruence`).

tst/standard/cong.tst
mtorpey: This is to ensure that `cong` doesn't have known `GeneratingPairsOfSemigroup`, which is set when you call `AsSemigroupCongruenceByGeneratingPairs` on it.

tst/standard/cong.tst
mtorpey: It should be the lookup one - as above, we reconstruct `cong` to make sure it doesn't have generating pairs.

gap/congruences/cong.gi
```diff
@@ -390,17 +424,69 @@ function(class, elm)
   return EquivalenceClassOfElementNC(cong, Representative(class) * elm);
 end);
 
+SEMIGROUPS.GenericCongLookup := function(cong)
+  local S, lookup, class, nr, elm;
```
mtorpey: :+1:

gap/congruences/cong.gi
mtorpey: Okay, I've adapted this bit.  If the semigroup is infinite, `EquivalenceRelationLookup` can never have a sensible answer, so I'm sticking with an error message (but quoting `EquivalenceRelationLookup` instead of `SEMIGROUPS.GenericCongLookup).  If it's finite but non-enumerable then we go to `TryNextMethod()`

mtorpey:
Updated to address the comments.

PR #241, opened 2016-12-15
Left/right congruences: add missing doc

mtorpey:
A few operations for left and right congruences (essentially synonyms for equivalence versions) weren't documented.


james-d-mitchell:
Looks good!

PR #242, opened 2016-12-16
Rename OutNeighboursCopy to OutNeighboursMutableCopy (digraphs)

mtorpey:
I just made a pull request in Digraphs which changes the name of `OutNeighboursCopy` to `OutNeighboursMutableCopy`.  This change keeps compatibility between the two packages.

### DIFF
--- a/gap/greens/gren.gi
+++ b/gap/greens/gren.gi
@@ -530,7 +530,7 @@ function(S)
   gr := Digraph(List([1 .. Length(l)], i -> Concatenation(l[i], r[i])));
   gr := QuotientDigraph(gr, GreensDRelation(S)!.data.comps);
 
-  return List(OutNeighboursCopy(gr), Set);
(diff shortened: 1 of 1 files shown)


mtorpey:
Note that Travis tests will fail for this PR until the Digraphs change is merged.

wilfwilson:
We have realised that copying the out-neighbours is not necessary here. The current sole use is the line:
```
return List(OutNeighboursCopy(gr), Set);
```
However, the function `List` does not change in-place, it already makes a copy, so `OutNeighbours(gr)` would suffice here (I believe). I'll be happy to be told that I'm wrong.

I suggest Michael updates this PR so that the above line reads:
```
return List(OutNeighbours(gr), Set);
```
And then this PR will require no change from Digraphs.

mtorpey:
I've updated this as Wilf suggested.

PR #244, opened 2016-12-20
WIP: add free congruences

mtorpey:
This PR adds `IsFreeCongruence`, a category of congruence over free semigroups and free monoids, which uses our semigroups++ methods for finitely presented semigroups and monoids.

The code and documentation is (in my opinion) complete.

There are no standard tests, so a file `standard/congfree.tst` should be added - this would be enough to make this PR merge-worthy.

The file `extreme/pairs.tst.disabled` was what originally made us think to do all this, and the final commit of this PR has a few fixes so that it's compatible with the new code.  However, it's full of (mostly harmless) diffs which I haven't gone through methodically yet.  This should probably also be done at some point.


james-d-mitchell:
I fixed things in this PR and am about to push directly to unstable-3.0, so I closed this PR

PR #245, opened 2017-01-19
Make GraphOfRightActionOnPairs a digraph object

wilfwilson:
`SEMIGROUPS.GraphOfActionOnPairs` is now a Digraph, and is called `DigraphOfActionOnPairs`.

`IsSynchronizingSemigroup` and `RepresentativeOfMinimalIdeal` were the only methods which used it. These have been updated.

### DIFF
--- a/doc/semitrans.xml
+++ b/doc/semitrans.xml
@@ -209,7 +209,10 @@ gap> CyclesOfTransformationSemigroup(S);
 
 <#GAPDoc Label="IsSynchronizingSemigroup">
   <ManSection>
-    <Oper Name = "IsSynchronizingSemigroup" Arg = "S[, n]"/>
+    <Prop Name = "IsSynchronizingSemigroup" Arg = "S"
+      Label = "for a transformation semigroup" />
+    <Oper Name = "IsSynchronizingSemigroup" Arg = "S, n"
+      Label = "for a transformation semigroup and a positive integer" />
     <Returns><K>true</K> or <K>false</K>.</Returns>
     <Description>
       For a positive integer <A>n</A>,
@@ -226,9 +229,10 @@ gap> CyclesOfTransformationSemigroup(S);
       <A>S</A>.
       <P/>
       
-      Note that the semigroup consisting of the identity
-      transformation has degree <C>0</C>, and for this special case the
-      function <C>IsSynchronizingSemigroup</C> will return <K>false</K>.<P/>
+      Note that the semigroup consisting of the identity transformation is the
+      unique transformation semigroup with degree <C>0</C>. In this special
+      case, the function <C>IsSynchronizingSemigroup</C> will return
+      <K>false</K>.<P/>
 
       <Example><![CDATA[
 gap> S := Semigroup(
--- a/doc/semitrans.xml
+++ b/doc/semitrans.xml
@@ -246,3 +246,77 @@ Transformation( [ 7, 8, 8, 7, 8, 8, 8, 7, 8, 8 ] )]]></Example>
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="DigraphOfActionOnPairs">
+  <ManSection>
+    <Attr Name = "DigraphOfActionOnPairs" Arg = "S"
+      Label = "for a transformation semigroup" />
+    <Oper Name = "DigraphOfActionOnPairs" Arg = "S, n"
+      Label = "for a transformation semigroup and an integer" />
+    <Returns>A digraph, or <K>fail</K>.</Returns>
+    <Description>
+      If <A>S</A> is a transformation semigroup and <A>n</A> is a non-negative
+      integer such that <A>S</A> acts on the points <C>[1 .. <A>n</A>]</C>, then
+      <C>DigraphOfActionOnPairs(<A>S</A>, <A>n</A>)</C> returns a digraph
+      representing the <Ref Oper="OnSets" BookName="ref"/> action of <A>S</A> on
+      the pairs of points in <C>[1 ..  <A>n</A>]</C>.
+      <P/>
+      
+      If the optional argument <A>n</A> is not specified, then by default the
+      degree of <A>S</A> will be chosen for <A>n</A>; see <Ref
+        Attr="DegreeOfTransformationSemigroup" BookName="ref" />. If the
+      semigroup <A>S</A> does not act on <C>[1 ..  <A>n</A>]</C>, then
+      <C>DigraphOfActionOnPairs(<A>S</A>, <A>n</A>)</C> returns <K>fail</K>.
+      <P/>
+
+      The digraph returned by <C>DigraphOfActionOnPairs</C> has <M><A>n</A> + {n
+        \choose 2}</M> vertices: the vertices <C>[1 .. <A>n</A>]</C> correspond
+      to the points in <C>[1 .. <A>n</A>]</C>, and the remaining vertices
+      correspond to the pairs of points in <C>[1 .. <A>n</A>]</C>. This
+      correspondence is stored in the vertex labels of the digraph; see <Ref
+        Oper="DigraphVertexLabels" BookName="Digraphs" />.
+      <P/>
+
+      The edges of the digraph are defined as follows. For each pair <C>{i,
+        j}</C> in <C>[1 .. <A>n</A>]</C>, and for each generator <C>f</C> in
+      <C>GeneratorsOfSemigroup(<A>S</A>)</C> (see <Ref
+        Attr="GeneratorsOfSemigroup" BookName="ref" />), there is an edge from
+      the vertex corresponding to <C>{i, j}</C> to the vertex corresponding to
+      <C>{i ^ f, j ^ f}</C>.  Since <C>f</C> is a transformation, the set <C>{i
+        ^ f, j ^ f}</C> may correspond to a pair (in the case that <C>i ^ f
+        &lt;&gt; j ^ f</C>), or to a point (in the case that <C>i ^ f = j ^
+        f</C>).  The label of an edge (<Ref Oper="DigraphEdgeLabels"
+        BookName="Digraphs" />) is the position of the first transformation
+      within <C>GeneratorsOfSemigroup(<A>S</A>)</C> that maps the pair
+      corresponding to the source vertex to the pair/point corresponding to the
+      range vertex.
+      <P/>
+
+      Note that the digraph returned by <C>DigraphOfActionOnPairs</C> has no
+      multiple edges; see <Ref Prop="IsMultiDigraph" BookName="Digraphs"/>.
+
+      <Example><![CDATA[
+gap> x := Transformation([2, 4, 3, 4, 7, 1, 6]);;
+gap> y := Transformation([3, 3, 2, 3, 5, 1, 5]);;
+gap> S := Semigroup(x, y);
+<transformation semigroup of degree 7 with 2 generators>
+gap> gr := DigraphOfActionOnPairs(S);
+<digraph with 28 vertices, 41 edges>
+gap> OnSets([2, 5], x);
+[ 4, 7 ]
+gap> DigraphVertexLabel(gr, 16);
+[ 2, 5 ]
+gap> DigraphVertexLabel(gr, 25);
+[ 4, 7 ]
+gap> DigraphEdgeLabel(gr, 16, 25);
+1
+gap> gr := DigraphOfActionOnPairs(S, 4);
+<digraph with 10 vertices, 11 edges>
+gap> DigraphVertexLabels(gr);
+[ 1, 2, 3, 4, [ 1, 2 ], [ 1, 3 ], [ 1, 4 ], [ 2, 3 ], [ 2, 4 ], 
+  [ 3, 4 ] ]
+gap> DigraphOfActionOnPairs(S, 5);
+fail]]></Example>
+    </Description>
+  </ManSection>
+<#/GAPDoc>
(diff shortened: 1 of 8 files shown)


PackageInfo.g
```diff
@@ -11,7 +11,7 @@
 ##  <#GAPDoc Label="PKGVERSIONDATA">
 ##  <!ENTITY VERSION "3.0.0">
 ##  <!ENTITY GAPVERS "4.9.0">
-##  <!ENTITY DIGRAPHSVERS "0.5.1">
+##  <!ENTITY DIGRAPHSVERS "0.6">
```
james-d-mitchell: This should be 0.6.0 not 0.6

PackageInfo.g
```diff
@@ -211,7 +211,7 @@ Dependencies := rec(
   GAP := ">=4.9.0",
   NeededOtherPackages := [["orb", ">=4.7.3"],
                           ["io", ">=4.4.4"],
-                          ["digraphs", ">=0.5.1"],
+                          ["digraphs", ">=0.6"],
```
james-d-mitchell: Same here (this is probably not strictly necessary here, but for consistency).

doc/semitrans.xml
```diff
@@ -251,3 +251,83 @@ Transformation( [ 7, 8, 8, 7, 8, 8, 8, 7, 8, 8 ] )]]></Example>
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="DigraphOfActionOnPairs">
+  <ManSection>
+    <Attr Name = "DigraphOfActionOnPairs" Arg = "coll" Label = "for ever" />
+    <Oper Name = "DigraphOfActionOnPairs" Arg = "coll[, n]" Label = "never" />
```
james-d-mitchell: The label's here look weird. These will be visible in the index, so could they be more descriptive? "for a transformation collection" and "for a transformation collection and positive integer"

doc/semitrans.xml
```diff
@@ -251,3 +251,83 @@ Transformation( [ 7, 8, 8, 7, 8, 8, 8, 7, 8, 8 ] )]]></Example>
...
+      non-negative integer, then <C>DigraphOfActionOnPairs(<A>coll</A>,
+        <A>n</A>)</C> returns a digraph representing the <Ref Oper="OnSets"
+        BookName="ref"/> action of the transformations in <A>coll</A> on the
+      pairs of points in <C>[1 ..  <A>n</A>]</C>.  If the optional argument
+      <A>n</A> is not specified, then by default the value of
+      <C>DegreeOfTransformationCollection(<A>coll</A>)</C> will be chosen for
+      <A>n</A>.  <P/>
+
+      The <C>DigraphOfActionOnPairs</C> has <M><A>n</A> + {n \choose 2}</M>
+      vertices: the vertices <C>[1 .. <A>n</A>]</C> correspond to the points in
```
james-d-mitchell: Please check what `{n \choose 2}` looks like in the text version of the doc, and if it looks crap, please replace this with a description in words.

doc/semitrans.xml
```diff
@@ -251,3 +251,83 @@ Transformation( [ 7, 8, 8, 7, 8, 8, 8, 7, 8, 8 ] )]]></Example>
...
+      <C>[1 .. <A>n</A>]</C>, and the remaining vertices correspond to the pairs
+      of points in <C>[1 .. <A>n</A>]</C>. This correspondence is stored in
+      <Ref Oper="DigraphVertexLabels" BookName="Digraphs" />.
+      <P/>
+
+      The edges of <C>DigraphOfActionOnPairs</C> are defined as follows,
+      depending on whether or not <A>coll</A> is a semigroup. If <A>coll</A> is
+      a semigroup, define <C>gens</C> to be
+      <C>GeneratorsOfSemigroup(<A>coll</A>)</C>; otherwise define <C>gens =
+        <A>coll</A></C>.
```
james-d-mitchell: Isn't this a bit complicated? Wouldn't it make more sense to just define this as an operation on a collection of transformations? This would be at the cost of recalculating this it is needed more than once, and also a bit more typing, but would be more "pure" (I guess?)

doc/semitrans.xml
```diff
@@ -251,3 +251,83 @@ Transformation( [ 7, 8, 8, 7, 8, 8, 8, 7, 8, 8 ] )]]></Example>
...
+      <P/>
+
+      <List>
+        <Mark>Edges between pairs</Mark>
+        <Item>
+          For each two distinct pairs <C>{i, j}</C> and <C>{k, l}</C>, there is
+          an edge from the vertex corresponding to the first pair to the vertex
+          corresponding to the second pair if there exists a transformation
+          <C>f</C> in <C>gens</C> such that <C>OnSets([i, j]) = [k, l]</C>.  In
```
james-d-mitchell: There is a missing argument `f` in `OnSets` here

doc/semitrans.xml
```diff
@@ -251,3 +251,83 @@ Transformation( [ 7, 8, 8, 7, 8, 8, 8, 7, 8, 8 ] )]]></Example>
...
+          other words, there exists such an edge if some transformation in
+          <C>gens</C> maps the first pair to the second.
+        </Item>
+
+        <Mark>If <A>coll</A> is not a semigroup</Mark>
+        <Item>
+          For each pair <C>{i, j}</C> and for each point <C>k</C>, there is an
+          edge from the vertex corresponding to the pair to the vertex
+          corresponding to the point if there exists a transformation <C>f</C>
+          in <A>coll</A> such that <C>OnSets([i, j]) = [k]</C>. In other words,
```
james-d-mitchell: Same comment as above

james-d-mitchell: Also since you defined `gens` to be `coll` if `coll` is not a semigroup and `GeneratorsOfSemigroup(S)` if `coll` is a semigroup, isn't the documentation repeated unnecessarily here?

gap/semigroups/semitrans.gi
```diff
@@ -283,49 +283,72 @@ SEMIGROUPS.LargestElementRClass := function(R)
...
-  local nrgens, nrpairs, PairNumber, NumberPair, in_nbs, labels, pair,
-  isolated_pair, act, range, i, j;
-  nrgens     := Length(gens);
-  nrpairs    := Binomial(n, 2);
+InstallMethod(DigraphOfActionOnPairs,
+"for a transformation semigroup",
+[IsTransformationSemigroup],
+function(S)
+  return DigraphOfActionOnPairs(GeneratorsOfSemigroup(S),
+                                DegreeOfTransformationSemigroup(S));
```
james-d-mitchell: Isn't the `GeneratorsOfSemigroup(S)` here unnecessary? Since you anyway check in the main function if a collection is a semigroup and if it is replace it with the generators? Doesn't that make this whole method unnecessary?

gap/semigroups/semitrans.gi
```diff
@@ -283,49 +283,72 @@ SEMIGROUPS.LargestElementRClass := function(R)
...
 
-  in_nbs := List([1 .. n + nrpairs], x -> []);
-  labels := List([1 .. n + nrpairs], x -> []);
-  for i in [(n + 1) .. (n + nrpairs)] do
+  if IsSemigroup(coll) then
+    coll := GeneratorsOfSemigroup(coll);
+  fi;
+
+  nr         := Length(coll);
+  nrpairs    := Binomial(n, 2);
```
james-d-mitchell: Weird space before the `:=` here

gap/semigroups/semitrans.gi
```diff
@@ -283,49 +283,72 @@ SEMIGROUPS.LargestElementRClass := function(R)
...
-      Add(in_nbs[range], i);
-      Add(labels[range], j);
-      if range <> i and isolated_pair then
-        isolated_pair := false;
+    for j in [1 .. nr] do
+      range := NumberPair(n, OnSets(pair, coll[j]));
+      if i <> range and not i in inn_nbs[range] then
+        Add(inn_nbs[range], i);
+        Add(out_nbs[i], range);
+        Add(label[i], j);
```
james-d-mitchell: Wouldn't this be faster (both here and in later use) if you just did `AddSet`?

doc/semitrans.xml
wilfwilson: Oops! Temporary placeholders that I then overlooked...

doc/semitrans.xml
wilfwilson: In the text version this thankfully gets rendered as `The  DigraphOfActionOnPairs  has n + n choose 2 vertices`

doc/semitrans.xml
wilfwilson: The paragraphs are similar but they don't repeat the same information. However the heading for the two similar paragraphs were incorrect.

Rather than the difference between the paragraphs being the case of semigroup/not semigroup, the difference is that the first paragraph describes edges between points, and the second describes edges from a pair to a point.

I've updated the headings in the documentation.

doc/semitrans.xml
wilfwilson: I'll address this below in a separate comment on the PR, since it's fundamental. I certain agree that it's an issue.

gap/semigroups/semitrans.gi
wilfwilson: If I didn't need to keep track of edge labels, I would do this. However I am labelling edges which would make this tricky.

The ordering of `label[i]` corresponds with the ordering of `out_nbs[i]`. For example, there is an edge from `i` to `out_nbs[i][2]` with label `label[i][2]`. Suppose that I want to store the edge `i -> range` with label `j` by doing `AddSet(out_nbs[i], range)`. Then what is the proper position of `label[i]` in which to insert `j`? Or did the edge `i -> range` already exist, meaning that its label doesn't need to be stored?

Do you see a way around these problems?

wilfwilson:
I'm trying to satisfy two different uses with this function:

1. You have a transformation semigroup `S`, and you want to know the action of `S` on pairs.
2. You have a collection of transformations `coll`, and you want to know the action of those specific transformations on pairs.

They're obviously very similar. The important case in my opinion is the first one; I'm not sure if the second is needed. In the second case you obviously apply each transformation `f` in `coll` to each pair. That *would work* for the first case, but you **don't want to do that,** because it involves enumerating the whole semigroup, and you can get (essentially) the same digraph by using only the generators of the semigroup. Note however, that `DigraphOfActionOnPairs(S)` and `DigraphOfActionOnPairs(Elements(S))` are different in general (the latter will be something like the transitive closure of the first).

I see several options for better combining these two cases:

### Options

#### Only support semigroups: copy action on points

@james-d-mitchell (I presume?) wrote a currently undocumented function `DigraphOfActionOnPoints`, which accepts a transformation semigroup and an optional positive integer as arguments. This ignores the case of applying the function to an arbitrary collection of transformations.

#### Different names: copy synchronizing stuff

The Semigroups package has methods `IsSynchronizingTransformationCollection` and `IsSynchronizingTransformationSemigroup` with different names to distinguish the options. I'll just note that the documentation for `IsSynchronizingTransformationCollection` is not clear and incorrect - I'll address this in an addition to this PR.

## How to proceed

I don't think we want different names. I think that either `DigraphOfActionOnPairs` should only support semigroups; or, it should be documented primarily as a method for semigroups, but the documentation should include a note which says that for a transformation collection `coll`,
```
DigraphOfActionOnPairs(coll, n) = DigraphOfActionOnPairs(Semigroup(coll), n)
```
What do you think @james-d-mitchell ?

gap/semigroups/semitrans.gi
wilfwilson: Yes

wilfwilson:
@james-d-mitchell Actually, perhaps the better option is to have two functions:

```DigraphOfActionOfTransformationSemigroupOnPairs```

and

```DigraphOfActionOfTransformationCollectionOnPairs,```

like we do for `IsSynchronizingSemigroup` and `IsSynchronizingTransformationCollection`.

The documentation will describe `DigraphOfActionOfTransformationCollectionOnPairs` and then say that `DigraphOfActionOfTransformationSemigroupOnPairs` calls the former on the generators of the semigroup.

File: gap/semigroups/semitrans.gi (hunk 8 above)
Comment (wilfwilson) [inline gap/semigroups/semitrans.gi]: The problem then is that `DigraphOfActionOnPairs(S)` calls `DigraphOfActionOnPairs(S, DegreeOfTransformationCollection(S))`, but the method for `DegreeOfTransformationCollection` involves enumerating the semigroup... I'll add a new method to fix this.

**Comment by mtorpey** [comment]
This is interesting, though I'm not sure which approach is best.  My first thought is that for a semigroup `S`, calling `MyFunction(S)` should always return the same thing as `MyFunction(Elements(S))`.  This is incompatible with the idea of having
```
DigraphOfActionOnPairs(coll, n) = DigraphOfActionOnPairs(Semigroup(coll), n)
```
which I think would have the potential for confusion.

The safest thing would be to have different names for the two behaviours, though we might need shorter names.
Something like `PairsDigraphOfTransformationSemigroup` and `PairsDigraphOfTransformationColl`, perhaps?

james-d-mitchell:
@mtorpey @wilfwilson I think I'd go for making this just an attribute of a semigroup, and scrapping the other method. We could scrap the method `IsSynchronisingTransformationCollection` too

wilfwilson:
I agree with this approach

james-d-mitchell:
@wilfwilson great, can you update the PR accordingly?

wilfwilson:
Of course, I should be able to complete it at the Wednesday meeting.

wilfwilson:
@james-d-mitchell I've made some progress on this. I'll let you know when I'm finished.

wilfwilson:
This PR is ready to be merged in my opinion.

PR #246, opened 2017-01-19
Improve Idempotents by rank for a transformation semigroup

wilfwilson:
This addresses a FIXME in the `semitrans.gi` file.

Instead of calculating all the idempotents and then filtering them for the ones of correct rank, we now only enumerate those idempotents which are in a regular D-class of the appropriate rank.

### DIFF
--- a/gap/semigroups/semitrans.gi
+++ b/gap/semigroups/semitrans.gi
@@ -328,17 +328,30 @@ SEMIGROUPS.GraphOfRightActionOnPairs := function(gens, n, stop_on_isolated_pair)
 end;
 
 # FIXME can probably do better than this
+# WW has improved this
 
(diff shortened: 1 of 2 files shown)


gap/semigroups/semitrans.gi
```diff
@@ -328,17 +328,30 @@ SEMIGROUPS.GraphOfRightActionOnPairs := function(gens, n, stop_on_isolated_pair)
 end;
 
 # FIXME can probably do better than this
+# WW has improved this
```
james-d-mitchell: If this really improves the performance, then you could remove the two comments before this method. In particular, please remove the FIXME.

wilfwilson:
I'll just get rid of this PR. I didn't notice that this was the generic method and that there's already a better method for acting semigroups.

PR #248, opened 2017-01-19
Add operation to convert a partial perm bipartition to a block bijection

wilfwilson:
This addresses a card from Trello.

A partial perm bipartition naturally defines a partial perm, and a partial perm can be converted into a block bijection in a nice way (via a monoid embedding). Thus we should be able to convert partial perm bipartitions to block bijections in this way.

I believe that the resulting conversion is indeed a monoid embedding.

### DIFF
--- a/doc/bipart.xml
+++ b/doc/bipart.xml
@@ -433,8 +433,8 @@ the argument does not define a partial perm,]]></Example>
 
       If the optional second argument <A>n</A> is not present, then the maximum
       of the degree and codegree of <A>x</A> plus 1 is used by default. If the
-      second argument <A>n</A> is not greater than this maximum, then
-      <K>fail</K> is returned.
+      second argument <A>n</A> is not greater than this maximum, then an Error
+      is returned.
       <P/>
 
       This is the value at <A>x</A> of the embedding of the symmetric inverse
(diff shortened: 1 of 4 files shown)


doc/bipart.xml
```diff
@@ -433,29 +433,39 @@ the argument does not define a partial perm,]]></Example>
 
       If the optional second argument <A>n</A> is not present, then the maximum
       of the degree and codegree of <A>x</A> plus 1 is used by default. If the
-      second argument <A>n</A> is not greater than this maximum, then
-      <K>fail</K> is returned.
+      second argument <A>n</A> is not greater than this maximum, then an Error
```
james-d-mitchell: Please change to lower case "Error", and is an error not "given", or "then an error will occur." It isn't really returned.

wilfwilson: I agree.

wilfwilson:
@james-d-mitchell I've changed the documentation as requested.

I saw no reason for there to exist a method to allow `AsBlockBijection(x, 0)` for a partial perm `x`.

The main reason is that the existence of such a method contradicts the documentation, both as it existed previously, and as it exists now in this PR. The documentation previously stated:

> If the optional second argument <A>n</A> is not present, then the maximum
> of the degree and codegree of <A>x</A> plus 1 is used by default. If the
> second argument <A>n</A> is not greater than this maximum, then
> <K>fail</K> is returned.

(My PR changes this a bit - when the second argument is too small then the method gives an error rather than returning `fail`. It seems to be our policy to do this kind of thing in the package.)

The documentation suggests that the second argument is certainly never allowed to be 0: the (co)degree of a partial perm is at least 0, and so max{degree, codegree} + 1 is at least 1. Thus the documented behaviour of `AsBlockBijection(x, 0)` is to return `fail` for *all* partial perms `x`. So the method which returns the empty bipartition for any partial perm when the second argument is `0` contradicts this defined behaviour.

Furthermore, I would argue that having an exception for `0` is just weird: it breaks the "continuity" of the function. Suppose that `x` is a partial perm of degree 2 and codegree 3.
Then `AsBlockBijection(x)` is the same as `AsBlockBijection(x, 4)`; and happily accepts any higher second argument. When given a second argument that is a positive integer lower  than `4`, the method returns `fail`. But... there's then the case that `AsBlockBijection(x, 0)` returns the empty bipartition? Seeing as `1`, `2`, and `3` don't work, I wouldn't expect `0` to either.

PR #249, opened 2017-01-19
Fix IsSemigroupWithCommutingIdempotents

wilfwilson:
Previously `IsSemigroupWithCommutingIdempotents` was a synonym for `IsBlockGroup`. This was mathematically incorrect.

A semigroup with commuting idempotents is necessarily a block group, but the converse is not necessarily true.

### DIFF
--- a/doc/properties.xml
+++ b/doc/properties.xml
@@ -94,12 +94,10 @@ true]]></Example>
 <#GAPDoc Label="IsBlockGroup">
   <ManSection>
     <Prop Name = "IsBlockGroup" Arg = "S"/>
-    <Prop Name = "IsSemigroupWithCommutingIdempotents" Arg = "S"/>
     <Returns><K>true</K> or <K>false</K>.</Returns>
     <Description>
-      <C>IsBlockGroup</C> and <C>IsSemigroupWithCommutingIdempotents</C> return
-      <K>true</K> if the semigroup <A>S</A> is a block group and <K>false</K>
-      if it is not.<P/>
+      <C>IsBlockGroup</C> returns <K>true</K> if the semigroup <A>S</A> is a
+      block group and <K>false</K> if it is not.<P/>
 
       A semigroup <A>S</A> is a <E>block group</E> if every &L;-class and every
       &R;-class of <A>S</A> contains at most one idempotent. Every semigroup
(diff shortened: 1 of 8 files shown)


doc/properties.xml
```diff
@@ -846,6 +844,40 @@ true]]></Example>
...
+    <Returns><K>true</K> or <K>false</K>.</Returns>
+    <Description>
+      <C>IsSemigroupWithIdempotents</C> returns <K>true</K> if the idempotents
+      of the semigroup <A>S</A> commute, and <K>false</K> if they do not. Note
+      that such an semigroup is also a block group; see <Ref
+        Prop="IsBlockGroup"/>. <P/>
+
+      The idempotents of every inverse semigroup commute, and the idempotents
+      commute in any semigroup of partial permutations and any semigroup of
+      block bijections commute.
```
james-d-mitchell: Is there an extra `commute` here?

gap/attributes/attr.gi
```diff
@@ -407,6 +407,9 @@ end);
 InstallMethod(StructureDescription, "for a group as semigroup",
 [IsGroupAsSemigroup],
 function(S)
+  if IsGroup(S) then
+    TryNextMethod();
+  fi;
   return StructureDescription(Range(IsomorphismPermGroup(S)));
 end);
```
james-d-mitchell: Shouldn't this whole chunk be in a separate commit/PR since it doesn't have anything to do with this change? Or does it?

doc/properties.xml
wilfwilson: Yes.

gap/attributes/attr.gi
wilfwilson: This chunk is relevant - without it, the tests fail.

I added several `InstallTrueMethod`s which concern `IsSemigroupWithCommutingIdempotents`. One of them means that `RankFilter(IsGroupAsSemigroup)` is now higher than `RankFilter(IsGroup)` is at the point that GAP is loaded (and the default `StructureDescription` method for `IsGroup` is installed). Thus, for any group in the category `IsGroup`, method selection now chooses this `IsGroupAsSemigroup` method for `StructureDescription`... which calls itself, ad infinitum. Many tests were failing after encountering such a recursion depth trap.

The `TryNextMethod` avoids this recursion.

wilfwilson:
@james-d-mitchell I've made the change to the documentation. Since the change to `StructureDescription` fixes tests that would otherwise be broken by the other changes, I firmly believe that they belong in the same commit.

Issue #253, opened 2017-02-17
Bugs with IsIdempotentGenerated


wilfwilson:
I've found a couple of problems with the code for `IsIdempotentGenerated`.

The main method uses `Generators(S)` rather than `GeneratorsOfSemigroup(S)`, so for ideals the incorrect answer can be returned: the generators of the ideal being idempotents has little relation to whether the semigroup itself is idempotent generated:
```
gap> S := Semigroup([Transformation([3, 2, 1]), Transformation([2, 2, 2])]);;
gap> IsIdempotentGenerated(S); # false
gap> I := SemigroupIdeal(S, IdentityTransformation);
gap> IsIdempotentGenerated(I); # true
gap> I = S; # true
```
The method for RZMS only works if the `UnderlyingSemigroup` is a group, and this is currently assumed:
```
gap> mat := [[
>  Transformation([2, 3, 1]),
>  Transformation([2, 1]),
>  Transformation([1, 2, 1])]];;
gap> R := ReesZeroMatrixSemigroup(FullTransformationMonoid(3), mat);
gap> IsIdempotentGenerated(R); # causes error
```
I'm working on a fix for these problems, but I thought it would be good to create an issue to help us with bug tracking. The first problem is present in `2.8`, but in `2.8` there is no method specified method installed which applies to Rees 0-matrix semigroups, and so the second bug only exists in the `unstable-3.0` branch.

mtorpey:
This is fixed by pull request #254.

PR #254, opened 2017-02-17
Fix Issue #253 (IsIdempotentGenerated)

wilfwilson:
This PR addresses Issue #253 and makes some small improvements.

The default `IsIdempotentGenerated` method, given a semigroup `S`, first checks where the "generators" of `S` are idempotents. That's a good idea for semigroups and semigroup generating sets, but for ideals it was using the ideal generating set. Therefore I've replaced `Generators` by `GeneratorsOfSemigroup`.

I've added another check here, for monoids. The only elements of a monoid which can be multiplied together to produce an element of the group of units are themselves units. Since the only unit which is an idempotent is the identity, and since the subsemigroup generated by the identity is trivial, the idempotent generated subsemigroup of a monoid has a trivial group of units. Therefore, for a monoid to be idempotent generated, its group of units must be trivial.

When none of the easy cases give an answer, the method then creates the idempotent generated subsemigroup, or, when the semigroup is an acting semigroup, a (possibly smaller) idempotent generated subsemigroup, in order to check whether the generators of `S` are contained in this subsemigroup. This code called `Idempotent(S)` and then filtered them by rank. In the case that we know the idempotents already, that's fine. However, my experiments have shown that when the idempotents are not known, it's best to find the idempotents of the particular ranks that we want, using `Idempotents(S, rank)` for each appropriate rank.

Also, for this special code which applies only to acting semigroups, I now create this idempotent generated subsemigroup as an acting semigroup. This allows me to include examples such as
 ```IsIdempotentGeneratedSemigroup(Semigroup(FullTransformationMonoid(9), rec(acting := true)));```
in the tests.

Finally, I fixed the issues that I found with `IsIdempotentGenerated` for a RZMS (it gave an error when the underlying semigroup was not `IsGroup`; it now works for the more general `IsGroupAsSemigroup` but properly tries the next method otherwise), and I've introduced an analogous method for RMS's.

I made a related change: the idempotent generated subsemigroup of a semigroup now knows that it is idempotent generated, and the idempotent generated subsemigroup of an inverse semigroup now knows that it is a semilattice.

### DIFF
--- a/gap/attributes/properties.gi
+++ b/gap/attributes/properties.gi
@@ -594,13 +594,18 @@ end);
 InstallMethod(IsIdempotentGenerated, "for a semigroup",
 [IsSemigroup],
 function(S)
-  local gens, T, min, new;
+  local gens, T, ranks, min, new, max, i;
 
   if not IsFinite(S) then
     TryNextMethod();
   fi;
 
-  gens := Generators(S);
+  if HasIsMonoidAsSemigroup(S) and IsMonoidAsSemigroup(S)
+      and not IsTrivial(GroupOfUnits(S)) then
+    return false;
+  fi;
+
+  gens := GeneratorsOfSemigroup(S);
 
   if ForAll(gens, IsIdempotent) then
     Info(InfoSemigroups, 2, "all the generators are idempotents");
@@ -610,12 +615,21 @@ function(S)
   if HasIdempotentGeneratedSubsemigroup(S) or not IsActingSemigroup(S) then
     T := IdempotentGeneratedSubsemigroup(S);
   else
-    min := MinimumList(List(gens, x -> ActionRank(S)(x)));
-    new := Filtered(Idempotents(S), x -> ActionRank(S)(x) >= min);
+    ranks := List(gens, x -> ActionRank(S)(x));
+    min := MinimumList(ranks);
+    if HasIdempotents(S) then
+      new := Filtered(Idempotents(S), x -> ActionRank(S)(x) >= min);
+    else
+      max := MaximumList(ranks);
+      new := [];
+      for i in [min .. max] do
+        Append(new, Idempotents(S, i));
+      od;
+    fi;
     if new = [] then
       return false;
     fi;
-    T := Semigroup(new);
+    T := Semigroup(new, rec(acting := true));
   fi;
 
   # T is not always the idempotent generated subsemigroup!
@@ -1441,7 +1455,7 @@ function(S)
   elif not IsFinite(S) then
     TryNextMethod();
   fi;
-  return ForAll(GreensDClasses(S), IsTrivial);
+  return IsDTrivial(S);
 end);
 
 InstallMethod(IsSimpleSemigroup, "for a finite semigroup",
--- a/gap/semigroups/semirms.gi
+++ b/gap/semigroups/semirms.gi
@@ -898,6 +898,21 @@ function(R)
   return Semigroup(Filtered(MatrixEntries(N), x -> x <> 0)) = U;
 end);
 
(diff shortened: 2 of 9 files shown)


gap/semigroups/semirms.gi
```diff
@@ -881,16 +881,21 @@ function(R)
...
-  RR := Range(RZMSNormalization(R));
-  return Group(Filtered(MatrixEntries(RR), x -> x <> 0)) =
-    UnderlyingSemigroup(RR);
+  U := UnderlyingSemigroup(R);
+  if not IsGroupAsSemigroup(U) then
+    TryNextMethod();
+  fi;
+  N := Range(RZMSNormalization(R));
+  return Semigroup(Filtered(MatrixEntries(N), x -> x <> 0)) = U;
 end);
```
james-d-mitchell: Just to double check, since you are replacing a more specific method with a more general one (which is good, of course!) is there any loss of performance in the case that the more specific method applies?

gap/semigroups/semirms.gi
```diff
@@ -881,16 +881,21 @@ function(R)
...
+    TryNextMethod();
+  elif not IsConnectedDigraph(RZMSDigraph(R)) then
     return false;
   fi;
-  RR := Range(RZMSNormalization(R));
-  return Group(Filtered(MatrixEntries(RR), x -> x <> 0)) =
-    UnderlyingSemigroup(RR);
+  U := UnderlyingSemigroup(R);
+  if not IsGroupAsSemigroup(U) then
+    TryNextMethod();
```
james-d-mitchell: I guess there is not next method at present, right? If so, can you add a comment to that effect, please?

gap/attributes/properties.gi
```diff
@@ -610,12 +615,21 @@ function(S)
...
+      new := [];
+      for i in [min .. max] do
+        Append(new, Idempotents(S, i));
+      od;
+    fi;
     if new = [] then
       return false;
     fi;
-    T := Semigroup(new);
+    T := Semigroup(new, rec(acting := true));
```
james-d-mitchell: What's the purpose of the `acting := true` here? What if `S` is not an acting semigroup (for some reason determined by the user)?

james-d-mitchell: Oh wait I see, we wouldn't be in this case if `S` is not acting.

james-d-mitchell: Ok, I still don't really see why this is necessary, can you explain? Won't `T` automatically have `acting := true` because the `S` does?

gap/semigroups/semirms.gi
wilfwilson: The next method is the usual one in `gap/attributes/properties` which (basically) generates the idempotent generated semigroup and compares it to the semigroup itself. It's the one I've made a couple of changes to below that you've commented on.

gap/attributes/properties.gi
wilfwilson: I see your issue. I did it because of the tests. I wanted to check `IsIdempotentGeneratedSubsemigroup` on a large semigroup which is something like `T10`. The acting code can handle this completely fine, whereas the generic code struggles.

Therefore in the test file I create this example using the `acting := true` option. This is fine when the acting methods are on. However, when we run the tests with `SEMIGROUPS.Test`, in its second pass it sets the options record to have `acting := false`. So although I create the semigroup in the test file as an acting semigroup, and although it happily gets to the line containing `T := Semigroup(new);`, this new semigroup `T` is a generic semigroup which is too big to be computed. Therefore I force it to be acting.

james-d-mitchell:
Fair enough

mtorpey:
Looks good to me too!

Issue #171, opened 2016-05-05
Add code-coverage checks and gaplint to travis


james-d-mitchell:
It would be good to add code-coverage and gaplint checks to travis. Not sure how complicated this would be, but its been on my mind for ages, so I thought I'd add an issue for it.

PR #257, opened 2017-02-18
Use gaplint and code-coverage in the Travis tests

wilfwilson:
In Issue #171, @james-d-mitchell suggests adding code-coverage checking and lint checking to out Travis setup. I think these are both good intentions.

### Linting

After this PR, Travis downloads the gaplint tool. After running the normal GAP tests and the saved workspace tests, Travis now runs a linting check. **We should discuss what the linter should be run on, and what we should do when we encounter trouble.**

Currently I've got it doing `gaplint.py gap/*/*.gi`, and failing the Travis instance if _anything_ goes wrong. This checks the linting of the `.gi` files which reside in a subdirectory of the `gap` subdirectory of the package.

Today I spent a good chunk of time linting the Semigroups package and fixing the gaplint tool so that, at least in `unstable-3.0`, running `gaplint.py gap/*/*.gi` causes no warnings or errors or crashes. I'm happy with this change, as it means that from now on, we will have to make sure that we write well-formatted code in the first instance, under pain of breaking the tests.

However I am wary about extending the scope of the linter at the moment, at least in a way that would cause more Travis failures.

* in my experience, gaplint is not totally robust (no offence James), and since it is quite a complicated tool, I envisage a common occurrence: new code causes gaplint to crash or behave incorrectly, which means that the Travis tests fail. The tests will keep failing until gaplint is fixed, which renders Travis essentially useless in this interim period.
* when run on `.gd` files, Travis will complain about missing documentation, and things being in the wrong files, etc. Obviously we don't want missing documentation, but I'm not convinced that this is robust enough yet to give **no** false positives (surely we don't want false positives breaking the Travis tests). **Also, often we want to push something to our own repositories which isn't complete, and so maybe it lacks documentation, but we still want to be able to rely on Travis to tell us whether our branch is working or not.**
* when run on `.gd` files at the moment, a _lot_ goes wrong. If we started checking these files now, it would require a heck of a lot of work before the Travis tests passed again.

For these reasons, I'd argue that we should use gaplint in this more limited sense, perhaps until the release of 3.0.

### Code coverage

After this PR, Travis installs the profiling package. However, I currently only do this when GAP is compiled in 64-bit mode, since compilation seems to fail in 32-bit.

I have also adapted the code coverage script to make a version for Travis. After the linting checks, this coverage script is run on each test file in `semigroups/tst/standard`, and the coverage of its corresponding `.gi` file is found from the html output. If this coverage is below 95%, I print a warning. However, I've not yet set this up to cause the test instance to fail.

We should decide what the threshold should be, and whether we should just print warnings, or whether we should actually fail the tests.

## Your thoughts?

@james-d-mitchell @mtorpey what are your opinions on the lint checking that this PR proposes?

### DIFF
--- a/scripts/travis-coverage.py
+++ b/scripts/travis-coverage.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+
+import tempfile, subprocess, sys, os, ntpath, re
+from os.path import exists, isdir, isfile
+
+_ERR_PREFIX = '\033[31mtravis-coverage.py: error: '
+_WARN_PREFIX = '\033[31mtravis-coverage.py: warning: '
+_BOLD_PREFIX = '\033[1m'
+_BLUE_PREFIX = '\033[34m'
+
+f = str(sys.argv[1])
+if not (exists(f) and isfile(f)):
+    sys.exit(_ERR_PREFIX + f + ' does not exist!\033[0m')
+
+_DIR = tempfile.mkdtemp()
+_COMMANDS = 'echo "Test(\\"' + f + '\\");;\n'
+_COMMANDS += '''UncoverageLineByLine();;
+LoadPackage(\\"profiling\\", false);;
+filesdir := \\"''' + os.getcwd() + '''/gap/\\";;\n'''
+_COMMANDS += 'outdir := \\"' + _DIR + '\\";;\n'
+_COMMANDS += 'x := ReadLineByLineProfile(\\"' + _DIR + '/profile.gz\\");;\n'
+_COMMANDS += 'OutputAnnotatedCodeCoverageFiles(x, filesdir, outdir);"'
+
+pro1 = subprocess.Popen(_COMMANDS, stdout=subprocess.PIPE, shell=True)
+_RUN_GAP = '../../bin/gap.sh -A -q -r -m 1g -T --cover ' + _DIR + '/profile.gz'
+
+try:
+    pro2 = subprocess.Popen(_RUN_GAP,
+                            stdin=pro1.stdout,
+                            shell=True)
+    pro2.wait()
+except KeyboardInterrupt:
+    pro1.terminate()
+    pro1.wait()
+    pro2.terminate()
+    pro2.wait()
+    sys.exit('\033[31mKilled!\033[0m')
+except (subprocess.CalledProcessError, IOError, OSError):
+    sys.exit(_ERR_PREFIX + 'Something went wrong calling GAP!\033[0m')
+
+filename = _DIR + '/index.html'
+if not (exists(filename) and isfile(filename)):
+    sys.exit('\n' + _ERR_PREFIX + 'Failed to find file://' + filename + '\033[0m')
+
+gi_file = ntpath.basename(f).split('.')[0] + '.gi'
+for line in open(filename):
+    if gi_file in line:
+        break
+
+search = re.search('coverage\d\d+[\'"]>(\d+)</td><td>([\d,]+)</td><td>([\d,]+)</td>', line)
+if search == None:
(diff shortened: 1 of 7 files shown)


wilfwilson:
After discussions with @james-d-mitchell, I have made the following changes:

* gaplint no longer causes the Travis tests to fail. It is useful to be able to look on Travis and see the state of the linting, so we may as well keep including it for now. However, we feel that gaplint is not sufficiently robust to rely on. If gaplint ever stops working we will simply remove it.
* if any `x.gi` file has code coverage < 95% (when only the file `x.tst` has been run) then the Travis tests will now fail. Perhaps in future this number can be increased further, since many (all?) of the old deficiencies in the profiling package (which made it very hard to reach 100%) seem to have been resolved.
* certain tst files can be excluded by adding them to the new `.covignore` file. Currently this contains only `idealenum.tst`.

As you can see, the Travis tests currently fail. This is because of the following files:

* `congpairs.gi` has 93% coverage: 218/233 lines
* `cong.gi` has 94% coverage: 243/256 lines
* `gracreg.gi` has 94% coverage: 164/173 lines
* `grpffmat.gi` has 92% coverage: 96/104 lines
* `obsolete.gi` has 85% coverage: 47/55 lines
* `semiact.gi` has 92% coverage: 478/515 lines
* `semipbr.gi` has 92% coverage: 46/50 lines
* `semipperm.gi` has 92% coverage: 487/524 lines
* `setup.gi` has 13% coverage: 51/367 lines

`setup.gi` is the odd one out here. For the others, I'll have a go right now at improving the coverage. Any thoughts @mtorpey?

wilfwilson:
@james-d-mitchell I've now taken `setup.gi` from 13% to 100% code coverage. This means that all standard test files (apart from `idealenum.tst`, which is explicitly excluded for now), when run independently, have at least 95% code coverage of their corresponding `.gi` files.

In my opinion, this PR is ready to be merged.

wilfwilson:
It turns out that the newest version of the profiling package, `1.2.0`, does indeed work on 32-bit (`1.1.0` is the one which is listed on gap-system.org, which is why I had been using that). Thus I've managed to make the code coverage checks work on both 32- and 64-bit Travis builds.

The only problem is that whilst the code coverage checks take about an additional 6 minutes in 64-bit, they take an additional 14 minutes in 32-bit, giving an overall time of 17 minutes versus 25 minutes for the 64-bit vs. 32-bit build.

If we're bothered about this, then I can exclude the coverage checks from 32-bit (could there be any reason why the coverage would be different across the different builds?). Otherwise, we can leave it as it is and the builds will just take longer than before.

james-d-mitchell:
@wilfwilson great work! I'd probably vote for not including the code coverage in 32-bit, just to save a bit of time. It is probably only necessary to check code coverage one time in the travis tests. If you make that change I'll merge.

scripts/travis-coverage.py
```diff
@@ -0,0 +1,61 @@
...
+
+pro1 = subprocess.Popen(_COMMANDS, stdout=subprocess.PIPE, shell=True)
+_RUN_GAP = '../../bin/gap.sh -A -q -r -m 1g -T --cover ' + _DIR + '/profile.gz'
+
+try:
+    pro2 = subprocess.Popen(_RUN_GAP,
+                            stdin=pro1.stdout,
+                            shell=True)
+    pro2.wait()
+except KeyboardInterrupt:
```
james-d-mitchell: Not completely sure that this line is required for a script running on travis, but it won't do any harm either 👍

wilfwilson: Of course! I included it in case I ever want to run the script locally on my computer (which I have done and I've used the `KeyboardInterrupt`!).

wilfwilson:
@james-d-mitchell I've removed the 32-bit coverage checks. All seems to be fine now.

PR #258, opened 2017-02-22
Congruences over FP semigroups and monoids

mtorpey:
This pull request modifies the existing implementation of congruences by generating pairs, to allow for congruences over free semigroups and finitely presented semigroups (which may be infinite).  This is primarily to allow for the fixing of Issue #191; the corresponding GAP test is added to `testinstall.tst`.

We also change the implementation of FP semigroups in `semifp.gi` to use congpairs objects on the GAP level, instead of interfacing directly with C in the `semifp.cc` file (which is removed).  We also remove `congfree.gd/gi`, a temporary hack which is no longer necessary.

All of this relies on changes to libsemigroups that are made in [this pull request](https://github.com/james-d-mitchell/libsemigroups/pull/10).  Hence, until that pull request is merged, **the Travis tests for this PR will fail**.


wilfwilson:
@mtorpey I'd just like to point out that currently in `unstable-3.0`, the files `cong.gi` and `cong-pairs.gi` each have 100% code coverage, [as seen on Travis ](https://travis-ci.org/gap-packages/Semigroups/jobs/207439590). Although Travis accepts any coverage above 95%, I would recommend you aim for 100% coverage (unless there's a good reason you've added new unreachable code). You'll be able to see the coverage by looking at the Travis tests once the libsemigroups PR is merged, or by running the code coverage tool on each of the files `cong.tst` and `cong-pairs.tst` independently.

mtorpey:
This should now work with the latest libsemigroups PR (batch processing).

Note that Travis still uses `master` for libsemigroups, so this will fail until that PR is merged in.

wilfwilson:
@mtorpey Your libsemigroups PR has now been merged! So now we can work on getting this merged too.

gap/congruences/cong.gi
```diff
@@ -494,3 +494,18 @@ _GenericCongCanonicalLookup);
 
 MakeReadWriteGlobal("_GenericCongCanonicalLookup");
 Unbind(_GenericCongCanonicalLookup);
+
+InstallMethod(EquivalenceRelationCanonicalPartition,
```
wilfwilson: Similar comment as before. This method is exactly the same as for Left/RightSemigroupCongruences. In my opinion, therefore this method can be deleted. Since a (two-sided) congruence is both a left and a right congruence, one of those (identical) methods will be selected instead. You could add a comment line stating that for two-sided congruences, the left/right versions apply.

gap/congruences/cong.gi
```diff
@@ -21,46 +21,49 @@
...
+BindGlobal("_GenericCongEquality",
+function(cong1, cong2)
+  local S;
+  if Range(cong1) <> Range(cong2) then
+    return false;
   fi;
-  return Range(c1) = Range(c2)
-         and EquivalenceRelationCanonicalLookup(c1) =
-             EquivalenceRelationCanonicalLookup(c2);
+  S := Range(cong1);
```
wilfwilson: 100% pedantic, I admit it, but you could put this line `S := Range(cong1)`; before the `if` statement, so that it then reads `if S <> Range(cong2) then`. It saves `Range(cong1)` having to be called twice, which I understand isn't a big overhead, but it would make me happier :).

gap/congruences/cong.gi
```diff
@@ -21,46 +21,49 @@
...
 
 InstallMethod(\=, "for a right and a left semigroup congruence",
 [IsRightSemigroupCongruence, IsLeftSemigroupCongruence],
+_GenericCongEquality);
+
+InstallMethod(\=, "for two right semigroup congruences",
+[IsRightSemigroupCongruence, IsRightSemigroupCongruence],
+_GenericCongEquality);
+
+# Since two-sided congs are both left and right, this covers all cases
```
wilfwilson: Cool, this is the sort of thing I was commenting on in the previous commit.

gap/congruences/congpairs.gi
```diff
@@ -442,8 +477,27 @@ function(cong1, cong2)
...
+"for dense list and enumerable semigroup congruence",
+[IsDenseList, IsEnumerableSemigroupCongruence],
+function(pair, cong)
+  local S;
+  S := Range(cong);
+  if Size(pair) <> 2 then
+    ErrorNoReturn("Semigroups: \\in (for a congruence): usage,\n",
+                  "the first arg <pair> must be a list of length 2,");
+  elif not (pair[1] in S and pair[2] in S) then
+    ErrorNoReturn("Semigroups: \\in (for a congruence): usage,\n",
```
wilfwilson: I'm not sure (genuinely not sure) what the right behaviour of `\in` should be. Here, if the pair of elements aren't both in the semigroup, then obviously the pair isn't in the congruence. But should this trigger an error, or return false?

Compare with the following:

```
gap> S := FullTransformationMonoid(1);;
gap> EmptyPartialPerm() in S;
false
```
The empty partial perm is a partial perm, and `S` is a transformation semigroup, so obviously it's not in `S`. But this returns `false` rather than an error.

It's not something I care deeply about, but I think it's something worth thinking about, especially for the sake of consistency with the rest of the package and GAP.

tst/standard/cong.tst
```diff
@@ -376,21 +376,21 @@ gap> EquivalenceRelationCanonicalLookup(ccong);
...
-Error, no 2nd choice method found for `ImagesSet' on 2 arguments
-gap> cong2 = cong1;
-Error, no method found! For debugging hints type ?Recovery from NoMethodFound
-Error, no 2nd choice method found for `ImagesSet' on 2 arguments
+# These examples would run forever trying to calculate EquivalenceRelationPartition
+#gap> F := FreeSemigroup(2);;
+#gap> cong1 := ReesCongruenceOfSemigroupIdeal(SemigroupIdeal(F, [F.1]));;
+#gap> cong2 := SemigroupCongruence(F, [F.1, F.1^2]);;
+#gap> cong1 = cong2;
+#false
```
wilfwilson: Is it a deliberate choice to make this regression? Previously, I presume, we could ask whether `cong1 = cong2` and get an answer; and now it will never terminate. How was it terminating before? If it was because it was using a GAP method, isn't this kind of issue the whole reason that this PR exists?

src/congpairs.cc
```diff
@@ -31,6 +31,7 @@ using libsemigroups::Congruence;
 using libsemigroups::word_t;
 using libsemigroups::relation_t;
 using libsemigroups::RecVec;
+using libsemigroups::Partition;
```
wilfwilson: Is this supposed to be part of this commit? Just curious why it's been added (I'm obviously not an expert on this kind of stuff).

mtorpey: It's used in the very next commit, in `CONG_PAIRS_NONTRIVIAL_CLASSES`.  Maybe I'll move this change to that commit to make this more readable.

gap/congruences/congpairs.gi
mtorpey: This is a good point, and one James and I have talked about before.  There's a good case for `\in` just returning `false`, but we decided at some point that it would be better to have an error message (since the question of "in" doesn't really make sense if it's not a pair in the semigroup).  For consistency with congruences in the rest of the package, I've continued to implement `\in` in this way here.

gap/semigroups/semifp.gi
```diff
@@ -486,3 +486,18 @@ IsCollsElms, [IsFpSemigroup, IsElementOfFpSemigroup],
 function(S, x)
   return SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(x));
 end);
+
+InstallMethod(MultiplicativeZero,
+"for an fp semigroup",
+[IsFpSemigroup],
```
mtorpey: I've followed all this advice.

src/congpairs.cc
```diff
@@ -175,38 +218,42 @@ Obj CONG_PAIRS_NR_CLASSES(Obj self, gap_cong_t o) {
...
+    lhs = plist_to_word_t(ELM_LIST(pair, 1));
+    rhs = plist_to_word_t(ELM_LIST(pair, 2));
   } else {
-    gap_rec_t  data  = fropin(S, INTOBJ_INT(-1), 0, False);
-    gap_list_t words = ElmPRec(data, RNam_words);
+    gap_semigroup_t S = cong_obj_get_range_obj(o);
+    size_t lhs_pos    = INT_INTOBJ(EN_SEMI_POSITION(0L, S, ELM_LIST(pair, 1)));
+    size_t rhs_pos    = INT_INTOBJ(EN_SEMI_POSITION(0L, S, ELM_LIST(pair, 2)));
+
+    // FIXME check if l/rhs_pos are undefined!!
```
mtorpey: Fixed.

doc/cong.xml
```diff
@@ -437,6 +437,37 @@ gap> EquivalenceRelationCanonicalLookup(cong);
...
 <#/GAPDoc>
 
+<#GAPDoc Label="EquivalenceRelationCanonicalPartition">
+  <ManSection>
+    <Attr Name = "EquivalenceRelationCanonicalPartition" Arg = "cong"/>
+    <Returns>A list of lists.</Returns>
+    <Description>
+      This attribute returns a list of lists of elements of the underlying set
+      of the semigroup congruence <A>cong</A>.  These lists are precisely the
+      nonsingleton equivalence classes of <A>cong</A>.  The order in which the
```
mtorpey: Done.

gap/congruences/cong.gd
```diff
@@ -21,6 +21,11 @@
 ## Some general functions are also implemented in cong.gi
 ##
 
+# These are all equivalence relations
+InstallTrueMethod(IsEquivalenceRelation, IsSemigroupCongruence);
```
mtorpey: Quite right.  Removed.

gap/congruences/cong.gi
mtorpey: Yup.

gap/congruences/cong.gi
mtorpey: Hadn't spotted that.  I've now changed it.

tst/standard/cong.tst
mtorpey: I removed this for valid reasons, which I've now actually explained in the test file.  Basically, this test was invalid and was passing because of luck.  My new version makes this much clearer.

mtorpey:
I've now redone this pull request, taking on board all @wilfwilson's feedback, improving code coverage, linting, valgrinding, and making a few things in the code clearer.  Thank you for all the useful comments, Wilf!

The only thing I haven't finished yet is an implementation of congruences on fp *monoids* - this is basically done, but is currently failing for some bizarre reason which I've distilled into Issue #267.  If anyone can help with that, I'd be very grateful.  However, since the fp monoid thing is reasonably separate, I'd be happy to submit it in a separate PR once it's working.

Enjoy!

src/congpairs.cc
wilfwilson: Fine, just wanted to make sure it wasn't some extraneous code that you'd forgotten to delete.

gap/congruences/congpairs.gi
wilfwilson: I won't argue further. I think consistency is the most important thing here.

wilfwilson:
I'm happy with this.

mtorpey:
Now rebasing onto the most recent master...

mtorpey:
Now adding a few more changes I noticed needed done...

mtorpey:
And finally, adding the stuff for fp monoids...

james-d-mitchell:
@mtorpey you'll have to rebase this onto master again, sorry. Also looks like one of the tests is failing.

gap/semigroups/semitrans.gi
```diff
@@ -797,6 +797,42 @@ function(S)
...
+
+  iso := function(x)
+    return EvaluateWord(gen, MinimalFactorization(S, x));
+  end;
+
+  inv := function(x)
+    return EvaluateWord(GeneratorsOfSemigroup(S), Factorization(T, x));
+  end;
+
+  #TODO replace this with SemigroupIsomorphismByImagesOfGenerators
```
wilfwilson: Is this a TODO that you can resolve now?

mtorpey: This TODO reflects a similar TODO in the method for enumerable semigroups.  I don't understand why it's there, but I imagine something similar will want to be done here.

mtorpey: To clarify: as far as I know, the code is fine as it is.

wilfwilson:
I've now looked at the commits about fp-monoids, and I'm happy for it all to be merged.

PR #260, opened 2017-02-23
Remove duplicated UF code

mtorpey:
The union-find class `UF` is now in libsemigroups, so we don't need it in Semigroups any more.  This removes the class, as well as renaming all instances of `ufdata` to simply `uf`, to be more consistent with what we did in libsemigroups.


wilfwilson:
Seems good to me.

PR #261, opened 2017-02-23
Lattice of congruences: info warning

mtorpey:
Finding all the congruences of a semigroup can take a long time, except for very small semigroups.  This PR sends an info warning to inform the user that this is the case - full disclosure!

### DIFF
--- a/gap/congruences/conglatt.gi
+++ b/gap/congruences/conglatt.gi
@@ -61,6 +61,10 @@ SEMIGROUPS.PrincipalXCongruencePosetNC := function(S, restriction, cong_func)
   local report, pairs, total, congs, nrcongs, children, parents, last_collected,
         nr, pair, badcong, newchildren, newparents, newcong, i, c, p, po, poset;
 
+  # Display warning
+  Info(InfoSemigroups, 1,
+       "Finding a semigroup's congruences can take a very long time!");
+
   # Suppress reporting
   report := SEMIGROUPS.OptionsRec(S).report;
   SEMIGROUPS.OptionsRec(S).report := false;


james-d-mitchell:
Couldn't the warning be a bit more informative? You could mention that it involves processing `Binomial(Size(S), 2)` congruences on `S`?  Also if `S` is trivial, then this will look weird as a warning, wouldn't it be a good idea to only display this if the size of the semigroup exceeds some threshold?

wilfwilson:
I don't support the idea of this kind of warning. Perhaps including these kinds of warnings in the documentation might be a good idea, but this seems a bit weird to me. There are (surely) many places in our code where computation is going to take a long time, such as calculating maximal subsemigroups, and I don't see what's special about lattice of congruences.

james-d-mitchell:
I'm on the fence with this one. Would you still be against this @wilfwilson if there were some more checks about the size of the semigroup (and maybe some properties, like `IsZeroSemigroup`) before  a warning was given?

wilfwilson:
I have softened somewhat on this issue. I wouldn't be upset if you merge.

My main problem is that I think that the package should be consistent. If we're going to warn about possibly-lengthy computations in `LatticeOfCongruences`, then why not for all functions? Pretty much any function in the Semigroups package will take forever if you give it a large enough semigroup. In my opinion, any user of the Semigroups package should have that in mind whilst they use the package. Adding warnings of this kind is not functionality that I'd like to add - it would be too difficult and time consuming, and hard to make accurate.

But yes, if this is to be merged, I agree that the warning should only be displayed if the semigroup exceeds some experimentally-deduced threshold(s).

Issue #264, opened 2017-02-28
MultiplicativeZero for a Rees 0-matrix subsemigroup


wilfwilson:
The method for `MultiplicativeZero` for a Rees 0-matrix subsemigroup wrongly assumes that the only possible multiplicative zero is the special 0 element of the parent Rees 0-matrix semigroup. However this is not true, for example:
```
gap> R := ReesZeroMatrixSemigroup(Group(()), [[()]]);;
gap> U := Semigroup(RMSElement(R, 1, (), 1));;
gap> MultiplicativeZero(U);
fail
gap> IsTrivial(U);
true
```
The multiplicative zero here is the sole element `(1,(),1)`. This should be easy to fix, it'll possibly involve using a `TryNextMethod`.

wilfwilson:
This was resolved by PR #265

PR #265, opened 2017-03-01
Fix issue #264 (MultiplicativeZero for RZMS)

wilfwilson:
This PR resolves issue #264.

Let `R` be a Rees 0-matrix semigroup and let `U` be a subsemigroup of `R`. To check whether `U` had a multiplicative zero, the code previously essentially returned `true` if and only if `U` contained the special `0` element of `R`.

However, there is more to it than this. If `U` doesn't contain the `0` element of `R`, then it still might contain a multiplicative zero. If `U` contains elements from multiple rows or columns of `R`, then (fairly obviously) it can't have a multiplicative zero. I have added code to check this case. Otherwise, when `U` contains elements all from the same row and column `R`, there is more work to be done. I have added a `TryNextMethod` at the end of the function to handle this case - it will use the more general `MultiplicativeZero` methods.

### DIFF
--- a/gap/semigroups/semirms.gi
+++ b/gap/semigroups/semirms.gi
@@ -240,14 +240,27 @@ end);
 InstallMethod(MultiplicativeZero, "for a Rees 0-matrix semigroup",
 [IsReesZeroMatrixSubsemigroup],
 function(R)
-  local rep, x;
-  rep := Representative(R);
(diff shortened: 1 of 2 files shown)


wilfwilson:
Can someone review this please?

james-d-mitchell:
Looks good, thanks @wilfwilson, don't know why I didn't look at this earlier.

PR #266, opened 2017-03-22
Add methods for identities

ffloresbrito:
This pull request adds a chapter, and methods, on verifying identities and operations that relate to these identities. Documentation and tests are included.


james-d-mitchell:
The tests are failing here, but this is probably not your fault @ffloresbrito, we will likely have to wait for @mtorpey to fix somethings in his PR #258 and in libsemigroups, before this will work.

wilfwilson:
@ffloresbrito You should rebase this on the most recent version of `unstable-3.0`, and then it should pass.

mtorpey:
(@ffloresbrito Feel free to call by our office, 119, after PPS if you want some help doing that.)

wilfwilson:
@ffloresbrito We just spoke in the common room about the Travis tests. Here's some quick extra information.

There are several reasons that the Travis tests are now failing:
* linting
* some of your manual examples use random functions, which (obviously) give different answers sometimes, and so the manual examples are failing for that reason
* you were correct that relying on smallsemi might be causing some of the trouble. You are using the smallsemi function `Nr3NilpotentSemigroups`, but smallsemi is not a required package. So the tests are failing, since they cannot find that function. Perhaps @james-d-mitchell is best-placed to suggest a solution to this one.

PR #271, opened 2017-04-25
McAlister triples

ChristopherRussell:
This pull request implements the function `McAlisterTriple` which creates McAlister triple semigroups, along with basic operations for McAlister triples and their elements.

A McAlister triple is a representation of an E-unitary inverse semigroup defined by a partial order and a join semilattice, a group and a group. Amongst several other conditions, the semilattice is a suborder of the partial order and the group acts on them both.

TODO: improve the method for GeneratorsOfSemigroup for these objects in the future.


wilfwilson:
@ChristopherRussell You'll need to update the required version of Digraphs. Your stuff uses things like `IsPartialOrderDigraph`. However these don't exist yet in Digraphs 0.7.1, which is the version required by the Semigroups package currently. To fix this, you should open the file `PackageInfo.g` and replace the occurrences of `0.7.1` with a newer version of Digraphs - I think `0.8.0` should work, although `0.8.1` has been released since then too. Once you've done this, the Travis tests should pass (I hope).

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+
+<#GAPDoc Label="IsMcAlisterTripleSemigroup">
+  <ManSection>
+    <Prop Name = "IsMcAlisterTripleSemigroup" Arg = "S"/>
+    <Returns><K>true</K> or <K>false</K>.</Returns>
+    <Description>
+      A <E>McAlister triple semigroup</E> is an E-unitary inverse semigroup
+      <Ref Oper= "IsEUnitaryInverseSemigroup" BookName= "Semigroups" />
+      defined by a partial order, a semilattice and a group which acts by
+      order automorphisms on the partial order. All E-unitary inverse
```
wilfwilson: I think "Any E-unitary inverse semigroup..." sounds better.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+      BookName= "Digraphs" />, respecitvely. <P/>
+
+      A rigorous treatment of McAlister triples may be found in Chapter 5
+      of <Cite Key = "Howie1995aa"/>. The precise conditions
+      for a triple in &GAP; are as follows:
+
+      <List>
+        <Mark> M1 </Mark>
+        <Item>
+          <C>Y</C> is a semilattice which is a subdigraph of <C>X</C>.
```
wilfwilson: What is X?

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+    <Returns>A McAlister triple semigroup.</Returns>
+    <Description>
+      <C>McAlisterTripleSemigroup</C> requires a digraph <A>X</A> satisfying
+      <Ref Oper= "IsPartialOrderDigraph" BookName= "Digraphs" />, a 
+      <Ref Oper= "Digraph" BookName= "Digraphs" /> <A>Y</A> satisfying
+      <Ref Oper= "IsJoinSemilatticeDigraph" BookName= "Digraphs" />, a group
+      <A>G</A>, and a function <A>act</A> defining an action of <A>G</A> on
+      <A>X</A>. <C>McAlisterTripleSemigroup</C> also has a three argument
+      version which assumes that  <A>act</A> is <Ref Func= "OnPoints"
+      BookName= "GAPDoc"/>. Furthermore there are three and four argument
```
wilfwilson: Shouldn't be `GAPDoc` here. Probably should be `ref`. Change it to ref and see if the warnings go away when you compile the documentation.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+#############################################################################
+##
+
+<#GAPDoc Label="IsMcAlisterTripleSemigroup">
+  <ManSection>
+    <Prop Name = "IsMcAlisterTripleSemigroup" Arg = "S"/>
+    <Returns><K>true</K> or <K>false</K>.</Returns>
+    <Description>
+      A <E>McAlister triple semigroup</E> is an E-unitary inverse semigroup
+      <Ref Oper= "IsEUnitaryInverseSemigroup" BookName= "Semigroups" />
```
wilfwilson: Get rid of the book name bit here. You don't need to do it when you're referring to stuff in the current package. If you've got `BookName = "Semigroups"` elsewhere then get rid of it too.

doc/semieunit.xml
wilfwilson: Change "which is" to "that is"

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+
+      <List>
+        <Mark> M1 </Mark>
+        <Item>
+          <C>Y</C> is a semilattice which is a subdigraph of <C>X</C>.
+        </Item>
+        <Mark> M2 </Mark> 
+        <Item>
+          For all <C>A</C> in <C>X</C> and for all <C>B</C> in <C>Y</C>:
+          if <C>A</C> <M>\leq</M> <C>B</C> then <C>A</C> in <C>Y</C>.
```
wilfwilson: Add a comma before then

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+        </Item>
+        <Mark> M3 </Mark> 
+        <Item>
+          For every <C>A</C> in <C>X</C> there exists some <C>B</C> in
+          <C>Y</C> and <C>g</C> in <C>G</C> such that <C>A</C> is equal to
```
wilfwilson: What is G?

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+          <C>Y</C> is a semilattice which is a subdigraph of <C>X</C>.
+        </Item>
+        <Mark> M2 </Mark> 
+        <Item>
+          For all <C>A</C> in <C>X</C> and for all <C>B</C> in <C>Y</C>:
+          if <C>A</C> <M>\leq</M> <C>B</C> then <C>A</C> in <C>Y</C>.
+        </Item>
+        <Mark> M3 </Mark> 
+        <Item>
+          For every <C>A</C> in <C>X</C> there exists some <C>B</C> in
```
wilfwilson: Probably put some commas in this sentence cos there are a lot of parts to it.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+          <C>Y</C> and <C>g</C> in <C>G</C> such that <C>A</C> is equal to
+          <C>B</C> acted on by <C>g</C>.
+        </Item>
+        <Mark> M4 </Mark> 
+        <Item>
+          For all <C>g</C> in <C>G</C> we have that the intersection of <C>Y</C>
```
wilfwilson: comma after `<C>G</C>`, and get rid of "we have that".

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+#Y  Copyright (C) 2016                                   Christopher Russell
+##
+##  Licensing information can be found in the README file of this package.
+##
+#############################################################################
+##
+
+<#GAPDoc Label="IsMcAlisterTripleSemigroup">
+  <ManSection>
+    <Prop Name = "IsMcAlisterTripleSemigroup" Arg = "S"/>
```
wilfwilson: I've not looked at your implementation yet, but shouldn't this be a filter rather than a property?

doc/z-chap15.xml
```diff
@@ -66,5 +66,121 @@
     <#Include Label = "IsMonogenicInverseMonoid">
 
   </Section>
- 
+
+  <!--**********************************************************************-->
+  <!--**********************************************************************-->
+
+  <Section Label = "McAlister triple semigroups">
```
wilfwilson: @james-d-mitchell Should this be its own chapter? Is this the right place for this doc? If so, should the title of this chapter be changed from `15 Properties and attributes specific to inverse semigroups`?

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+[ (1, ()), (2, ()), (3, ()), (1, (2,4)(3,5)) ]
+gap> AsSemigroup(IsPartialPermSemigroup, T);
+<inverse partial perm monoid of size 4, rank 4 with 4 generators>]]></Example>
+    </Description>
+  </ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="McAlisterTripleSemigroup">
+  <ManSection>
+    <Oper Name = "McAlisterTripleSemigroup" Arg = "G, act, X, Y"
```
wilfwilson: I think the order of arguments in the 4-argument versions should be the same as in the 3-argument versions.

So you have `G, X, Y, act` or `G, X, Y`, and similarly `G, X, sub_ver, act` and `G, X, sub_ver`.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+##
+#############################################################################
+##
+
+<#GAPDoc Label="IsMcAlisterTripleSemigroup">
+  <ManSection>
+    <Prop Name = "IsMcAlisterTripleSemigroup" Arg = "S"/>
+    <Returns><K>true</K> or <K>false</K>.</Returns>
+    <Description>
+      A <E>McAlister triple semigroup</E> is an E-unitary inverse semigroup
```
wilfwilson: You should talk with James already about how to organise this documentation. My intuition would be to have this part of the doc explain literally what filter `IsMcAlisterTripleSemigroup` describes (technically), and for all this mathematical info to be in the documentation for the creation functions.

wilfwilson: @james-d-mitchell will have better advice about this.

wilfwilson: You also need to add doc for everything you defined, including `IsMcAlisterTripleSemigroupElementRep` etc.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+      <Ref Oper= "Digraph" BookName= "Digraphs" /> <A>Y</A> satisfying
+      <Ref Oper= "IsJoinSemilatticeDigraph" BookName= "Digraphs" />, a group
+      <A>G</A>, and a function <A>act</A> defining an action of <A>G</A> on
+      <A>X</A>. <C>McAlisterTripleSemigroup</C> also has a three argument
+      version which assumes that  <A>act</A> is <Ref Func= "OnPoints"
+      BookName= "GAPDoc"/>. Furthermore there are three and four argument
+      versions where the homogeneous list <A>sub_ver</A> of vertices of
+      <A>X</A> replaces <A>Y</A> as an argument. When <A>sub_ver</A> is
+      provided, a McAlister triple semigroup is created with semilattice
+      <C>InducedSubdigraph(<A>X</A>, <A>sub_ver</A>)</C>. <P/>
```
wilfwilson: If I read this documentation, I have absolutely no idea what properties the arguments must satisfy.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+<#GAPDoc Label="McAlisterTripleSemigroupElement">
+  <ManSection>
+    <Oper Name = "McAlisterTripleSemigroupElement" Arg = "S, A, g"/>
+    <Oper Name = "MTE" Arg = "S, A, g"/>
+    <Returns>A McAlister triple semigroup element.</Returns>
+    <Description>
+      Returns the <E>McAlister triple semigroup</E> element of the McAlister
+      triple semigroup <A>S</A> which corresponds to the semilattice vertex
+      <A>A</A> and group element <A>g</A>, if this specifices an element of
+      <A>S</A>. The functions <Ref Prop= "MTE"/> and <Ref Prop=
```
wilfwilson: You've called them `Prop`s here - are they?

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+  <ManSection>
+    <Oper Name = "McAlisterTripleSemigroupElement" Arg = "S, A, g"/>
+    <Oper Name = "MTE" Arg = "S, A, g"/>
+    <Returns>A McAlister triple semigroup element.</Returns>
+    <Description>
+      Returns the <E>McAlister triple semigroup</E> element of the McAlister
+      triple semigroup <A>S</A> which corresponds to the semilattice vertex
+      <A>A</A> and group element <A>g</A>, if this specifices an element of
+      <A>S</A>. The functions <Ref Prop= "MTE"/> and <Ref Prop= 
+        "McAlisterTripleSemigroupElement"/> are identicial.
```
wilfwilson: Can you declare them as synonyms? You should if possible.

doc/semieunit.xml
ChristopherRussell: Are you referring to line 19 or line 16?

doc/semieunit.xml
ChristopherRussell: I think theses should be operations. MTE is just:

ChristopherRussell: These should be referred to as operations. MTE is just:

function(S, A, g)
  return McAlisterTripleSemigroupElement(S, A, g);
end);

Is that ok?

doc/semieunit.xml
ChristopherRussell: I currently have the mathematical details in z-chap15.xml at the start of my section and some of it mentioned again later. Should I rearrange like this:

-Introduction without much detail at start of the section
-Technical details in IsMcAlisterTripleSemigroup
-Mathematical details in McAlisterTripleSemigroup

doc/semieunit.xml
wilfwilson: 19. The way it is written now could be interpreted as "there exists a mcalister triple semigroup S such that every e-unitary inverse semigroup is isomorphic to S".

doc/semieunit.xml
wilfwilson: In the `gd` file one of them should be declared as an operation, yes (probably the one with the long name). The other one, `MTE`, should be declared using `DeclareSynonym` (see [this chapter of the GAP manual](http://www.gap-system.org/Manuals/doc/ref/chap79.html)) for how to do it - pretty sure that is the right way of doing it.

Since this is the piece of the documentation that is about these things, you don't need to have these bits as `Ref`s here. This sentence can be replaced by: `<C>MTE</C> is a synonym for <C>McAlisterTripleSemigroupElement</C>`, or something similar.

wilfwilson: You've also spelt identical wrongly.

doc/semieunit.xml
wilfwilson: Ah okay I hadn't realised. However it should work, it should be clear, when I type `?IsMcAlisterTripleSemigroup` what the documentation means - if it relies on notions defined elsewhere, then it should say so and link to that place.

wilfwilson:
`OneImmutable` for a McAlister triple semigroup has no code coverage.

doc/semieunit.xml
wilfwilson: Sorry this too harsh! What I meant is that, from this bit of text, it's not clear how `G`, `X`, `Y` etc have to be related. If all you did was type `?McAlisterTripleSemigroup` into the command line, it wouldn't be obvious where to get these details.

ChristopherRussell:
Thank you @wilfwilson for all the good suggestions. I have acted on them all but I am still deciding on the organisation of the mathematical and technical details amongst the doc at start of the section as well as the doc for `IsMcAlisterTripleSemigroup` and `McAlisterTripleSemigroup`. Will post again when this is done.

wilfwilson:
Thanks @ChristopherRussell, I'll take a look at it again sometime soon. James will be able to give you some guidance about how best to organise the documentation - it's certainly a good idea to give it some thought.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+        and not (x in DigraphVertexLabels(Y)) then
+      ErrorNoReturn("Semigroups: McAlisterTripleSemigroup: \n",
+                    "Condition M3 is not satisfied - see the documentation ",
+                    "for details,");
+    fi;
+  od;
+
+  # Check condition M4 (essentially, check that X is connected.)
+  min := DigraphSinks(DigraphRemoveLoops(X))[1];
+  if ForAny(GeneratorsOfGroup(Image(hom)), g -> min ^ g <> min) then
```
wilfwilson: This `if` statement checks that the first sink of `X` is fixed by the group `G`. Is that what you're trying to do? Why?

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+  Y := McAlisterTripleSemigroupSemilattice(S);
+  return MTE(S, DigraphSources(DigraphRemoveLoops(Y))[1], ());
+end);
+
+# (A, g) in S if and only if Ag^-1 is a vertex of the semilattice of S
+InstallMethod(AsList, "for a McAlister triple semigroup",
+[IsMcAlisterTripleSemigroup],
+function(S)
+  local out, g, A, V;
+    out := [];
```
wilfwilson: All this code is indented too much.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+          IsSubset(DigraphEdges(X), y_edges)) then
+    ErrorNoReturn("Semigroups: McAlisterTripleSemigroup: usage,\n",
+                  "the fourth argument must be a subdigraph of the third ",
+                  "argument,");
+  elif not IsJoinSemilatticeDigraph(Y) then
+    ErrorNoReturn("Semigroups: McAlisterTripleSemigroup: usage,\n",
+                  "the fourth argument must be a join semilattice digraph,");
+  fi;
+
+  # Check condition M2 (check that Y is an order ideal of X.)
```
wilfwilson: M2 says that if x in X is less than some y in Y, then x is in Y. Could you explain to me what the next block of code is doing? I can't work it out.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+  # TODO: implement IsOrderIdeal for a subset of a partial order digraph.
+  out_nbrs := OutNeighbors(X);
+  for x in DigraphVertexLabels(X) do
+    if not x in DigraphVertexLabels(Y) then
+      for y in DigraphVertexLabels(Y) do
+        if Intersection(out_nbrs[x], out_nbrs[y]) = out_nbrs[x] then
```
wilfwilson: This bit is especially confusing to me.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+function(S, T)
+  local iso_g, iso_x;
+
+  if not IsIsomorphicDigraph(McAlisterTripleSemigroupSemilattice(S),
+                             McAlisterTripleSemigroupSemilattice(T)) then
+    return fail;
+  fi;
+
+  iso_x := IsomorphismDigraphs(McAlisterTripleSemigroupPartialOrder(S),
+                               McAlisterTripleSemigroupPartialOrder(T));
```
wilfwilson: Just wondering: are you assuming that `iso_x` (an iso between the partial orders) restricts to an isomorphism between the respective semilattice sub digraphs?

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+  fi;
+
+  return fail;
+end);
+
+InstallMethod(IsIsomorphicSemigroup, "for two McAlister triple semigroups",
+[IsMcAlisterTripleSemigroup and HasGeneratorsOfSemigroup,
+IsMcAlisterTripleSemigroup and HasGeneratorsOfSemigroup],
+function(S, T)
+  if IsomorphismSemigroups(S, T) = fail then
```
wilfwilson: this function could be shortened to `return IsomorphismSemigroups(S, T) <> fail;`

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+      return false;
+    fi;
+  od;
+  return true;
+end);
+
+InstallMethod(IsFInverseMonoid, "for a McAlister triple semigroup",
+[IsMcAlisterTripleSemigroup],
+function(S)
+  return IsFInverseSemigroup(S) and IsMonoid(S);
```
wilfwilson: Swap these conditions around. It's easier to check `IsMonoid` (in fact it's immediate) so it should come first.

wilfwilson:
In the `master` branch, the required version of Digraphs has been increased to `0.10.0`. This change will conflict with your change, which increased the version to `0.8.1`. To resolve this conflict, go with `0.10.0`.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+      Finally, the user may use <C>McAlsiterTriple</C> with a single argument,
+      <A>S</A>, which is an E-unitary inverse semigroup and it will return a
+      McAlister triple semigroup that is isomorphic to <A>S</A>.
+  </Description>
+  </ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="IsMcAlisterTripleSemigroupElement">
+  <ManSection>
+    <Prop Name = "IsMcAlisterTripleSemigroupElement" Arg = "x"/>
```
ChristopherRussell: I think this came from copy pasting. Will fix.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+    <Prop Name = "IsMTE" Arg = "x"/>
```
ChristopherRussell: Okay, will make it a synonym as with MTE.

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+    </Description>
+  </ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="AsMcAlisterTripleSemigroup">
+  <ManSection>
+    <Oper Name = "AsMcAlisterTripleSemigroup" Arg = "S"/>
+    <Returns>A McAlister triple semigroup.</Returns>
+    <Description>
+      This function can be used to find a McAlister triple semigroup which is
```
ChristopherRussell: I have just realised. `IsomorphismMcAlisterTripleSemigroup` should by implemented as `IsomorphismSemigroup` and I think `AsSemigroup` with filter `IsMcAlisterTripleSemigroup` will then work based on that. Does this sound right?

doc/semieunit.xml
```diff
@@ -0,0 +1,292 @@
...
+    <Returns><K>true</K> or <K>false</K>.</Returns>
+    <Description>
+      This function determines whether a given semigroup is an F-inverse
+      semigroup. F-inverse semigroup are a class of E-unitary inverse
+      semigroups. A McAlister triple semigroup
+      <C>McAlisterTripleSemigroup(G,act,X,Y)</C> is F-inverse precisely when
+      the partial order digraph <C>X</C> is a join-semilattice digraph. This
+      function uses <Ref Oper = "AsMcAlisterTripleSemigroup"/> to find a
+      McAlister triple semigroup isomorphic to <A>S</A> then checks if the
+      McAlister triple semigroup is F-inverse.
```
ChristopherRussell: I hadn't thought of that characterisation and it seems much less complicated that the typical definition of F-Inverse semigroup. I will investigate and get back to you.

doc/semieunit.xml
wilfwilson: Oh yes, these should be intregrated into `IsomorphismSemigroup` and `AsSemigroup`, as you say.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+#
+# This code is based on the description of McAlister triple semigroups found
+# in Fundamentals of Semigroup Theory by John Howie.
+#
+#############################################################################
+# Methods for creating McAlister triple semigroups
+#############################################################################
+InstallMethod(McAlisterTripleSemigroup,
+"for a perm group, action, digraph, and digraph",
+[IsGroup, IsFunction, IsDigraph, IsDigraph],
```
ChristopherRussell: Sorry, I think I forgot to change the label at some point. It should work for any finite group, when there is also a valid action given as an argument.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+                  "automorphisms on the third argument (a partial order ",
+                  "digraph),");
+  elif not IsPartialOrderDigraph(X) then
+    ErrorNoReturn("Semigroups: McAlisterTripleSemigroup: usage,\n",
+                  "the third argument must be a partial order digraph,");
+  fi;
+
+  # To check if the edges of y are a subset of the edges of x we need to apply
+  # the labeling of y's vertices to its edges.
+  y_edges := List(DigraphEdges(Y), edge -> [DigraphVertexLabels(Y)[edge[1]],
```
ChristopherRussell: Y is meant to be a subdigraph of X. The vertices of Y should be labelled by the vertex of X they are sent to by some embedding into X. y_edges are the edges of Y but (Label(a),Label(b)) instead of (a,b). So these are supposed to be the edges of Y sent by some embedding into X.

ChristopherRussell: I agree this should be documented. I wrote it a long time ago and was told that the labelling system maybe changing in Digraphs so that I didn't have to hack edge labels like this.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,715 @@
...
+                                            DigraphVertexLabels(Y)[edge[2]]]);
+
+  if not (IsSubset(DigraphVertexLabels(X), DigraphVertexLabels(Y)) and
+          IsSubset(DigraphEdges(X), y_edges)) then
+    ErrorNoReturn("Semigroups: McAlisterTripleSemigroup: usage,\n",
```
ChristopherRussell: My definition would be that there is an embedding f:Vertices(Y) -> Vertices(X) such that for all e in Edges(Y) then f(e) is in Edges(X)

gap/semigroups/semieunit.gi
wilfwilson: But later on in the code, there is a line that checks whether the group is a subgroup of the automorphism group of the digraph -- and the automorphism group of a digraph (as implemented at the moment) is a perm group. So the code, as written, only works for perm groups at the moment.

gap/semigroups/semieunit.gi
wilfwilson: Oh right. So I get that the label of a vertex of `Y` (i.e. `Label(a)`) should tell you which vertex of `X` that it corresponds to -- but, does `Label(a)` give you the corresponding *vertex number* of X, or the *vertex label*? Do you get what I'm asking?

gap/semigroups/semieunit.gi
ChristopherRussell: I think that line checks whether the `Image(hom)` is a subgroup (Image(hom) is the representation of the action of the group as a perm group) so this should be fine!

wilfwilson: Ah brilliant, I missed that.

gap/semigroups/semieunit.gi
ChristopherRussell: The label of the vertex of `Y` should tell you the label of the vertex of `X` which it corresponds to.

gap/semigroups/semieunit.gi
wilfwilson: Suppose X is the digraph `1 -> 2 -> 3`, which has labels `[2, 1, 3]`. By that, I mean that `label(1) = 2`, `label(2) = 1`, and `label(3) = 3`.

Suppose Y is the digraph  `1 -> 2`, with vertex labels `[1, 3]`. So `Y = InducedSubdigraph(X, [2, 3])`.

Therefore `Y` is a perfectly valid digraph for the purposes of this code.

However, `DigraphEdges(X)` is `[[1,2], [2,3]]`, and `y_edges` is `[[1,3]]`, so this condition in the `if` statement is failed.

There is an inconsistency going on in your code, depending on whether you're referring to vertices of `X` by their vertex numbers, or by their vertex labels.

gap/semigroups/semieunit.gi
ChristopherRussell: For each vertex of `X` which does not correspond to a vertex of `Y` we check whether there are any vertices which are 'less' than it which correspond to a vertex of `Y`. Vertex `a` is less than vertex `b` when there is an edge from `b` to `a`. 
The intersection of the out neighbours of two vertices is equal to their join. The join of a vertex in `X` with a vertex in `Y` is either less than or equal to a vertex of `Y` or is equal to that vertex in `X`. I should be using `PartialOrderDigraphJoinOfVertices` here!

gap/semigroups/semieunit.gi
wilfwilson: (This links into my message that I posted a minute ago). In that case, in the code below, instead of doing `IsSubset(DigraphEdges(X), y_edges))`, you should first construct `x_edges` from `DigraphEdges(X)` by sending each vertex number to its vertex label.

gap/semigroups/semieunit.gi
wilfwilson: So you're doing this by contrapositive basically: for each vertex of `X` that is not in `Y`, if the vertex is less than something in `Y`, then we fail.

Since `X` is a partial order, it is transitive. So if we want to know whether `a` is less than `b` by checking whether `b` is in the out-neighbours of `a`.

wilfwilson: There's also another vertex number/label of `X` conundrum here: `x` is one of the vertex labels of `X`, *not* necessarily the actualy vertex number. However, `out_nbrs[x]` gives the out-neighbours (as vertex numbers, not labels) of the vertex whose number (not label) is `x`. Do you see the problem?

i.e `x` might be vertex `1`, however it could have label `2`. And `out_nbrs[x]` is `out_nbrs[2]`, which might have no relation to `out_nbrs[1]`, which describes the actual out-neighbours of `x`.

ChristopherRussell: You are correct regarding just checking out-neighbours vs intersections.

I see that using the labels of `X` is a bit of an issue and I don't really recall why I chose to do so now. Using the labels of `Y` was necessary to know how `Y` embeds into `X` but perhaps I used them mistakingly here or thought the user might like to label their digraph! I think it should be changed to loop over `DigraphVertices(X)` instead of the labels.

gap/semigroups/semieunit.gi
ChristopherRussell: I see. I think I should change things so that `X` is always referred to by it's vertex numbers.

gap/semigroups/semieunit.gi
wilfwilson: I agree. Using vertex labels of `Y` makes sense, but using the vertex labels of `X` seems to complicate things. So the label of a vertex of `Y` should refer to a vertex *number* of a vertex of `X`. There might be a few parts of your code where you'll need to check that the right thing is being done, and of course the documentation needs to be precise.

gap/semigroups/semieunit.gi
ChristopherRussell: This should be the first sink of `Y` not of `X`. The only sink of `Y` should be the minimal element of the semilattice. 

M4: for all g in `G`,   The intersection of `gY` and `Y` is non-empty

Automorphisms of finite semilattices should fix the minimal element so checking this checks M4.

(Perhaps the comment about checking `X` is connected isn't very helpful. M3 checks that `X = GY` (i.e. all elements of X are images of elements of Y by G). So checking the minimal element of `Y` is fixed is equivalent to showing `X` is connected.)

wilfwilson: Given that it's meant to be `X` not `Y`, I'm happy with this now.

gap/semigroups/semieunit.gi
ChristopherRussell: What should be different?

gap/semigroups/semieunit.gi
ChristopherRussell: You are right, I need to check that. Should I do this by using `RestrictedMapping`... then is there something existing which can check if this is an isomorphism?

gap/semigroups/semieunit.gi
ChristopherRussell: Good point

gap/semigroups/semieunit.gi
wilfwilson: The line containing `out := [];` should be 2 characters to the left, lined up with `local`. And same for the lines below it.

gap/semigroups/semieunit.gi
ChristopherRussell: Actually `Y` should be the induced subdigraph not just any subdigraph. Sorry for the confusion. So I will change this to check to:

if Y = InducedSubdigraph(X, DigraphVertexLabels(Y)) then

gap/semigroups/semieunit.gi
wilfwilson: Not sure how is best to do this. Certainly `iso_x` is an isomorphism from the `X` for `S` to the `X` for `T`, and `iso_x` embeds the `Y` for `S` into some copy in the `X` for `T`. However, perhaps it gets mapped to a different one than the *actual* `Y` for `T`. If there exists *some* `iso_x` however, then there must exist one that does what you want it to. The question is how do you get your hands on it? Not sure right now.

ChristopherRussell: I will comment this function out until we can think of something.

ChristopherRussell:
I made some commits but I'm not finished yet. Still need to:

-Fix IsomorphismSemigroups for two McAlister triple semigroups, as discussed above not sure how to do this.
-Document `McAlisterTripleSemigroupElementReps`
-Decide on location of the functions `EUnitaryCover`, `IsFInverseSemigroup`, `IsFInverseMonoid` and their documentation.
-Implement Wilf's suggested improvement for EUnitaryCover involving `PartialOrderOfDClasses`

ChristopherRussell:
I tried searching for other examples of Reps. There is `IsEnumerableSemigroupRep` but other examples I found by searching .gd files (ag DeclareRepresentation -G .gd) were undocumented:

```IsEnumerableSemigroupGreensClassRep
IsEnumerableSemigroupGreensRelationRep
IsPlistRowBasisOverFiniteFieldRep
IsPlistMatrixOverFiniteFieldRep
IsCongruenceByGeneratingPairsRep
IsCongruenceClassByGeneratingPairsRep```

The only thing I can think to say in the doc for `McAlisterTripleSemigroupElementReps` is that these are objects created by McAlisterTripleSemigroupElement.

File: doc/semieunit.xml (hunk 26 above)
Comment (ChristopherRussell) [inline doc/semieunit.xml]: Actually this doesn't work. 

`McAlisterTripleSemigroup(Group((4,5)), Digraph([[1],[1,2],[1,2,3],[1,2,3,4],[1,2,3,5]]), [1 .. 4]);`

is not an F-inverse semigroup since the 2nd argument is not a join-semilattice yet the partial order of D-classes

`[ [ 1, 2, 3, 4 ], [ 2 ], [ 2, 3 ], [ 2, 4 ] ]`

defines a join-semilattice digraph.

Comment (wilfwilson) [inline doc/semieunit.xml]:
```
gap> Digraph([[1],[1,2],[1,2,3],[1,2,3,4],[1,2,3,5]]);
<digraph with 5 vertices, 14 edges>
gap> IsJoinSemilatticeDigraph(last);
true
```
It says it's a join-semilattice for me

Comment (ChristopherRussell) [inline doc/semieunit.xml]: Sorry, the Digraph was meant to be `Digraph([[1],[1,2],[1,3],[1,2,3,4],[1,2,3,5]])` with no edge from 3 to 2.

Comment (wilfwilson) [inline doc/semieunit.xml]: Okay good work, you'll have to go back to your old method then.

**Comment by ChristopherRussell** [comment]
I have now pushed a fix to `IsomorphismSemigroups` for two `McAlisterTripleSemigroup` semigroups, added comments in the .gd file describing `McAlisterTripleSemigroupElementRep` and `McAlisterTripleSemigroupDefaultRep` (as opposed to writing documentation for them) and created a new chapter (now chapter 12) for McAlister triple semigroup documentation.

File: gap/semigroups/semieunit.gi [hunk 30]
```diff
@@ -0,0 +1,721 @@
...
+      g -> ForAll(DigraphVertices(McAlisterTripleSemigroupPartialOrder(S)),
+      x -> (McAlisterTripleSemigroupAction(S)(x, g)) ^ (rep * iso_x) =
+      McAlisterTripleSemigroupAction(T)((x ^ iso_x), (g ^ iso_g)) ^ rep)) then
+    return MappingByFunction(S, T, s -> MTSE(T, s[1] ^ iso_x, s[2] ^ iso_g));
+  fi;
+
+  return fail;
+end);
+
+InstallMethod(IsIsomorphicSemigroup, "for two McAlister triple semigroups",
```
wilfwilson: You can get rid of this method: the one in `gap/attributes/isomorph.gi` works perfectly fine for MTSs.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,721 @@
...
+
+InstallMethod(ViewString, "for a McAlister triple semigroup",
+[IsMcAlisterTripleSemigroup],
+function(S)
+  local G;
+  G := McAlisterTripleSemigroupGroup(S);
+  return Concatenation("<McAlister triple semigroup over ", ViewString(G), ">");
+end);
+
+# Currently this does not work. iso_x restricted to the semilattice is not
```
wilfwilson: Is this comment out of date now?

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,721 @@
...
+  fi;
+
+  im_YS := List(DigraphVertexLabels(YS), a -> a ^ iso_x);
+  # if the restriction of iso_x to DigraphVertexLabels(YS) is not
+  # DigraphVertexLabels(YT) then we need to compose iso_x with an
+  # automorphism of McAlisterTripleSemilattice(T). Composing this with
+  # iso_x will restrict to an isomorphism from (the labels of) YS to YT.
+  if not im_YS = DigraphVertexLabels(YT) then
+    A := AutomorphismGroup(XT);
+    rep := RepresentativeAction(A, im_YS, DigraphVertexLabels(YT), OnSets);
```
wilfwilson: It is possible that `RepresentativeAction` returns `fail` here: in this case, the method should return `fail`.
i.e
```
if rep = fail then
  return fail;
fi;
```
I'll post an example below.

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,721 @@
...
+#############################################################################
+# Methods for McAlister triple elements
+#############################################################################
+InstallMethod(McAlisterTripleSemigroupElement,
+"for a McAlister triple semigroup, pos int, and perm",
+[IsMcAlisterTripleSemigroup, IsPosInt, IsMultiplicativeElementWithInverse],
+function(S, A, g)
+  if not A in DigraphVertexLabels(McAlisterTripleSemigroupSemilattice(S)) then
+    ErrorNoReturn("Semigroups: McAlisterTripleSemigroupElement: usage,\n",
+                  "second input should be a vertex label of the ",
```
wilfwilson: input -> argument

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,721 @@
...
+                  "join-semilattice of the McAlister triple,");
+  elif not g in McAlisterTripleSemigroupGroup(S) then
+    ErrorNoReturn("Semigroups: McAlisterTripleSemigroupElement: usage,\n",
+                  "third input must an element of the group of the McAlister ",
```
wilfwilson: input -> argument

doc/semieunit.xml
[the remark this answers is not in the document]
ChristopherRussell: Oops. So `Prop` should be `Filt`, right?

gap/semigroups/semieunit.gi
ChristopherRussell: Thanks for spotting this and the example.

doc/semieunit.xml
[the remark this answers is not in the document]
wilfwilson: Yes

ChristopherRussell:
I have made the most recent suggested changes now @wilfwilson. Apologies for the delay, I was away and without my laptop.

wilfwilson:
@james-d-mitchell Can you tell me what you think the correct behaviour should be here?

I'll create two `McAlisterTripleSemigroup` objects:
```
gap> S := McAlisterTripleSemigroup(Group([(1, 2, 3)(4, 5, 6), (8, 9)]),
>  DigraphFromGraph6String("+H_A?GC_Q@G~wA?G"), [1, 4, 5, 7, 8]);
gap> T := McAlisterTripleSemigroup(Group([(1, 2, 3)(4, 5, 6), (8, 9)]),
>  DigraphFromGraph6String("+H_A?GC_Q@G~wA?G"), [1, 4, 5, 7, 8]);
```
`S` and `T` are mathematically identical, because they're created in exactly the same way. But they're different objects, and their elements belong to different families, since each McAlisterTripleSemigroup has its own family. However, the equality method for these two objects returns `true:`
```
gap> S = T;
true
```
If I create two mathematically identical objects in each of these semigroups:
```
s := MTSE(S, 1, ());
t := MTSE(T, 1, ());
```
then, again, the equality method for `McAlisterTripleSemigroupElements` will say that they are equal, since their semigroups and parameters are equal:
```
gap> s = t;
true
```
But then the `in` method is not consistent with this:
```
gap> s in S and t in T;
true
gap> s in T or t in S;
false
```

This has knock-on implications, such as when it comes to re-creating something from its `Print` or `String` value. If `String` gives the following:
```
"Semigroup([
  MTSE(McAlisterTripleSemigroup(SymmetricGroup([2 .. 5]),
        Digraph([[1], [1, 2], [1, 3], [1, 4], [1, 5]]), [1 .. 4]), 3, (3, 4)),
  MTSE(McAlisterTripleSemigroup(SymmetricGroup([2 .. 5]),
        Digraph([[1], [1, 2], [1, 3], [1, 4], [1, 5]]), [1 .. 4]), 1, (2, 4))])"
```
then `EvalString(last)` fails for obvious reasons: the two generators are created relative to two semigroups that, although mathematically the same, have different families, and you can't create a semigroup out of objects in different families.

How should things be? Certainly I think that the behaviour should be consistent, and would preferably allow a `McAlisterTripleSemigroup` or subsemigroup to be re-created from its `String`, but I'm not sure if that's possible.

wilfwilson:
@ChristopherRussell I made some minor changes following my most recent review, and I've put them in a pull request to your e-unitary branch in your fork of the repository. So you just need to merge that once to get those changes. They'll then appear as part of this pull request.

Once you merge in my PR, and once @james-d-mitchell and I have resolved the way that equality should work with MTS's and MTSE's, I'm happy for this to be merged in. I'm hoping to do it at the Wednesday meeting tomorrow.

james-d-mitchell:
Semigroups consist of sets, and two sets are equal if and only if they have there same elements. Therefore the correct solution is for elements of distinct McAlister triples should never be equal. I think this is the case with Rees matrix semigroups, and it would be consistent if McAlister triples behaved the same. I don't see how to fix the string method. I'll write some more about this tomorrow

wilfwilson:
I agree.

ChristopherRussell:
I made the fix to the equality methods and changed tests accordingly.

wilfwilson:
@ChristopherRussell Can you push your changes please?

ChristopherRussell:
I did push them. I put them in with my last commit (95e1d9f). It's the one before your fix to String. Sorry if that was confusing.

wilfwilson:
Thanks @ChristopherRussell

james-d-mitchell:
Hmm, I have some comments on this too. @ChristopherRussell can you make a new pull request to fix the one or two minor issues I am about to point out?

gap/semigroups/semieunit.gi
```diff
@@ -0,0 +1,719 @@
...
+[IsPermGroup, IsDigraph, IsHomogeneousList],
+function(G, X, sub_ver)
+  return McAlisterTripleSemigroup(G, X, InducedSubdigraph(X, sub_ver),
+                                  OnPoints);
+end);
+
+#############################################################################
+# Methods for McAlister triple semigroups
+#############################################################################
+# InstallMethod(\=, "for two McAlister triple semigroups",
```
james-d-mitchell: This commented out method should either be deleted, or you could include this in the isomorphism semigroup code if it would be worthwhile doing this.

gap/semigroups/semieunit.gd
```diff
@@ -0,0 +1,100 @@
...
+#############################################################################
+##
+
+DeclareCategory("IsMcAlisterTripleSemigroupElement",
+                IsAssociativeElement and IsMultiplicativeElementWithInverse);
+
+# This is a representation for McAlister triple semigroup elements, which are
+# created via the function McAlisterTripleSemigroupElement.
+#
+# The components are:
```
james-d-mitchell: This says that there are 3 components, it should really say that there are 3 positions, and what is stored in each position, and then in `DeclareRepresentation` you only declare 2 positions, so which is it, 2 or 3?

ChristopherRussell: The components 'Vertex' and 'Group element' are accessible by [] but parent is not. Instead parent is accessed via `McAlisterTripleSemigroupElementParent` (at some point we decided it was not user friendly for it to be accessible by [] because these elements are printed as pairs, not triples).

Should there be 2 positions declared or 3?
(Is this number the positions accesible by [] or the ones accessible by ![]?)

james-d-mitchell: It should be a description of the internal representation, so the ones accessible by ![], and so it should be 2 positions, according to your description.

ChristopherRussell: There are 3 positions accessible by ![]. Did you mean to say it should be 3?

james-d-mitchell:
Oh sorry I misunderstood what you wrote, yes, 3.

PR #273, opened 2017-05-15
Add CongruenceTestMembershipNC

mtorpey:
**tldr: this is a "no checks" version of `\in` for congruences.**

We want to be able to use
```gap
[x, y] in cong;
```
to check whether a pair is in a congruence.  Previously, we have had a different `\in` method for each type of congruence, which checks input and then does the appropriate thing.  However, we want to be able to call these methods without doing checks first, and even without constructing lists (which is also an overhead).

This PR replaces all the `\in` methods with methods for `CongruenceTestMembershipNC`, a new operation which takes three arguments `cong, elm1, elm2` and returns true iff `elm1` and `elm2` are congruent with respect to `cong`.  This faster method can now be used when we already know the input is valid (e.g. in `LatticeOfCongruences`, which at the moment is incorrectly using `CONG_PAIRS_IN`).  There is a single `\in` method in `cong.gi` (well, one for left congs and one for right congs) which does all the checks and then calls `CongruenceTestMembershipNC(cong, pair[1], pair[2])` and returns the answer.

While I was at it, I modified `CONG_PAIRS_IN` in `congpairs.cc` to work in the same way.

### DIFF
--- a/gap/congruences/cong.gd
+++ b/gap/congruences/cong.gd
@@ -118,3 +118,8 @@ DeclareOperation("LeftCongruenceClassOfElement", [IsLeftSemigroupCongruence,
                                                   IsMultiplicativeElement]);
 DeclareOperation("RightCongruenceClassOfElement", [IsRightSemigroupCongruence,
                                                    IsMultiplicativeElement]);
+
+# No-checks version of the "\in" operation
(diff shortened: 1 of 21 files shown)


wilfwilson:
Seems fine to me.

Issue #274, opened 2017-05-24
Heisenbug regarding comparing fp semigroup elements


mtorpey:
This shows the problem:

```gap
repeat
#T# A right congruence example
F := FreeMonoid(2);;
M := F / [[F.1 * F.2 ^ 2, F.2 ^ 2],
              [F.2 ^ 3, F.2 ^ 2],
              [F.1 ^ 4, F.1],
              [F.2 * F.1 ^ 2 * F.2, F.2 ^ 2],
              [F.2 * F.1 ^ 3 * F.2, F.2],
              [(F.2 * F.1) ^ 2 * F.2, F.2],
              [F.2 ^ 2 * F.1 ^ 3, F.2 ^ 2],
              [F.2 * (F.2 * F.1) ^ 2, F.2 ^ 2 * F.1 ^ 2]];;
cong := RightSemigroupCongruence(M, [M.1, M.2 ^ 3]);;
semicong := cong!.semicong;;
semipart := EquivalenceRelationPartition(semicong);;
elm1 := M.1 ^ cong!.iso;;
part1 := First(semipart, l -> elm1 in l);;
set1 := SortedList(part1);
until ForAny(part1, x-> not x in set1);
```

james-d-mitchell:
I can't reproduce this, running for a couple of minutes now with no problems.

james-d-mitchell:
And that's with reporting on and off, that I can't reproduce this.

wilfwilson:
Just to record: I also couldn't reproduce this. However @james-d-mitchell and @mtorpey believe they have identified solved the problem. It was a bug in libsemigroups. Michael has submitted [a PR to libsemigroups](https://github.com/james-d-mitchell/libsemigroups/pull/34) that resolves this.

mtorpey:
This is now resolved, after my libsemigroups PR that Wilf mentioned.

PR #275, opened 2017-05-29
Remove mentions and examples containing projective matrices in doc

ChristopherRussell:
See trello card for the removed examples. The mentions in the doc have just been commented out.


james-d-mitchell:
This appears to be failing because of the Issue #274, which we will resolve shortly.

PR #278, opened 2017-05-29
doc: add stubs for IsEnumerableSemigroupRep and IsActingSemigroup

mtorpey:
There are TODOs in them which someone (probably @james-d-mitchell) should fill in later.


doc/z-chap06.xml
```diff
@@ -7,28 +7,12 @@
   be created in &Semigroups;, and the options that are available at the time of
   creation.
  
-  <Section Label = "Acting semigroups"><!--TODOFIXME: move this-->
```
james-d-mitchell: Is deleting these two big chunks of text a good idea? Unless they are repeated elsewhere they should probably just be commented out for now.

mtorpey:
@james-d-mitchell I changed my commit so that it just *comments out* the old doc instead of deleting it entirely.

PR #279, opened 2017-05-29
doc: multiple labels for Matrix and RandomMatrix functions

mtorpey:
`Matrix` and `RandomMatrix` were multiply defined, so I've added labels.

Also fixed a little LaTeX error to do with using `\text`.


james-d-mitchell:
Great!

PR #291, opened 2017-05-30
Document and slightly change DigraphOfActionOnPoints

wilfwilson:
`DigraphOfActionOnPoints` was undocumented. I've documented it. In the process, I thought it made sense to make it consistent with `DigraphOfActionOnPairs`. Thus the digraph now has edge labels that give the index of the generator that corresponds to an edge.

Due to this, the method takes a small performance hit - we previously added new edges using `AsSet` to avoid creating duplicates, but this sorts the out-neighbours list in a way that makes it impossible to keep track of edge labels. Since the out-neighbours are not sorted, this (harmlessly) changes the output of some tests, with functions that use `DigraphOfActionOnPoints`.

### DIFF
--- a/gap/semigroups/semitrans.gi
+++ b/gap/semigroups/semitrans.gi
@@ -207,32 +207,33 @@ end);
 InstallMethod(DigraphOfActionOnPoints,
 "for a transformation semigroup with known generators",
 [IsTransformationSemigroup and HasGeneratorsOfSemigroup],
-function(S)
-  local n;
-
-  n := DegreeOfTransformationSemigroup(S);
-  if n = 0 then
-    return EmptyDigraph(0);
-  fi;
-  return DigraphOfActionOnPoints(S, n);
-end);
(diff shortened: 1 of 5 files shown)


gap/semigroups/semitrans.gi
```diff
@@ -207,35 +207,43 @@ end);
...
-    for j in [1 .. n] do
-      k := j ^ x;
-      if k > n then
+  for i in [1 .. n] do
+    for j in [1 .. Length(gens)] do
+      range := i ^ gens[j];
+      if range > n then
         return fail;
+      elif not range in out[i] then
+        Add(out[i], range);
```
james-d-mitchell: As per our discussion right now, remove `labels` and use `AddSet` here.

wilfwilson:
@james-d-mitchell I sorted this out.

mtorpey:
Wilf has made the "one minor change" so I'm merging this.

PR #294, opened 2017-05-30
doc: minor matrix semigroups changes

mtorpey:
When we mention that users should use `RandomSemigroup` for random matrix semigroups, it should be immediately obvious which filter to use: `IsMatrixOverFiniteFieldSemigroup`.

And a couple of other tiny things.

### DIFF
--- a/doc/z-chap05.xml
+++ b/doc/z-chap05.xml
@@ -151,9 +151,7 @@
     </Heading>
 
     In this section we describe the operations, properties, and attributes in
-    &Semigroups; specifically for Boolean matrices. These are in addition to
-    those given elsewhere in this chapter for arbitrary matrices over
-    semirings. These include:
+    &Semigroups; specifically for Boolean matrices.  These include:
 
     <List>
       <Item>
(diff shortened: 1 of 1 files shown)


doc/z-chap05.xml
```diff
@@ -344,15 +344,16 @@
     &GAP; provides a way to define matrices which are in the
     filter <Ref Filt="IsMatrix" BookName="ref"/>.
     For technical reasons, the matrix semigroup functions in &Semigroups; rely
-    on a custom wrapper for matrices <Ref Filt="IsMatrixOverSemiring"/>.<P/> 
+    on a custom wrapper for matrices <Ref Filt="IsMatrixOverSemiring"/>.
     We take precautions to hide this fact from the user of &Semigroups; and
-    also provide conversion functions between the two representations.
+    also provide conversion functions between the two representations. <P/>
```
james-d-mitchell: I'm not sure where this paragraph came from, but I don't think it is accurate, and I think the whole paragraph should be deleted. I think this is a left over from the Semigroups 2.8 doc, written by @markuspf, which was true at the time, but is no longer the case.

james-d-mitchell:
This comment is more about the content of this chapter than this particular PR. 

1.
>     Random matrix semigroups can be created by using the functions
>   <Ref Func="RandomSemigroup"/> or <Ref Func="RandomMonoid"/> with the filter
>    <Ref Filt="IsMatrixOverFiniteFieldSemigroup"/>.  The same filter can be used
 >   with the functions <Ref Attr="IsomorphismSemigroup"/> and
 >   <Ref Oper="AsSemigroup"/> to create a matrix semigroup isomorphic to a given
 >   semigroup. <P/>

Random matrix semigroups can be created using any of the filters `IsXMatrixSemigroup` or `IsXMatrixMonoid`, i.e. `Is IsMatrixOverFiniteFieldSemigroup`, `IsNTPMatrixSemigroup` etc. The documentation above is overly restrictive and misleading, it should be changed to include a reference to the appropriate section. 

2.
>  In this section we describe operations and attributes in
>    &Semigroups; specifically for matrix semigroups. These are in addition to
>   those given elsewhere in this chapter for arbitrary matrices over
>    semirings. 

The second sentence of this should be deleted, as it does not make sense in the context. 

3. The matrix groups section does not make it clear that this only applies to matrices of over finite fields, and to no other type of matrix semigroup. This should be made clear.

mtorpey:
Done - all in one commit.

PR #295, opened 2017-05-30
Update bipartition documentation: multiplication and annular

wilfwilson:
I've reviewed the beginning of the chapter on bipartitions. My changes mostly focuses on the multiplication, and on what it means for a partition to be annular.

The previous definition of multiplication was incorrect.

### DIFF
--- a/doc/z-chap03.xml
+++ b/doc/z-chap03.xml
@@ -3,13 +3,15 @@
     Bipartitions and blocks
   </Heading>
 
-  In this chapter we describe the functions in &Semigroups; for creating
-  and manipulating bipartitions and semigroups of bipartitions. We begin by
+  In this chapter we describe the functions in &Semigroups; for creating and
+  manipulating bipartitions and semigroups of bipartitions. We begin by
   describing what these objects are.
   <P/>
 
   A <E>partition</E> of a set &setX; is a set of pairwise disjoint non-empty
-  subsets of &setX; whose union is &setX;.
+  subsets of &setX; whose union is &setX;. A partition of &setX; is the
+  collection of equivalence classes of an equivalence relation on &setX;, and
+  vice versa.
   <P/>
 
   Let <M>n\in</M>&bbN;, let &bfn;<M> = \{1, 2, \ldots, n\}</M>, and let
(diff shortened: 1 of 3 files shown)


wilfwilson:
I recommend that whoever reviews this reviews it by checkout out the branch, compiling the documentation, and then viewing it in the terminal/browser.

doc/z-chap03.xml
```diff
@@ -33,70 +35,107 @@
...
     </Item>
- 
   </List>
 
-  This product can be shown to be associative, and so the collection of
-  bipartitions of any particular degree is a monoid; the identity element is the
-  partition
-  <Alt Not = "Text"><M>\left\{\{i,-i\}:i\in\mathbf{n}\right\}</M>.</Alt>
-  <Alt Only = "Text">{{<C>i</C>,-<C>i</C>}:<C>i</C><M>\in</M><E>n</E>}.</Alt>
-  A bipartition is a unit if and only if each block is of the form
```
mtorpey: I believe each `<C>i</C>` in the following paragraph should be replaced by `<M>i</M>` (and the same for `-i`).

doc/z-chap03.xml
```diff
@@ -33,70 +35,107 @@
...
+  Alternatively, the product &x;&y; can be described as follows. Define
+  &bfn;'<M> = \{1', 2', \ldots, n'\}</M>. From &x;, create a partition &x;' of
+  the set &bfn;<M>\cup</M>&bfn;' by replacing each negative point <C>-i</C> in a
+  block of &x; by the point <C>i'</C>, and create from &y; a partition &y;' of
+  the set &bfn;'<M>\cup</M>-&bfn; by replacing each positive point <C>i</C> in a
+  block of &y; by the point <C>i'</C>. Then define a relation on the set
+  &bfn;<M>\cup</M>&bfn;'<M>\cup</M>-&bfn;, where &i; and &j; are related if they
+  lie in the same part of either &x;' or &y;', and let &p; be the transitive
```
mtorpey: `the same part` should be `the same block` for consistency with the rest of the paragraph.

doc/z-chap03.xml
```diff
@@ -121,17 +160,8 @@
 
   A bipartition &x; of degree &n; is called <E>planar</E> if there does not exist
```
mtorpey: if there **do** not exist

(sorry, I know you haven't actually modified this bit)

doc/z-chap03.xml
```diff
@@ -121,17 +160,8 @@
...
-    <Item>
-      <M>a &lt; b &lt; a' &lt; b'</M>; or
-    </Item>
-    <Item>
-      <M>b &lt; a &lt; b' &lt; a'</M>.
-    </Item>
-  </List>
-  Or equivalently, &x; is planar if for each distinct blocks <M>A, B \in</M> &x;,
+  <M>b, b' \in B</M>, such that <M>a &lt; b &lt; a' &lt; b'</M>.
+  Equivalently, &x; is planar if for each distinct blocks <M>A, B \in</M> &x;,
```
mtorpey: "each distinct blocks" is bad English.

doc/z-chap03.xml
```diff
@@ -148,10 +178,33 @@
       <M>b &lt; b' &lt; a &lt; a'</M>.
     </Item>
   </List>
```
mtorpey: Perhaps we might scrap the whole "equivalently" definition, since it's much more complicated and seems (to me) to add nothing.

Also, would it be good to use `a_1` and `a_2` instead of `a` and `a'`?  It's just that so much of the literature uses `i` for points on the top and `i'` for points on the bottom, it confused me for a bit here.

Both of these comments are suggestions, and since this PR only improves the situation, this won't stop me merging.

doc/z-chap03.xml
wilfwilson: I presume James had a reason for doing it the way he did - but there don't seem to be any problems when I made the change so I'll do it.

doc/z-chap03.xml
wilfwilson: We defined *block* to be the parts of a bipartition - the unholy equivalences `x` and `y` here are not bipartitions (they're defined on the wrong set). Do you think I should still use block? In fact I'll just say "...if they are related in either x or y...'

doc/z-chap03.xml
wilfwilson: I've deleted this bit.

wilfwilson:
@mtorpey Think I've sorted this out.

PR #308, opened 2017-06-01
Fix failing test in extreme/cong.tst

mtorpey:
This was caused by a naive assumption in `EquivalenceClasses` - namely that the fp semigroup in question is finite; in fact it might very well be infinite, but might have a finite number of equivalence classes.

This method could be improved further if we introduced a method in libsemigroups which returned a representative for a congruence class - just as we were discussing earlier!  James suggested that he and I might do that next week, and then an even better version of this method could perhaps go into Semigroups 3.1

### DIFF
--- a/gap/congruences/congpairs.gi
+++ b/gap/congruences/congpairs.gi
@@ -404,22 +404,26 @@ InstallMethod(EquivalenceClasses,
 "for an fp semigroup congruence",
 [IsFpSemigroupCongruence],
 function(cong)
-  local part, enum, reps, classes, next, i;
-
-  part := CongruenceByGeneratingPairsPartition(cong);
+  local enum, nrclasses, classes, class_added, count, i, word, id;
   enum := Enumerator(Range(cong));
-  reps := List(part, x -> enum[x[1]]);
+  nrclasses := NrEquivalenceClasses(cong);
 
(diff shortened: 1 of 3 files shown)


gap/congruences/congpairs.gi
```diff
@@ -404,22 +404,24 @@ InstallMethod(EquivalenceClasses,
...
+  repeat
+    i := i + 1;
+    word := SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(enum[i]));
+    id := CONG_PAIRS_ELM_COSET_ID(cong, word);
+    if not class_added[id] then
+      classes[id] := EquivalenceClassOfElementNC(cong, enum[i]);
+      SetCongruenceClassByGeneratingPairsCosetId(classes[id], id);
+      class_added[id] := true;
+    fi;
+  until not false in class_added;
```
wilfwilson: Presumably this requires a linear search every time the loop happens. How about: have a counter, it starts at `0`, you add `1` to it each time you hit the `if` block of code, and then the `until` line would be `until count = nrclasses;`, or whatever.

mtorpey: Good idea!

mtorpey:
@wilfwilson, I've done as you suggested.

PR #316, opened 2017-06-02
GeneratorsSmallest Doc: fix and move to attr

ffloresbrito:
In the documentation of GeneratorsSmallest: reference to transformation semigroups is removed, and a new example is added. Also, the documentation is moved from semitrans.xml to attr.xml, and relocated from "Attributes of transformation semigroups" to "Generating sets" (in z-chap13.xml).
Lastly, some linting is done in the file attr.xml that was not previously identified by gaplint.


james-d-mitchell:
This is failing because the manual examples are not correctly formatted. Run `SEMIGROUPS.TestManualExample` to fix it.

james-d-mitchell:
You failed to remove several references to transformations, and the tests don't pass.

PR #325, opened 2017-06-04
Install specific IsIdempotentGenerated methods for RMS/RZMS

wilfwilson:
For a regular Rees (0-)matrix semigroup over a group, it's easy to say write down a fairly small generating set for the idempotent generated subsemigroup. For Rees matrix semigroups, I think the method gives a minimal generating set (it's certainly irredundant). For Rees 0-matrix semigroups, the code now makes use of my recent addition to Digraphs called `UndirectedSpanningForest`. A spanning forest of the `RZMSDigraph` corresponds to a generating set for the idempotent generated subsemigroup. Whilst this certainly does not return a irredundant generating set in general, it has the potential to return a much smaller generating set that was obtained previously.

Additionally, this PR allows me to address a long-standing TODO from my maximal subsemigroups code. At a certain part in those algorithms, to specify the generating set of a maximal subsemigroup, we require the set of idempotents of a particular Rees (0-)matrix semigroup. However, we only actually require a generating set for these idempotents. Since there is a special method installed for this now, we may as well use this.

### DIFF
--- a/gap/attributes/attr.gi
+++ b/gap/attributes/attr.gi
@@ -555,6 +555,79 @@ function(S)
   return out;
 end);
 
+InstallMethod(IdempotentGeneratedSubsemigroup,
+"for a Rees matrix subsemigroup",
+[IsReesMatrixSubsemigroup],
+function(R)
+  local mat, I, J, nrrows, nrcols, min, max, gens, out, i;
+
+  if not IsFinite(R) or not IsReesMatrixSemigroup(R)
+      or not IsGroup(UnderlyingSemigroup(R)) then
+    TryNextMethod();
+  fi;
+
+  mat := Matrix(R);
+  I := Rows(R);
+  J := Columns(R);
+  nrrows := Length(I);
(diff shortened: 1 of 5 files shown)


gap/attributes/attr.gi
```diff
@@ -555,6 +555,78 @@ function(S)
...
+  nrrows := Length(I);
+
+  for i in [1 .. nrrows] do
+    for j in gr[i] do
+      k := J[j - nrrows];
+      Add(gens, RMSElement(R, I[i], mat[k][I[i]] ^ -1, k));
+    od;
+  od;
+
+  out := RegularSemigroup(gens);
```
mtorpey: I think you decided to stop using this - certainly you replaced it in the RMS case.  Is this safe?

mtorpey:
I'm happy.  Will merge once the tests pass.

gap/attributes/attr.gi
wilfwilson: I changed it over.

PR #326, opened 2017-06-04
Extend IsomorphismSemigroups to finite simple, 0-simple, and monogenic semigroups

wilfwilson:
IsomorphismSemigroups now works fully for all finite simple, 0-simple, and monogenic semigroups. Since a Rees matrix semigroup over a group is a finite simple semigroup, and a regular Rees 0-matrix semigroup over a group is a finite 0-simple semigroup, even when the objects are not `IsWholeFamily`, and not over `IsPermGroup`, then this PR fully resolves issue #167.

For finite simple semigroups, we take an isomorphism to/from a Rees matrix semigroup with `IsomorphismReesMatrixSemigroup`, and use the stuff in `isorms.gi` to find the isomorphism between the two Rees matrix semigroups. Since `IsomorphismReesMatrixSemigroup` is not guaranteed to give a Rees matrix semigroup over a perm group (for instance, if the simple semigroup is already a Rees matrix semigroup, but not over a perm group and not `IsWholeFamily`), then I had to update `IsomorphismSemigroups` for two Rees matrix semigroups to takes isomorphisms to/from Rees matrix semigroups over perm groups, if necessary.

Something very similar happens for finite 0-simple semigroups.

For finite monogenic semigroups, it's just a case of identifying the respective generators, and then mapping an element between the two semigroups by finding the right power of the generator.

This PR also modifies `IsIsomorphicSemigroup` to simply delegate to `IsomorphismSemigroups`, and return `true` or `false` (or no method found) as appropriate. This removes the previous functionality, whereby `IsIsomorphicSemigroup` could return `true` or `false` for very small semigroups by using `SmallestMultiplicationTable`, since we don't have isomorphisms between these things implemented yet. However, I've updated the documentation for `IsIsomorphicSemigroup` to explain that `SmallestMultiplcationTable` can be used in exactly this way (as is already described in the documentation for `SmallestMultiplicationTable` itself).


gap/attributes/isomorph.gi
```diff
@@ -88,36 +88,97 @@ function(S)
...
-[IsSemigroup and HasGeneratorsOfSemigroup, IsSemigroup and
- HasGeneratorsOfSemigroup],
+InstallMethod(IsIsomorphicSemigroup, "for semigroups",
+[IsSemigroup, IsSemigroup],
 function(S, T)
-  local pS, pT, iso;
+  return IsomorphismSemigroups(S, T) <> fail;
+end);
+
+InstallMethod(IsomorphismSemigroups, "for semigroups",
```
james-d-mitchell: Split this into different methods, rather than one monolithic function

gap/attributes/isorms.gi
```diff
@@ -663,119 +663,245 @@ end);
...
+  if not (IsGroupAsSemigroup(G) and IsGroupAsSemigroup(H)) then
+    TryNextMethod();
+  elif Size(S) <> Size(T)
+      or Length(Columns(S)) <> Length(Columns(T))
+      or Length(Rows(T)) <> Length(Rows(T)) then
+    return fail;
+  fi;
+
+  # Take an isomorphism from <S> to <s> if <S> is not over a perm group
+  if not IsPermGroup(G) then
```
james-d-mitchell: IsomorphismReesMatrixSemigroupOverPermGroup

gap/attributes/isorms.gi
```diff
@@ -663,119 +663,245 @@ end);
...
+"for finite whole family Rees matrix semigroups",
+[IsReesMatrixSemigroup and IsWholeFamily and IsFinite,
+ IsReesMatrixSemigroup and IsWholeFamily and IsFinite],
+function(S, T)
+  local G, H, isoG, invG, s, first, isoH, invH, t, second, iso, mat, m, n, f,
+  isograph, isogroup, RMSInducedFunction, map, l, g, tup;
+
+  G := UnderlyingSemigroup(S);
+  H := UnderlyingSemigroup(T);
+
```
james-d-mitchell: if not IsPermGroup(G) or not IsPermGroup(H) then 
   TryNextMethod();
fi;

gap/attributes/isorms.gi
```diff
@@ -663,119 +663,245 @@ end);
...
   return fail;
 end);
 
-InstallMethod(IsomorphismSemigroups, "for Rees 0-matrix semigroups",
-[IsReesZeroMatrixSemigroup and IsWholeFamily,
- IsReesZeroMatrixSemigroup and IsWholeFamily],
-function(R1, R2)
-  local G1, G2, mat, m, n, f, groupiso, graph1, graph2, g, graphiso, tuples,
-        map, l, tup, RZMStoRZMSInducedFunction;
+InstallMethod(IsomorphismSemigroups,
```
james-d-mitchell: Same comment here as for the method for Rees matrix semigroups without zero.

gap/attributes/isorms.gi
wilfwilson: Seeing as we might be able to return `fail` in the next few lines, even if it's not over a perm group, I've put this perm group check a few lines lower down.

wilfwilson:
I think I've addressed all of your concerns @james-d-mitchell

PR #335, opened 2017-06-14
Add big IdempotentGeneratedSubsemigroup tests

wilfwilson:
I added some tests to `attr.tst` that show the small size of the generating sets now returned by `IdempotentGeneratedSubsemigroup` for a Rees (0-)matrix semigroup over a group.

### DIFF
--- a/tst/standard/attr.tst
+++ b/tst/standard/attr.tst
@@ -1662,6 +1662,25 @@ true
 gap> I = Semigroup(Idempotents(S));
 true
 
+#T# IdempotentGeneratedSubsemigroup, for R(Z)MS where new method is much better
+
+# RZMS: 2241 generators -> 384 generators
+gap> R := PrincipalFactor(
+>  DClass(FullTransformationMonoid(7),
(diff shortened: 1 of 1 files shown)


tst/standard/attr.tst
```diff
@@ -1662,6 +1662,25 @@ true
...
+>  Transformation([1, 1, 3, 4, 5, 4, 1])));;
+gap> Length(Rows(R));
+350
+gap> Length(Columns(R));
+35
+gap> IdempotentGeneratedSubsemigroup(R);
+<subsemigroup of 350x35 Rees 0-matrix semigroup with 384 generators>
+
+# RMS: 7200 generators -> 90 generators
+gap> R := RandomSemigroup(IsReesMatrixSemigroup, 80, 90, SymmetricGroup(1));
```
james-d-mitchell: Wouldn't it be safer to do `RectangularBand(IsReesMatrixSemigroup, 80, 90)` here?

wilfwilson: Yes that's exactly what I need.

wilfwilson:
I made that change. It's less code too!

PR #340, opened 2017-06-20
io: fix error messages

james-d-mitchell:
For some reason, when a file couldn’t be opened fail was displayed in
the error rather than the filename, this commit fixes this, and adjusts the 
tests.


james-d-mitchell:
@mtorpey any chance you could review this when you have minute?

PR #341, opened 2017-06-20
properties: fix IsEUnitaryInverseSemigroup

ChristopherRussell:
`IsEUnitaryInverseSemigroup` does not work for some semigroups which don't satisfy
`IsInverseSemigroup`. For example:

```
gap> S := Semigroup([Transformation([2, 2]), Transformation([2, 1, 2]),
> Transformation([3, 3, 2])]);;
gap> IsEUnitaryInverseSemigroup(S);
Error, no 2nd choice method found for `IsEUnitaryInverseSemigroup' on 1
arguments at /Users/crussell/gap/lib/methsel2.g:241 called from
<function "HANDLE_METHOD_NOT_FOUND">( <arguments> )
called from read-eval loop at *stdin*:3
you can 'quit;' to quit to outer loop, or
you can 'return;' to continue
brk> quit;
gap> IsInverseSemigroup(S);
false
```

### DIFF
--- a/gap/attributes/properties.gi
+++ b/gap/attributes/properties.gi
@@ -334,11 +334,13 @@ function(S)
   return IsMajorantlyClosed(S, IdempotentGeneratedSubsemigroup(S));
 end);
 
-InstallMethod(IsEUnitaryInverseSemigroup, "for an inverse semigroup",
-[IsInverseSemigroup],
+InstallMethod(IsEUnitaryInverseSemigroup, "for a semigroup",
+[IsSemigroup],
 function(S)
   if not IsFinite(S) then
     TryNextMethod();
+  elif not IsInverseSemigroup(S) then
+    return false;
   fi;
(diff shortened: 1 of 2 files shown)


gap/attributes/properties.gi
```diff
@@ -335,11 +335,14 @@ function(S)
 end);
 
 InstallMethod(IsEUnitaryInverseSemigroup, "for an inverse semigroup",
```
james-d-mitchell: "for an inverse semigroup" -> "for a semigroup"

gap/attributes/properties.gi
```diff
@@ -335,11 +335,14 @@ function(S)
...
-[IsInverseSemigroup],
+[IsSemigroup],
 function(S)
   if not IsFinite(S) then
     TryNextMethod();
   fi;
```
james-d-mitchell: `elif` and save a line

ChristopherRussell:
Thanks, changes made.

Issue #342, opened 2017-06-23
DirectProduct incorrectly adding identity


mtorpey:
`DirectProduct` is not behaving correctly.  It appears to be appending the identity transformation when it shouldn't.  Consider the following example (found by @ffloresbrito):

```gap
gap> S := Semigroup([Transformation([1, 2, 3, 3, 3])]);;
gap> CC := Semigroup([Transformation([2, 1])]);;
gap> Size(S); Size(CC);
1
2
gap> DP := DirectProduct(S, CC);;
gap> Elements(DP);
[ Transformation( [ 1, 2, 3, 3, 3 ] ), 
  Transformation( [ 1, 2, 3, 3, 3, 7, 6 ] ), IdentityTransformation, 
  Transformation( [ 1, 2, 3, 4, 5, 7, 6 ] ) ]
gap> Size(DP);
2
```

As you can see, the function seems to be treating `S` as though it contains `IdentityTransformation`, which it doesn't.  Hence two extra elements are ending up in the direct product, even though the size is being calculated correctly.

While I was at it, I noticed that `DirectProduct` for semigroups is not documented, meaning that the only manual entry for it is in the GAP manual, where it just talks about groups.  This should be rectified as well (I'll add it to Issue #298).

james-d-mitchell:
Hmm, something is for sure not right here...

ChrisJefferson:
What was the error?

james-d-mitchell:
Don't know but a semigroup of size 2 shouldn't have 4 elements

james-d-mitchell:
I've fixed this in PR #343

mtorpey:
Merged and fixed!

PR #343, opened 2017-06-29
Fix issue 342 (direct product of transformation semigroups)

james-d-mitchell:
The problem was that the `One` of the semigroups was used in the
generators instead of the `MultiplicativeNeutralElement`. This PR
fixes this.

### DIFF
--- a/gap/semigroups/semitrans.gi
+++ b/gap/semigroups/semitrans.gi
@@ -84,24 +84,36 @@ InstallMethod(DirectProductOp,
 "for a list and a transformation monoid as semigroup",
 [IsList, IsTransformationSemigroup], 1, # to beat the next method
 function(list, S)
-  local gens, deg, m, i, x;
+  local gens, deg, one, m, next, i, x;
 
   # Check the arguments.
   if IsEmpty(list) then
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
This resolves Issue #342.

gap/semigroups/semitrans.gi
```diff
@@ -84,24 +84,37 @@ InstallMethod(DirectProductOp,
...
   if IsEmpty(list) then
     ErrorNoReturn("Semigroups: DirectProductOp: usage,\n",
                   "the first argument must be a non-empty list,");
   elif not ForAll(list, T -> IsTransformationSemigroup(T)
-                  and IsMonoidAsSemigroup(T)) then
+                             and IsMonoidAsSemigroup(T)) then
     TryNextMethod();
   fi;
 
   gens := ShallowCopy(GeneratorsOfSemigroup(list[1]));
```
mtorpey: These first generators also need to be determined in terms of `one`.  At the moment, this assumes that all semigroups after `list[i]` have `IdentityTransformation` as their identity.

james-d-mitchell: Oh yes, thanks, will revise.

PR #348, opened 2017-07-19
Cayley digraphs

james-d-mitchell:
Replace `Left/RightCayleyGraphSemigroup` with `Left/RightCayleyDigraph` and introduce some functions for drawing these objects. This PR requires Digraphs version 0.10.0 which is not yet released, so the tests will probably fail.

These essentially just wrap the value previously returned by
`Left/RightCayleyGraphSemigroup` in a `Digraph` object. Methods for
`Left/RightCayleyGraphSemigroup` continue to exist (just calling
`OutNeighbours(Left/RightCayleyDigraph)` since they are defined in the
library. All other occurrences of `Left/RightCayleyGraphSemigroup` are
replaced with `Left/RightCayleyDigraph`.

### DIFF
--- a/doc/display.xml
+++ b/doc/display.xml
@@ -127,8 +127,23 @@
         </Item>
         <Mark>pbrs</Mark>
         <Item>
-          If <A>obj</A> is a PBR, then <C>TikzString</C> returns a graphical
-          representation <A>obj</A>; see Section <Ref Sect = "section-blocks"/>.
+          If <A>obj</A> is a <Ref Oper="PBR"/>, then <C>TikzString</C> returns a
+          graphical representation <A>obj</A>; see Chapter <Ref Sect =
+            "Partitioned binary relations (PBRs)"/>.
+        </Item>
+        <Mark>Cayley graphs</Mark>
+        <Item>
+          If <A>obj</A> is a <Ref Oper="Digraph" BookName="digraphs"/> in the
+          category <Ref Filt="IsCayleyDigraph" BookName="digraphs"/>, then
+          <C>TikzString</C> returns a picture of <A>obj</A>. No attempt is made
+          whatsoever to produce a sensible picture of the digraph <A>obj</A>,
+          in fact, the vertices are all given the same coordinates. Human
+          intervention is required to produce a meaningful picture from the
+          value returned by this method. It is intended to make the task of
+          drawing such a Cayley graph more straightforward by providing
+          everything except the final layout of the graph. Please use <Ref
+            Oper="DotString"/> if you want an automatically laid out diagram of
+          the digraph <A>obj</A>.
         </Item>
       </List>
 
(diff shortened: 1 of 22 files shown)


wilfwilson:
@james-d-mitchell This is failing because there is no code coverage of the new `Dot...` and `Tikz...` methods in `display.gi`.

james-d-mitchell:
Ooops, forgot to write the doc and the tests. Done it now.

doc/fropin.xml
```diff
@@ -183,27 +183,34 @@ true]]></Example>
...
-      <C>RightCayleyGraphSemigroup</C> returns the right
-      Cayley graphs of <A>S</A>, as a list <C>graph</C> where 
-      <C>graph[i][j]</C> is equal to 
+      <C>RightCayleyDigraph</C> returns the right
+      Cayley graphs of <A>S</A>, as a <Ref Oper="Digraph" BookName="digraphs"/>
+      <C>digraph</C> where vertex <C>OutNeighbours(digraph)[i][j]</C> is 
       <C>PositionCanonical(<A>S</A>, AsListCanonical(<A>S</A>)[i] *
         GeneratorsOfSemigroup(<A>S</A>)[j])</C>.
-      The list returned by <C>LeftCayleyGraphSemigroup</C> is defined analogously.
+      The list returned by <C>LeftCayleyDigraph</C> is defined analogously.<P/>
```
wilfwilson: This sentence implies that `<C>LeftCayleyDigraph</C>` returns a list, rather than a digraph.

gap/greens/gree.gi
```diff
@@ -391,8 +391,6 @@ function(S)
 
   l  := LeftCayleyGraphSemigroup(S);
   r  := RightCayleyGraphSemigroup(S);
-  # WW: in the future, when l and r are digraphs, gr can be created
-  #     by using DigraphEdgeUnion(l, r)
```
wilfwilson: Is there a reason why you didn't implement this suggestion?

doc/display.xml
```diff
@@ -288,6 +303,30 @@ gap> FileString("t3.dot", DotString(S));
...
 <#/GAPDoc>
 
+<#GAPDoc Label="DotStringDigraph">
+  <ManSection>
+    <Oper Name="DotString" Arg="digraph" Label="for a Cayley digraph"/>
+    <Returns>A string.</Returns>
+    <Description>
+      If <A>digraph</A> is a <Ref Oper="Digraph" BookName="digraphs"/> in the
+      category <Ref Filt="IsCayleyDigraph" BookName="digraphs"/>, then
+      <C>DotString</C> returns a graphical representation of <A>digraph</A>.
```
wilfwilson: I think you should be more descriptive about what the representation actually shows.

doc/fropin.xml
james-d-mitchell: Corrected

gap/greens/gree.gi
james-d-mitchell: Yes, because there are no methods for `Left/RightCayleyDigraph` for semigroups not in the category `IsEnumerableSemigroupRep` and so we use `Left/RightCayleyGraphSemigroup` which are defined in the library and hence not digraphs. So, basically the comment does not make sense in this method, although it does in the method for enumerable semigroups. Hence I deleted it.

doc/display.xml
james-d-mitchell: We agreed this was not necessary.

james-d-mitchell:
Updated according to your comments @wi

wilfwilson:
Travis is failing but it's not related to this PR. I'm trying to see if I can sort it out before merging.

PR #350, opened 2017-07-26
Fixed dead links

Alasdair-Macindoe:
It seems that all the URLs were pointed to gap-systems.github.io instead of the correct gap-packages.github.io. There is a legitimate argument to be made that they should point to the gap-systems.org pages, but I felt like the gap-packages.github.io would be better since you are looking to install the package and not necessarily read about it.

Additionally: It displays a lot more lines as edited than actually were edited, only 23-33 were edited.


james-d-mitchell:
Thanks for this it is very welcome. Can you please drop the changes to the lines that only contain whitespace changes? Then I'll be happy to merge this.

james-d-mitchell:
This is fixed by PR #354 which did not have the superfluous changes included in this one.

PR #354, opened 2017-08-10
Fix broken links in README.md

ssiccha:


ssiccha:
Ah, just saw that its a duplicate of #350 (but without the whitespaces).

james-d-mitchell:
Thanks for this!

Issue #356, opened 2017-08-15
Performance regression related to Cayley digraphs


wilfwilson:
```
M6 := Semigroup([
   Matrix(IsNTPMatrix, [[0, 0, 1], [0, 1, 0], [1, 1, 0]], 0, 6),
   Matrix(IsNTPMatrix, [[0, 0, 1], [0, 1, 0], [2, 0, 0]], 0, 6),
   Matrix(IsNTPMatrix, [[0, 0, 1], [0, 1, 1], [1, 0, 0]], 0, 6),
   Matrix(IsNTPMatrix, [[0, 0, 1], [0, 1, 0], [3, 0, 0]], 0, 6)]);
NrDClasses(M6);
```
The answer to this block of code is 16; `M6` is the monoid consisting of all 3x3 matrices over Z/6Z.

Before PR #348 was merged (in `v3.0.3`, for instance), this computation took about 28 seconds. Afterwards (in the current version of `master`, for instance), it takes about 42 seconds.

james-d-mitchell:
Just to double-check, everything is compiled the same way in both cases right?

wilfwilson:
Yes. Michael and I used `git bisect` to change between Semigroups changesets; we used the released version of libsemigroups and ran `make` every time we checked out a new Semigroups changeset.

mtorpey:
This is now fixed, and the timing looks better than ever.

PR #360, opened 2017-08-23
Fix issue #356

james-d-mitchell:
Previously we did `Digraph(EN_SEMI_RIGHT_CAYLEY_GRAPH(S))` which involved
copying the output of `EN_SEMI_RIGHT_CAYLEY_GRAPH(S)`, and performing
checks that the output was valid. We make the output of
`EN_SEMI_RIGHT_CAYLEY_GRAPH(S)` immutable, and use `DigraphNC` to avoid
these unnecessary checks. This seems to resolve Issue #356.


wilfwilson:
Thanks James.

Issue #363, opened 2017-08-30
Bug in MultiplicativeNeutralElement for ideals


wilfwilson:
```
gap> S := SingularFactorisableDualSymmetricInverseMonoid(3);
<inverse bipartition semigroup ideal of degree 3 with 1 generator>
gap> IsMonoidAsSemigroup(S);
true
```
This should return false. I'm fixing it.

james-d-mitchell:
Yikes, good catch.

wilfwilson:
Resolved by PR #364

PR #364, opened 2017-08-30
Fix MultiplicativeNeutralElement for ideals

wilfwilson:
This PR resolves issue #363.

Suppose that we suspect that `e` is the multiplicative neutral element of `S`. The bug was caused by testing whether `e` acts as the identity for all `Generators(S)`. However this isn't sufficient when `S` is an ideal. We need to use `GeneratorsOfSemigroup(S)` instead. I also add a more thorough check before calling `GeneratorsOfSemigroup(S)`, so that we avoid doing it for ideals if we can.

### DIFF
--- a/gap/attributes/attract.gi
+++ b/gap/attributes/attract.gi
@@ -256,8 +256,9 @@ function(S)
 
   r := GreensRClassOfElementNC(S, rep);
 
-  if not NrIdempotents(r) = 1 then
-    Info(InfoSemigroups, 2, "the number of idempotents in the R-class of the",
+  if NrIdempotents(r) <> 1
+      or NrIdempotents(GreensDClassOfElementNC(S, rep)) <> 1 then
+    Info(InfoSemigroups, 2, "the number of idempotents in the D-class of the",
          " first maximum rank");
     Info(InfoSemigroups, 2, " generator is not 1");
     return fail;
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
This looks good. One slight concern is that `GeneratorsOfSemigroup` for an ideal can be rather expensive, I can't currently think of a good way of checking if something is an identity of an ideal that doesn't involve this though. Perhaps that part of the code can be split into the current check if the semigroup has `GeneratorsOfSemigroup` and into some thing that checks if the idempotent has trivial action on the `LambdaOrb` and `RhoOrb` of the ideal? I think that's sufficient, and should be significantly easier than computing the generators if they are not already known. What do you think?

wilfwilson:
I think you should merge this PR, since it fixes the bug and it's better to have bad performance for ideals rather than incorrect results. The performance problem should be an issue and dealt with in a separate PR.

I think your idea works. Do you mean verifying `ForAll(LambdaOrb(I), x -> LambdaAct(I)(x, e) = x)`, and the equivalent for the RhoOrb(S)?

james-d-mitchell:
Yes, that's what I mean. Ok, I will merge, could you please file an issue about the performance?

wilfwilson:
Sure

james-d-mitchell:
Tah

PR #367, opened 2017-08-31
semieunit: fix IsMcAlisterTripleSemigroupElementRep declaration

ChristopherRussell:
There is also a change to the comments describing this representation
and a commented out method has been removed in semieunit.gi

### DIFF
--- a/gap/semigroups/semieunit.gd
+++ b/gap/semigroups/semieunit.gd
@@ -14,18 +14,18 @@ DeclareCategory("IsMcAlisterTripleSemigroupElement",
 # This is a representation for McAlister triple semigroup elements, which are
 # created via the function McAlisterTripleSemigroupElement.
 #
-# The components are:
+# If x belongs to the representation IsMcAlisterTripleElementRep, then the
+# components are:
 #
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
In the comment about the representation, please refer explicitly to the positions, i.e. 

> If x belongs to the representation X, then 
- x![1] is .. 
- x![2] is ..
- x![3] is ..

instead of using `Parent` etc.

ChristopherRussell:
I changed it to this:

```
# This is a representation for McAlister triple semigroup elements, which are
# created via the function McAlisterTripleSemigroupElement.
#
# If x belongs to the representation X, then the components are:
#
#   x![1]:   The McAlister triple semigroup which this element belongs to
#
#   x![2]:   A vertex of the McAlisterTripleSemigroupSemilattice of x![1]
#
#   x![3]:   An element of the McAlisterTripleSemigroupGroup of x![1]

```

james-d-mitchell:
But where X is replaced with `IsMcAlisterTripleElementRep`? ???

ChristopherRussell:
Sorry. Changed it.

Issue #366, opened 2017-08-31
Improve performance of MultiplicativeNeutralElement for ideals


wilfwilson:
For acting ideals, the method for MultiplicativeNeutralElement currently requires (in some cases) that `GeneratorsOfSemigroup` be called on the ideal. This can be costly. @james-d-mitchell suggested in the thread of PR #364 an improved method, which:

> checks if the idempotent has trivial action on the LambdaOrb and RhoOrb of the ideal

if the `GeneratorsOfSemigroup` of the ideal are not already known.

james-d-mitchell:
Please let me know if you are going to implement this @wilfwilson, otherwise, I might later today.

wilfwilson:
I'll have a quick go at it now. Do you want it submitted to `master` or `stable-3.0`?

james-d-mitchell:
stable

james-d-mitchell:
Resolved by PR #368.

PR #368, opened 2017-08-31
Improve MultiplicativeNeutralElement for ideals

wilfwilson:
This PR attempts to resolve issue #366, in the way indicated on that issue.

### DIFF
--- a/gap/attributes/attract.gi
+++ b/gap/attributes/attract.gi
@@ -234,7 +234,7 @@ end);
 InstallMethod(MultiplicativeNeutralElement, "for an acting semigroup",
 [IsActingSemigroup],
 function(S)
-  local gens, rank, lambda, max, rep, r, e;
+  local gens, rank, lambda, max, rep, r, e, lo, ro, lact, ract;
 
   gens := Generators(S);
   rank := LambdaRank(S);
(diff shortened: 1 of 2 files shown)


gap/attributes/attract.gi
```diff
@@ -256,18 +256,32 @@ function(S)
...
     return fail;
   fi;
 
   e := Idempotents(r)[1];
 
-  if ForAll(GeneratorsOfSemigroup(S), x -> x * e = x and e * x = x) then
-    return e;
+  if HasGeneratorsOfSemigroup(S) then
+    if ForAll(GeneratorsOfSemigroup(S), x -> x * e = x and e * x = x) then
+      return e;
```
james-d-mitchell: Why not combine these using `and` rather than having nested if's?

gap/attributes/attract.gi
```diff
@@ -256,18 +256,32 @@ function(S)
...
+    fi;
+  elif ForAll(gens, x -> x * e = x and e * x = x) then
```
james-d-mitchell: Why not do the negation of this and return `false`, to avoid having such deeply nested `if`s?

gap/attributes/attract.gi
```diff
@@ -256,18 +256,32 @@ function(S)
...
+    # S is an ideal without GeneratorsOfSemigroup
+    o := LambdaOrb(S);
+    Enumerate(o);
+    act := LambdaAct(S);
+    if ForAll([2 .. Length(o)], i -> act(o[i], e) = o[i]) then
```
james-d-mitchell: Same, why not negate and return false,

wilfwilson: Because you've put in `InfoStatement`s to the method that won't be triggered if I return early

wilfwilson: @james-d-mitchell Would you prefer me to remove the InfoStatement at the end of the method
```
  Info(InfoSemigroups, 2, "the unique idempotent in the D-class of the first",
                          " maximum rank");
  Info(InfoSemigroups, 2, " generator is not the identity");
  return fail;
```
and then restructure the code, with more `return fail`s, to reduce the nesting level?

gap/attributes/attract.gi
wilfwilson: Because if it does have generators of semigroup but doesn't satisfy the `ForAll`, then
it will proceed to perform the calculation using the `LambdaOrb` and `RhoOrb`.

gap/attributes/attract.gi
james-d-mitchell: Yes probably, if you can be bothered!

wilfwilson:
@james-d-mitchell Done

james-d-mitchell:
Nice

PR #374, opened 2017-09-13
Add One to GeneratorsOfMagma if needed in MonoidByGenerators

### LINKED ISSUE #371: Incorrect behaviour for a monoid created by MonoidByAdjoiningIdentity
In the GAP `lib/adjoin.gd` file, `MonoidByAdjoiningIdentity` is declared as an attribute that adjoins an identity to a semigroup (even if it already has one). Without the Semigroups package loaded, this is the following correct behaviour:
```
gap> R := ReesMatrixSemigroup(Group(()), [[()]]);;
gap> S := MonoidByAdjoiningIdentity(R);
gap> Size(S);
2
gap> Elements(S);
[ ONE, (1,(),1) ]
```
With the Semigroups package loaded, this is what we get:
```
gap> R := ReesMatrixSemigroup(Group(()), [[()]]);;
gap> S := MonoidByAdjoiningIdentity(R);
<commutative monoid with 1 generator>
gap> Size(S);
1
gap> Elements(S);
[ (1,(),1) ]
```
I've not dug into what's going on.

wilfwilson:
The `One` of a monoid needs to be in its `GeneratorsOfMagma`, since we don't know (without further work) whether the other elements generate (as a semigroup) the `One`.

Previously this was only done if the generators satisfy `CanEasilyCompareElements` and the `One` isn't already in there. If we can't easily compare elements, we still need to add the `One` if needed. Since it's too expensive to work out whether the `One` is already one of the generators, we should add the `One` regardless.

This resolves #371

### DIFF
--- a/gap/semigroups/semigrp.gi
+++ b/gap/semigroups/semigrp.gi
@@ -277,7 +277,7 @@ function(gens, opts)
 
   S := Objectify(NewType(FamilyObj(gens), filts), rec(opts := opts));
 
-  if CanEasilyCompareElements(gens) and not One(gens) in gens then
+  if not CanEasilyCompareElements(gens) or not One(gens) in gens then
     SetGeneratorsOfMagma(S, Concatenation([One(gens)], gens));
   else
     SetGeneratorsOfMagma(S, AsList(gens));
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Looks good to me, but it looks like this should be rebased since I merged your other PR #378 a moment ago.

Issue #353, opened 2017-08-04
Check for presence of libsemigroups


markuspf:
When building Semigroups without a copy of `libsemigroups` in the `src` folder one gets a rather unhelpful error message about a missing file (in my case `blocks.cc`).

It would be best to add a check to `configure.ac` that makes sure that `src/libsemigroups/src/semigroups.cc` (for example) is present, and if it is not prints a helpful error message.

james-d-mitchell:
Resolved by PR #387

PR #387, opened 2017-09-28
Add a --enable-debug configure option and add warning if libsemigroups is not available 

james-d-mitchell:


james-d-mitchell:
This PR requires libsemigroups 0.5.1 (not yet released)

james-d-mitchell:
libsemigroups 0.5.1 is out, and so this can be merged if the CI pass.

wilfwilson:
Now that libsemigroups 0.5.1 is out, Travis will be using 0.5.1 and so the tests for all the pull requests that are open at the moment will fail until this is merged into stable-3.0, and stable-3.0 is merged into master.

I think we need to avoid this problem. Currently Travis selects the newest released version of libsemigroups, and uses that. Instead, I think we need to have a variable in the Travis setup, which specifies which version of libsemigroups to use. Then, when you release a new version of libsemigroups, you should make your usual pull request to make Semigroups compatible with the newest version of libsemigroups - and as part of that PR, you should update that token in the `.travis.yml` file to the newest version of libsemigroups.

Sound like a good idea?

james-d-mitchell:
Yes, @wilfwilson this does sound like a good idea.

james-d-mitchell:
This resolves Issue #353.

PR #388, opened 2017-09-30
Set `IdempotentGeneratedSubsemigroup` to be a semilattice, if possible

wilfwilson:
If a semigroup has commuting idempotents, then its idempotent generated subsemigroup is commutative, and is therefore a band. In other words, it's a semilattice.

We should set this information if we already know that the semigroup has commuting idempotents (for instance, if it's any partial perm semigroup).

### DIFF
--- a/gap/attributes/attr.gi
+++ b/gap/attributes/attr.gi
@@ -557,6 +557,10 @@ function(S)
   fi;
   out := Semigroup(Idempotents(S), rec(small := true));
   SetIsIdempotentGenerated(out, true);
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
@wilfwilson I've changed this to go into `stable-3.0`, are you ok with that?

wilfwilson:
Either branch is fine by me.

james-d-mitchell:
Yeah, I tried to rebase it here, but that drags in all the changes from master, which we don't want.

wilfwilson:
Sorted. Travis will still fail because of #389.

Issue #391, opened 2017-10-01
Semigroups breaks testinstall/fpmon.tst in GAP


alex-konovalov:
The following happens when Semigroups package is loaded:
```
gap> Test("tst/testinstall/fpmon.tst");
# line 20 of 54 (37%)terminate called after throwing an instance of 'std::system_error'
  what():  Enable multithreading to use std::thread: Operation not permitted
Aborted
```

james-d-mitchell:
Which version please?

james-d-mitchell:
Also which compiler and version, what version of GAP, and the config.log if you can provide it? Thanks @alex-konovalov

alex-konovalov:
@james-d-mitchell: master branch of GAP, semigroups 3.0.6, gcc 4.8.5, and config.log as attached.

[config.log](https://github.com/gap-packages/Semigroups/files/1347541/config.log)

james-d-mitchell:
Thanks@alex-konovalov

james-d-mitchell:
I cannot reproduce this, `fpmon.tst` runs just fine.

alex-konovalov:
Test passes in 64-bit build on Mac (Apple LLVM version 8.0.0 (clang-800.0.42.1). The test fails on kovacs in both 32- and 64-bit builds.  @markuspf would you be able to put some newer version of gcc there, please?

james-d-mitchell:
What system is kovac running? @alex-konovalov ?

alex-konovalov:
@james-d-mitchell I am afraid Semigroups also breaks `testinstall/semigrp.tst`. Could you add it to the test as well? (Since your regression tests have GAP around anyway, just loading a package and running `testinstall.g` from GAP could be a good idea).

kovacs has Scientific Linux 7.3

james-d-mitchell:
Ok, noted. These look like harmless diffs except one, which is an error.

ChrisJefferson:
This isn't an old compiler problem, it's a mac/linux problem. Macs always link the pthread automatically, Linux needs a flag (`-pthread`) I think.

james-d-mitchell:
Yes, but everything works with no problems in our travis tests for Semigroups (which use Ubuntu), and -pthread is one of the compiler flags specified in the autotools setup of Semigroups. So, missing compiler/link flags is not the issue either.

james-d-mitchell:
[...]

PR #392, opened 2017-10-02
Add GAP testinstall/fpmon.tst

james-d-mitchell:
This is part of an attempt to reproduce the error reported in Issue #391.


james-d-mitchell:
This was only ever intended to serve as an example that Semigroups does not break the tests in `testinstall/fpmon.tst` on linux.

Issue #395, opened 2017-10-02
Semigroups breaking tst/testinstall/semigrp.tst and semitran.tst


james-d-mitchell:

wilfwilson:
@james-d-mitchell I have already started (and basically finished) work on fixing the diffs in `tst/testinstall/semigrp.tst`, except for the one about congruences.

james-d-mitchell:
I've just fixed the other error about congruences.

james-d-mitchell:
Resolved by [GAP PR 1757](https://github.com/gap-system/gap/pull/1757)

PR #396, opened 2017-10-02
Add generic method for NrEquivalenceClasses for a semigroup congruence

james-d-mitchell:
This commit fixes a diff in GAP's `tst/testinstall/semigrp.tst`, caused by
there being not method for `NrEquivalenceClasses` for some types of
congruences. This commit provides a fallback method for
`NrEquivalenceClasses`.

This addresses part of Issue #395.

### DIFF
--- a/gap/congruences/cong.gi
+++ b/gap/congruences/cong.gi
@@ -21,6 +21,14 @@
 ## cong.gd contains declarations for many of these.
 ##
 
+# Fallback method
+
+InstallMethod(NrEquivalenceClasses, "for a semigroup congruence",
+[IsSemigroupCongruence],
+function(cong)
+  return Length(EquivalenceClasses(cong));
+end);
+
 InstallMethod(\in,
 "for dense list and left semigroup congruence",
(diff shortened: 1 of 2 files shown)


gap/congruences/cong.gi
```diff
@@ -21,6 +21,14 @@
 ## cong.gd contains declarations for many of these.
 ##
 
+# Fallback method
+
+InstallMethod(NrEquivalenceClasses, "for a semigroup congruence",
+[IsSemigroupCongruence], 
+function(cong)
+  return Length(EquivalenceClasses(cong));
```
wilfwilson: If you really wanted, you could make this a lambda function.

james-d-mitchell:
Right, I'll update it now

gap/congruences/cong.gi
wilfwilson: I'm happy with it as it is.

james-d-mitchell:
Merge it @wilfwilson ?

PR #400, opened 2017-10-02
Run testinstall.g from GAP in the Travis setup

wilfwilson:
We should run some of the GAP tests so that we make sure we're not breaking things in GAP when the Semigroups package is loaded. As a start, we can try running testinstall.g. In the future, maybe we'd want to consider running more of the GAP tests or manual examples.

This will fail until my related PR to GAP is merged, and my related PR to Semigroups is merged, and this is either merged in, or rebased on top of that.

### DIFF
--- a/scripts/travis-test.sh
+++ b/scripts/travis-test.sh
@@ -22,48 +22,45 @@ elif [ ! -z "$GAP_BRANCH" ]; then
 
   # Work out the Digraphs versions/branches in which to run the tests
   if [ ! -z "$DIGRAPHS_BR" ]; then
-    DIGRAPHS_BRANCHES=($DIGRAPHS_BR)
+    DIGRAPHS_BRANCH=$DIGRAPHS_BR
   else
     cd $GAP_DIR/pkg/semigroups
-    DIGRAPHS_VERS=v`grep "\"digraphs\"" PackageInfo.g| awk -F'"' '{print $4}' | cut -c3-`
-    DIGRAPHS_BRANCHES=("master" "$DIGRAPHS_VERS")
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
As far as I can tell this actually passed, but somehow was reported as a failure. Maybe I missed something in the log or maybe the setup is slightly off.

wilfwilson:
I'll check it out.

wilfwilson:
`testinstall.g` prints out the message

`#I  No errors detected while testing`

and we grep for the string "error", which is setting it off.

alex-konovalov:
@wilfwilson you can run testinstall.g in a separate entry in the Travis test matrix. It will finish with the proper error code itself in case of any errors.

wilfwilson:
Yes, we could do that. For now, I'm happy to keep it with the other suites, to keep the tests as short as possible. Every time you add another entry to the test matrix, you have the overhead of setting up the system, and fetch and compiling GAP.

For now, I've removed grepping for "error", and replaced that with "Errors detected". It seems to work fine now.

In the longer term, we should move to a more robust setup, one that doesn't involve grepping for particular strings, and instead makes proper use of error codes as much as possible. In removing the string "error", perhaps some problems will now go undetected, but how are we to know what particular failures that string is connected to? I can guess that the "FAILED" string is for gaplint, and the "Diff" string is for diffs in the tests, but otherwise, it's harder to know.

We should also make use of the new GAP docker images, so that we don't have to clone and compile GAP and its packages in each Travis instance.

I have also changed the setup so that we only run the tests with the required version of Digraphs; previously we ran them with both the required and master version of Digraphs, but I think that was adding too much time to the overall tests. Since we all (presumably) use the master version of Digraphs, I reckon using the required version of Digraphs in Travis is the way to go.

james-d-mitchell:
@wilfwilson I agree with all of your comments, time permitting we'll make these changes in the future!

PR #347, opened 2017-07-14
Improved direct products, with new supporting methods

wilfwilson:
This PR includes several related changes. Most importantly:

* this PR introduces a new method for `DirectProductOp`, for a list of either transformation semigroups, or bipartition semigroups, or partial perm semigroups.

Previously, `DirectProductOp` methods existed for:
* a list of transformation monoids
* a list of transformation monoids-as-semigroups
* an arbitrary list of transformation semigroups
* a list of partial perm inverse monoids

This PR extends this functionality to support all partial perm semigroups, and all bipartition semigroups. It does this with a unified function. Notably, this function replaces the previous method for arbitrary finite transformation semigroups, which used a brute force search to find a generating set for the direct product.

My new method sometimes returns a semigroup with a larger generating set than the previous method, but it finds the answer far faster. To see an example of improved performance, try calculating the direct product of the singular transformation semigroup of degree 4 with itself.  This takes 40ms with the new code, and gives a semigroup with 96 generators. The old code takes 4800ms, although it does only use 53 generators. Try a higher degree and the time difference is far starker. The generators returned by the new method can be reduced by `SmallGeneratingSet`, or similar, if desired.

I have ideas for reducing the number of generators returned my method, but they're still only ideas, and in any case, they would require my (future) pull request concerning the partial order of L- and R-classes of a semigroup, and in particular, maximal L- and R-classes.

The new direct product code is reliant on the new methods called:

* `NonTrivialFactorization`.

Also introduced are the related new functions:

* `IsDecomposableSemigroup` (i.e. is `S^2 = S`?)
* `IndecomposableElements` (i.e. those elements in `S \ S^2`)

The indecomposable elements of `S` are precisely those elements in `GeneratorsOfSemigroup(S)` without a non-trivial factorization in the semigroup.

PR #407, opened 2017-10-20
Add some basic methods for MinimalSemigroupGeneratingSet

wilfwilson:
It is going to be very difficult to implement completely general methods for `MinimalSemigroupGeneratingSet` that have reasonable performance. However, I think we should still write methods that apply in special cases.

Here I've added several of those special cases. The `IndecomposableElements` attribute added in PR #347 was useful for this.

The biggest class of semigroups to which the new method applies is probably finite J-trivial semigroups.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -332,33 +332,21 @@ gap> Length(SmallGeneratingSet(S));
 <ManSection>
   <Attr Name="MinimalSemigroupGeneratingSet" Arg="S"/>
   <Attr Name="MinimalMonoidGeneratingSet" Arg="S"/>
-  <Attr Name="MinimalInverseSemigroupGeneratingSet" Arg="S"/>
-  <Attr Name="MinimalInverseMonoidGeneratingSet" Arg="S"/>
   <Returns>A minimal generating set for a semigroup.</Returns>
   <Description>
-
-    <B>Warning:</B> currently, no methods are installed to compute these
-    attributes.<P/>
-
-    The attributes <C>MinimalXGeneratingSet</C> return a minimal generating set
+    The attribute <C>MinimalXGeneratingSet</C> returns a minimal generating set
     for the semigroup <A>S</A>, with respect to length. The returned value of
     <C>MinimalXGeneratingSet</C>, where applicable, is a minimal-length list
     of elements of <A>S</A> with the property that
     <Log>
       X(MinimalXGeneratingSet(S)) = S;
     </Log>
-    where <C>X</C> is any of
-    <Ref Func="Semigroup" BookName="ref"/>,
-    <Ref Func="Monoid" BookName="ref"/>,
-    <Ref Func="InverseSemigroup" BookName="ref"/>, or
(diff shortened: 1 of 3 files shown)


gap/attributes/attr.gi
```diff
@@ -960,3 +960,87 @@ function(S)
...
+  fi;
+
+  # An irredundant generating set of a finite semigroup that contains at most
+  # one generator per D-class is minimal.
+  if IsFinite(S) and (IsDTrivial(S) or
+      Length(Set(gens, x -> DClass(S, x))) = Length(gens)) then
+    return gens;
+  fi;
+
+  ErrorNoReturn("not yet implemented,");
```
james-d-mitchell: This error message does not have the correct format.

doc/attr.xml
```diff
@@ -337,8 +337,9 @@ gap> Length(SmallGeneratingSet(S));
   <Returns>A minimal generating set for a semigroup.</Returns>
   <Description>
 
-    <B>Warning:</B> currently, no methods are installed to compute these
-    attributes.<P/>
```
james-d-mitchell: I'm confused, this documentation existed before, but there are no methods installed? Would it not make more sense to not document (or indeed declare) attributes where there are no methods installed? Maybe I misunderstood something here, but this seems like an oddity to me.

james-d-mitchell: And the example below seems to contradict the assertion that there are no methods installed for `MinimalInverseMonoidGeneratingSet`. The distinction between there being a method that "computes" a minimal generating set, and one that simply knows (like a monogenic inverse semigroup) a minimal generating set is probably too subtle. I suggest improving this warning to make it clearer what is going on.

james-d-mitchell:
@wilfwilson any comments on my comments?

wilfwilson:
Yes, I'll address it and hopefully make the changes at the next Wednesday meeting. A minimal monoid generating set is simply a minimal semigroup generating set, except you throw away the identity if present (except possibly for partial perms...). I'll have to think and talk to you about minimal inverse semigroup generating sets.

I'm happy to not document the attributes that genuinely have no methods installed. But it is useful to declare them so that we can set them when we create things that we know have minimal (whatever) generating sets: eg if we create a monogenic inverse monoid, for instance.

wilfwilson:
I think I've addressed your changes. I've added methods for `MinimalMonoidGeneratingSet`. We decided not to do a fallback brute-force method, and I'm not immediately should what is the right thing to do for inverse semigroups.

wilfwilson:
The Travis tests are failing because of the coverage job, even though it shouldn't be. I think it's a problem with the profiling package. I'll investigate.

wilfwilson:
I've changed Travis to use the master version of `profiling`. That seems to fix the tests. I'm happy for this to be reviewed again @james-d-mitchell and/or merged.

james-d-mitchell:
@wilfwilson happy to merge this if you rebase

wilfwilson:
Done, thanks @james-d-mitchell.

PR #408, opened 2017-11-01
Use SET_TYPE_POSOBJ to set type of POSOBJ

fingolfin:
... instead of trying to assign to TYPE_POSOBJ, which will become impossible
in a future version of the GAP kernel.


wilfwilson:
Thanks @fingolfin. I've cherry-picked this onto our `stable-3.0` branch, which will get merged into `master`. So I'll close this now.

PR #410, opened 2017-11-02
Fix immediate monogenic methods in prop.gi

wilfwilson:
Previously the immediate methods for `IsMonogenicSemigroup` and `IsMonogenicMonoid` called the line `CanEasilyCompareElements(S)`, for the semigroup `S`, when I actually wanted to know whether I could compare the elements of the semigroup `S`. So there was a block of code that never got executed. I fixed this (accidentally in the `master` branch), but now that the block of code is executed, it is too slow, and is causing a performance regression in `tst/standard/semitrans.tst`.

I've now made all of the appropriate changes.

Once this is merged, `stable-3.0` should be merged into `master`, and then the performance regression (that currently is only present in `master`) will be resolved.

### DIFF
--- a/gap/attributes/properties.gi
+++ b/gap/attributes/properties.gi
@@ -775,11 +775,12 @@ function(S)
   local gens;
 
   gens := GeneratorsOfSemigroup(S);
-  if CanEasilyCompareElements(S) then
-    gens := DuplicateFreeList(gens);
-  fi;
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Did you check that `semitrans.tst` is back to its old run time?

wilfwilson:
Yes

PR #411, opened 2017-11-07
Brandt semigroups

james-d-mitchell:
This PR adds a `BrandtSemigroup` constructor in the same vein as `TrivialSemigroup` and chums.

### DIFF
--- a/doc/semicons.xml
+++ b/doc/semicons.xml
@@ -196,6 +196,69 @@ true]]></Example>
   </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="BrandtSemigroup">
+  <ManSection>
+    <Func Name="BrandtSemigroup" Arg="[[filt, ]G, ]n"/>
+    <Returns>
+      An <A>n</A> by <A>n</A> Brandt semigroup over the group <A>G</A>.
+    </Returns>
+      <Description>
+        If <A>n</A> is a positive integer, then this function returns an
+        <A>n</A> by <A>n</A> Brandt semigroup over the group <A>G</A>
+        in the representation given by the filter <A>filt</A>. <P/>
+
+        The optional argument <A>filt</A> can be any of the following:
+        <List>
+          <Item><C>IsPartialPermSemigroup</C>
+            (the default, if <A>filt</A> is not specified),</Item>
+          <Item><C>IsReesZeroMatrixSemigroup</C>,</Item>
+          <Item><C>IsTransformationSemigroup</C>,</Item>
+          <Item><C>IsBipartitionSemigroup</C>,</Item>
+          <Item><C>IsPBRSemigroup</C>,</Item>
+          <Item><C>IsBooleanMatSemigroup</C>,</Item>
+          <Item><C>IsNTPMatrixSemigroup</C>,</Item>
+          <Item><C>IsMaxPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsMinPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsTropicalMaxPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsTropicalMinPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsProjectiveMaxPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsIntegerMatrixSemigroup.</C></Item>
+        </List>
+        
+        The optional argument <A>G</A> defaults to a trivial permutation group.
+        If present <A>G</A> must be a permutation group, unless <A>filt</A> is
+        <C>IsReesZeroMatrixSemigroup</C> when <A>G</A> may be any type of
+        finite group. 
+        <P/>
+        
+        See <Ref Prop="IsBrandtSemigroup"/> for more information about Brandt
+        semigroups.
+
+        <Example><![CDATA[
+gap> S := BrandtSemigroup(5);
+<0-simple inverse partial perm semigroup of rank 5 with 4 generators>
+gap> IsBrandtSemigroup(S);
+true
+gap> S := BrandtSemigroup(IsTransformationSemigroup, 15);
+<0-simple transformation semigroup of degree 16 with 28 generators>
+gap> Size(S);
+226
+gap> MultiplicativeZero(S);
+Transformation( [ 16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16, 16,
+  16, 16, 16 ] )
+gap> S := BrandtSemigroup(Group((1, 2)), 3);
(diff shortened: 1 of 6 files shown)


james-d-mitchell:
I expect the CI will fail since Orb is still broken...

wilfwilson:
Please rebase on `origin/master`. I've made some changes to the Travis setup so that it should pass now.

wilfwilson:
I think the tests are now only failing because of the formatting of the manual example.

gap/semigroups/semicons.gi
```diff
@@ -703,3 +705,97 @@ function(arg)
...
+    S := BrandtSemigroupCons(arg[1], Group(()), arg[2]);
+  elif Length(arg) = 2 and IsGroup(arg[1]) and IsPosInt(arg[2]) then
+    S := BrandtSemigroupCons(IsPartialPermSemigroup, arg[1], arg[2]);
+  elif Length(arg) = 3 and IsOperation(arg[1]) and IsGroup(arg[2])
+      and IsPosInt(arg[3]) then
+    S := BrandtSemigroupCons(arg[1], arg[2], arg[3]);
+  else
+    ErrorNoReturn("Semigroups: BrandtSemigroup: usage,\n",
+                  "the arguments must be a positive integer or ",
+                  "a filter and a positive integer, or a perm group and ",
```
wilfwilson: In the package, it seems to be our policy to add line breaks in error messages, to avoid going over 80 characters per line. Was this a deliberate choice to not put in the `\n` characters? I'm not particularly bothered what our policy is, but if this is our policy, then we should stick to it.

gap/semigroups/semicons.gi
```diff
@@ -703,3 +705,97 @@ function(arg)
...
+[IsPartialPermSemigroup, IsPermGroup, IsPosInt],
+function(filter, G, n)
+  local gens, one, m, i, x;
+
+  gens := [];
+
+  if IsTrivial(G) then
+    if n = 1 then
+      Add(gens, PartialPerm([1], [1]));
+      Add(gens, EmptyPartialPerm());
```
wilfwilson: You have this line twice. You could get rid of this line here, and then move the later block of code
```
    if n = 1 then
      Add(gens, EmptyPartialPerm());
    fi;
```
to occur immediately before the line `if IsTrivial(G) then`. Saves a line and I think it's neater.

gap/semigroups/semicons.gi
```diff
@@ -703,3 +705,97 @@ function(arg)
...
+    S := BrandtSemigroupCons(arg[1], arg[2], arg[3]);
+  else
+    ErrorNoReturn("Semigroups: BrandtSemigroup: usage,\n",
+                  "the arguments must be a positive integer or ",
+                  "a filter and a positive integer, or a perm group and ",
+                  "positive integer, or a filter, perm group, and positive ",
+                  "integer,");
+  fi;
+  SetIsZeroSimpleSemigroup(S, true);
+  SetIsBrandtSemigroup(S, true);
```
wilfwilson: We need to think carefully about what happens when the group is infinite. More later...

gap/semigroups/semicons.gi
```diff
@@ -703,3 +705,97 @@ function(arg)
...
+      for i in [2 .. n] do
+        Add(gens, PartialPerm([1], [i]));
+      od;
+    fi;
+  else
+    one := PartialPerm(MovedPoints(G), MovedPoints(G));
+    for x in GeneratorsOfGroup(G) do
+      Add(gens, one * x);
+    od;
+    m := LargestMovedPoint(G);
```
wilfwilson: I'd recommend using `m := LargestMovedPoint(G) - SmallestMovedPoint(G) + 1;` here.

With this change, if `G := Group((4,5));` and `n = 2`, then instead of `gens` being `(4,5), [4,9][5,10]`, they'd be `(4,5), [4,6][5,7]`. Same number of moved points, but smaller numbers overall.

gap/semigroups/semicons.gi
```diff
@@ -703,3 +705,97 @@ function(arg)
...
+end);
+
+InstallMethod(BrandtSemigroupCons,
+"for IsReesZeroMatrixSemigroup and a positive integer",
+[IsReesZeroMatrixSemigroup, IsGroup, IsPosInt],
+function(filter, G, n)
+  local mat, i;
+  mat := [];
+  for i in [1 .. n] do
+    Add(mat, [1 .. n] * 0);
```
wilfwilson: Now that https://github.com/gap-system/gap/pull/1786 has been merged, I think you should use
`ListWithIdenticalEntries(n, 0)` here instead of `[1 .. n] * 0`. It's clearer, and faster:
```
gap> for i in [1 .. 50000] do x := ListWithIdenticalEntries(n, 0); od; time;
78
gap> for i in [1 .. 50000] do x := [1 .. n] * 0; od; time;
347
```
(Typical performance).

gap/semigroups/semicons.gi
james-d-mitchell: Yeah, I'll enter some line breaks in this error.

gap/semigroups/semicons.gi
james-d-mitchell: I reformulated this several times, and I thought that it was clearer what was going on as it is.

gap/semigroups/semicons.gi
james-d-mitchell: Good idea

gap/semigroups/semicons.gi
james-d-mitchell: I didn't know this, I'll use `ListWithIdenticalEntries` instead

gap/semigroups/semicons.gi
james-d-mitchell: I think restricting this to finite semigroups is ok.

gap/semigroups/semicons.gi
wilfwilson: Fine.

doc/semicons.xml
```diff
@@ -196,6 +196,69 @@ true]]></Example>
...
+          <Item><C>IsTropicalMaxPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsTropicalMinPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsProjectiveMaxPlusMatrixSemigroup</C>,</Item>
+          <Item><C>IsIntegerMatrixSemigroup.</C></Item>
+        </List>
+        
+        The optional argument <A>G</A> defaults to a trivial permutation group.
+        If present <A>G</A> must be a permutation group, unless <A>filt</A> is
+        <C>IsReesZeroMatrixSemigroup</C> when <A>G</A> may be any type of
+        group.
```
wilfwilson: group -> finite group, please

PR #412, opened 2017-11-10
Require Orb v4.8.0

wilfwilson:
This is required due to changes in GAP that affected the hashing of permutations, and because Orb v4.8.0 renames `IsClosed` to `IsClosedOrbit`.


wilfwilson:
Oh yeah, oops, obviously there's more to be done to make Semigroups compatible with the new Orb.

PR #415, opened 2017-11-15
Add support for partial orders of L/R-classes

wilfwilson:
We already have `PartialOrderOfDClasses` and `MaximalDClasses`; it would be nice if we had similar for `L` and `R`. It would be useful for my code to do with direct products, in particular.

I've added methods for enumerable semigroups. I feel like it should be simple to do something clever for acting semigroups, but maybe not.

I still need to polish this PR (more manual examples and tests), so it's not ready to be merged, but I'm creating this PR to spur me to finish it, and possibly as a way to ask for advice.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -192,13 +192,17 @@ gap> MinimalDClass(S);
   </ManSection>
 <#/GAPDoc>
 
-<#GAPDoc Label="MaximalDClasses">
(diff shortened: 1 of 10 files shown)


james-d-mitchell:
I've just rebased this onto the most recent `master`, if the CI passes, I think I'll merge it, because it is essentially complete, and has been rotting for too long already (also I think it adds a useful feature!)

PR #417, opened 2017-11-17
Avoid GAP headers other than src/compiled.h

fingolfin:
I'd like to clean up the GAP src directory for GAP 4.10 (see <https://github.com/gap-system/gap/pull/1915>). However, this will break packages which directly include GAP headers other than src/compiled.h -- but luckily there are just two packages doing that, IO and semigroups. Hence this PR.

The name `src/compiled.h` is of course not very "intuitive", so I'd like to change that eventually, too (naming suggestions are welcome :-), but it'll stick around as long as packages use it.

### DIFF
--- a/src/bipart.cc
+++ b/src/bipart.cc
@@ -29,8 +29,7 @@
 #include <vector>
 
 #include "libsemigroups/src/semigroups.h"
-#include "src/permutat.h"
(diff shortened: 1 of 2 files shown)


wilfwilson:
Thanks @fingolfin

PR #422, opened 2017-12-06
kernel: update for libsemigroups v0.6.0

james-d-mitchell:


wilfwilson:
This won't fix the current problem with Travis, unfortunately, but near the top of `.travis.yml`, you should change the line `LIBSEMI=v0.5.2` to the new version.

wilfwilson:
If you drop the `travis: vivid -> xenial` commit, rebase, and make the change I mentioned previously, then it should pass now :)

PR #423, opened 2017-12-07
Use C99 init for StructInitInfo

fingolfin:
This is not against stable-2.8 because the file was renamed compared to there. But if you prefer, I can also make this PR against stable-2.8.

### DIFF
--- a/src/pkg.cc
+++ b/src/pkg.cc
@@ -639,18 +639,11 @@ static Int InitLibrary(StructInitInfo* module) {
 *F  InitInfopl()  . . . . . . . . . . . . . . . . . table of init functions
 */
 static StructInitInfo module = {
-    /* type        = */ MODULE_DYNAMIC,
-    /* name        = */ "semigroups",
-    /* revision_c  = */ 0,
-    /* revision_h  = */ 0,
-    /* version     = */ 0,
-    /* crc         = */ 0,
-    /* initKernel  = */ InitKernel,
-    /* initLibrary = */ InitLibrary,
-    /* checkInit   = */ 0,
-    /* preSave     = */ 0,
-    /* postSave    = */ 0,
-    /* postRestore = */ 0};
+    .type = MODULE_DYNAMIC,
+    .name = "semigroups",
+    .initKernel = InitKernel,
+    .initLibrary = InitLibrary,
+};
 
 extern "C" StructInitInfo* Init__Dynamic(void) {
   return &module;


james-d-mitchell:
@fingolfin thanks for this, if you could rebase against stable-3.0 that'd be great.

fingolfin:
Rebased

james-d-mitchell:
Somehow when you debased the ifdef preprocessor lines you added have disappeared, and now the tests fail. Can you please update ? @fingolfin

fingolfin:
I am not sure which "ifdef preprocessor lines" you are referring to?

Anyway, looking at the failure, it turns out that GCC does not accept C99 designated initializers in C++ code (tested with GCC 5.5, 6.4 and 7.2)-- while I am usually compiling with clang, which accepts it (tested with clang 3.7 and 5.0). Reading some more, it seems only C++20 will (perhaps) add support for it. (It's really annoying that C and C++ are diverging more and more, I must say)

So I guess this PR is not acceptable in this form for you guys :-(. Only way I can see is to split `pkg.cc` and move parts of it in to a plain C file.

james-d-mitchell:
Hmm maybe I'm misremembering but I thought this worked and passed the Ci checks at some point, then you rebased and it didn't work.

 We could move parts of pkg.cc to c if desirable, is it required for something?

fingolfin:
tl;dr: It's not strictly required, you'll probably be fine w/o this PR.

Long version: The motivation for this PR (which was in similar form submit to every package with a kernel extension), is (a) better readability of the struct initialization, and (b) more flexibility for the kernel to change `StructInitInfo` in the future: If all package use designated initializers, we can add or remove fields, reorder them etc., and packages only need to be recompiled, but require no other changes.

Now, of course we don't plan to go wild on this and change `StructInitInfo` like crazy :-). But e.g. the `revision_c` and `revision_h` fields are unused, and could be removed; and we might want to add a few new fields to deal with (de)initialization of per-thread module state in HPC-GAP (which we right now instead do via a separate "module registration API", but it would fit very naturally into `StructInitInfo`).

So, the worst that happens if you don't integrate this PR in some form, is that if we ever really want to make those changes in the GAP kernel, we'll have to coordinate with `semigroups` to make the required change there in parallel. I am pretty sure we can do that, too (though if one wanted to, say, stay compatible with both GAP 4.9 and 4.10, that might require some elaborate #ifdef'ing. Not *that* bad, though).

james-d-mitchell:
Thanks @fingolfin, I'm happy to make changes/releases whenever `StructInitInfo` might change, and if it is happening often, then will split the relevant part out of the `pkg.cc` file.

james-d-mitchell:
@fingolfin can we close this?

Issue #424, opened 2017-12-14
Factorization returns incorrect answers


james-d-mitchell:
I get the following in v3.0.10:

~~~~
gap> S := SymmetricInverseMonoid(4);;
gap> T := IdempotentGeneratedSubsemigroup(S);;
gap> MinimalFactorization(T, GeneratorsOfSemigroup(T)[2]);
[ 5 ]
~~~~

The answer should be `[ 2 ]`, not sure what's going on.

PR #426, opened 2017-12-15
semifp: Add a method for IsomorphismFpSemigroup

ChristopherRussell:
This method applies to factorizable inverse monoids of partial perms. Method is
based on the paper Presentations of Factorizable Inverse Monoids by D. Easdown,
J. East, and D. G. FitzGerald (2004).

### DIFF
--- a/gap/semigroups/semifp.gi
+++ b/gap/semigroups/semifp.gi
@@ -298,7 +298,7 @@ function(S)
     spos := Position(sgens, sgens[i]);
     mpos := Position(mgens, sgens[i], start[spos]);
     lookup[i] := mpos;
-    if spos <> fail then
+    if mpos <> fail then
       start[spos] := mpos;
     fi;
   od;
@@ -542,6 +542,10 @@ function(S, x)
   return SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(x));
 end);
 
+# This method is based on the following paper
+# Presentations of Factorizable Inverse Monoids
+# David Easdown, James East, and D. G. FitzGerald
+# July 19, 2004
 InstallMethod(IsomorphismFpSemigroup,
 "for an inverse partial perm semigroup",
 [IsPartialPermSemigroup and IsInverseActingSemigroupRep],
@@ -555,16 +559,11 @@ function(M)
   fi;
 
   add_to_odd_positions := function(list, s)
-    local i;
-    for i in [1, 3 .. Length(list) - 1] do
-      list[i] := list[i] + s;
-    od;
+    list{[1, 3 .. Length(list) - 1]} := list{[1, 3 .. Length(list) - 1]} + s;
     return list;
   end;
 
-  # FIXME: this shouldn't be necessary but for an error with
-  # MinimalFactorization
-  S := Semigroup(IdempotentGeneratedSubsemigroup(M), rec(acting := false));
+  S := Semigroup(IdempotentGeneratedSubsemigroup(M));
   SS := GeneratorsOfSemigroup(S);
   G := GroupOfUnits(M);
   GG := GeneratorsOfSemigroup(G);
@@ -590,7 +589,7 @@ function(M)
     Add(rels, [ObjByExtRep(fam, lhs), ObjByExtRep(fam, rhs)]);
   od;
 
-  # R_product
+  # R_product - see page 4 of the paper
   for x in [1 .. s] do
     for y in [1 .. g] do
       rhs := Factorization(S, SS[x] ^ (GG[y] ^ -1));
@@ -602,7 +601,7 @@ function(M)
   map := InverseGeneralMapping(IsomorphismPermGroup(G));
   H   := Source(map);
   o   := Enumerate(LambdaOrb(M));
-  #R_tilde
+  #R_tilde - see page 4 of the paper
   for m in [2 .. Length(OrbSCC(o))] do
     comp := OrbSCC(o)[m];
     U := SmallGeneratingSet(Stabilizer(H,
@@ -634,7 +633,7 @@ function(M)
   map := x -> ElementOfFpSemigroup(fam, EvaluateWord(GeneratorsOfSemigroup(F),
                                                      Factorization(T, x)));
   inv := x -> EvaluateWord(GeneratorsOfSemigroup(T),
-         SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(x)));
+                           SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(x)));
 
   return MagmaIsomorphismByFunctionsNC(M, MF, map, inv);
 end);
--- a/gap/semigroups/semifp.gi
+++ b/gap/semigroups/semifp.gi
@@ -541,3 +541,100 @@ IsCollsElms, [IsFpSemigroup, IsElementOfFpSemigroup],
 function(S, x)
   return SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(x));
 end);
+
+InstallMethod(IsomorphismFpSemigroup,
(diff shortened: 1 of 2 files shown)


ChristopherRussell:
I'm not sure if I have done lines 632 - 640 of semifp.gi in the best way yet it was the best way I could think of. Also BruteForceInverseCheck typically takes much longer than BruteForceIsoCheck - does this say that the way I have created the function inv is ineffective?

```
MF := F / rels; # FpSemigroup which is isomorphic to M, with different gens.
fam := ElementsFamily(FamilyObj(MF));
T := Semigroup(Concatenation(SS, GG)); # M with isomorphic generators to MF

map := x -> ElementOfFpSemigroup(fam, EvaluateWord(GeneratorsOfSemigroup(F),
                                                   Factorization(T, x)));
inv := x -> EvaluateWord(GeneratorsOfSemigroup(T),
        SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(x)));

  return MagmaIsomorphismByFunctionsNC(M, MF, map, inv);
end);
```

ChristopherRussell:
@james-d-mitchell I'm not sure why the travis tests are failing.

james-d-mitchell:
Hmm, @ChristopherRussell I'm not so sure, do the test all pass on your computer?

wilfwilson:
@ChristopherRussell if you rebase on the most recent master branch, the diffs should disappear.

ChristopherRussell:
Thanks @wilfwilson!

ChristopherRussell:
I have fixed some of the issues in sempfp.tst (made a fix to IsomorphismFpMonoid and stopped calling BrandtSemigroup which is not yet in stable-3.0). 

@james-d-mitchell I can't reproduce the remaining errors in the failed travis test, which only occur in one of the builds. Also I am am going to write a function `EvaluateExtRep` to save converting ExtReps to words then using `EvaluateWord`. The function would look something like this:

```
InstallMethod(EvaluateExtRep,
"for a partial perm coll and a list of positive ints of even length",
[IsMultiplicativeElementCollection, IsList],
function(gens, w)
  local res, i;
    if Length(w) = 0 then
      return One(gens); # need to account for semigroup not being a moniod?
    elif Length(w) mod 2 = 1 then
      ErrorNoReturn("Semigroups: EvaluateExtRep, the second argument must be ",
                     "a list of even length");
    fi;
    res := gens[AbsInt(w[1])] ^ w[2];
    for i in [3, 5 .. Length(w) - 1] do
      res := res * gens[AbsInt(w[i])] ^ w[i + 1];
    od;
    return res;
end);
```

We also noticed that `EvaluateWord` has several implementations in semigroups/gap/main/orbits.gi and you asked me to remind you that these should perhaps be replaced by a single method with filter `IsMultiplicativeElementColl`.

james-d-mitchell:
@ChristopherRussell Maybe the failure here is related to Issue #435? I also can't reproduce this, on my computer. Can you rebase on the top of stable-3.0 (which now contains some changes not present when travis ran) and see if the tests pass?

james-d-mitchell:
Also I like your suggestion for `EvaluateExtRep`, I'd call it `EvaluateExtRepObjWord` or similar though. I'd handle the case of monoid separately (i.e. in another method), with the filter being `IsMultiplicativeElementWithOneCollection` and in the version for `IsMultplicativeElementCollection` I'd just not allow the case of words of length 0 (error). 

Also I like the suggestion about combining the methods for `EvaluateWord`, I'd be tempted to have two methods again one for `IsMultiplicativeElementCollection` and another for `IsMultiplicativeElementWithOneCollection` they can both use the method for bipartitions.

ChristopherRussell:
I will make the `EvaluateExtRepObjWord` and `EvaluateWord` changes with PR #443

james-d-mitchell:
This looks good. Here are a couple of minor points:

1. You mention a bug in factorisation in a comment, what's this bug can you file and issue? 
2. Can you add an extreme test together with a comment that your method is quicker than the default one in that case? 
3. Can you please reference the paper as a comment before the method?

gap/semigroups/semifp.gi
```diff
@@ -541,3 +541,100 @@ IsCollsElms, [IsFpSemigroup, IsElementOfFpSemigroup],
...
+    TryNextMethod();
+  fi;
+
+  add_to_odd_positions := function(list, s)
+    local i;
+    for i in [1, 3 .. Length(list) - 1] do
+      list[i] := list[i] + s;
+    od;
+    return list;
+  end;
```
james-d-mitchell: Couldn't this whole method just be `list{[1, 3 .. Length(list) - 1]} := list{[1, 3 .. Length(list) - 1]} + s`?

gap/semigroups/semifp.gi
```diff
@@ -541,3 +541,100 @@ IsCollsElms, [IsFpSemigroup, IsElementOfFpSemigroup],
...
+
+  # FIXME: this shouldn't be necessary but for an error with
+  # MinimalFactorization
```
james-d-mitchell: Please file an issue for this error, or is this already resolved?

gap/semigroups/semifp.gi
```diff
@@ -541,3 +541,100 @@ IsCollsElms, [IsFpSemigroup, IsElementOfFpSemigroup],
...
+      rhs := Factorization(S, SS[x] ^ (GG[y] ^ -1));
+      Add(rels, [F.(s + y) * F.(x),
+                 EvaluateWord(GeneratorsOfSemigroup(F), rhs) * F.(s + y)]);
+    od;
+  od;
+
+  map := InverseGeneralMapping(IsomorphismPermGroup(G));
+  H   := Source(map);
+  o   := Enumerate(LambdaOrb(M));
+  #R_tilde
```
james-d-mitchell: Can we have a line/page reference in the paper for this?

gap/semigroups/semifp.gi
```diff
@@ -541,3 +541,100 @@ IsCollsElms, [IsFpSemigroup, IsElementOfFpSemigroup],
...
+                          Factorization(S, One(S)))]);
+
+  MF := F / rels; # FpSemigroup which is isomorphic to M, with different gens.
+  fam := ElementsFamily(FamilyObj(MF));
+  T := Semigroup(Concatenation(SS, GG)); # M with isomorphic generators to MF
+
+  map := x -> ElementOfFpSemigroup(fam, EvaluateWord(GeneratorsOfSemigroup(F),
+                                                     Factorization(T, x)));
+  inv := x -> EvaluateWord(GeneratorsOfSemigroup(T),
+         SEMIGROUPS.ExtRepObjToWord(ExtRepOfObj(x)));
```
james-d-mitchell: This should be indented so that it is aligned with the `(` in `(GeneratorsOf...`

gap/semigroups/semifp.gi
ChristopherRussell: Yes, changed.

gap/semigroups/semifp.gi
ChristopherRussell: I can't remember the issue now. It was something we noticed when writing code together. I am not sure if it was a bug or just that the format of the output didn't suit our purposes (e.g. due to how it dealt with inverses or the identity when factorizing).

My tests still pass when I remove the `rec(acting := false)`. MinimalFactorization uses `EN_SEMI_FACTORIZATION` so

james-d-mitchell: Aha, then this was fixed, it was Issue #424. Please remove the comment.

ChristopherRussell: Ah okay, great.

james-d-mitchell: Please also remove the unnecessary code.

PR #441, opened 2018-01-16
Update tests infrastructure

james-d-mitchell:
This PR updates the test infrastructure for Semigroups to be more like that of Digraphs. Some redundant or unused functions are removed in the process.


james-d-mitchell:
Test failure might be caused by using old version of io:

https://github.com/gap-system/gap/issues/2076

PR #443, opened 2018-01-16
Add EvaluateExtRepObjWord change EvaluateWord

ChristopherRussell:
Added new method `EvaluateExtRepObjWord` to help when dealing with ExtRep objects. Reduced the three methods for `EvaluateWord` to two methods that have less restrictive filters.

EDIT: I still need to add documentation and tests

### DIFF
--- a/gap/main/orbits.gd
+++ b/gap/main/orbits.gd
@@ -8,10 +8,9 @@
 #############################################################################
 ##
 
-DeclareOperation("EvaluateWord", [IsBipartitionCollection, IsList]);
-DeclareOperation("EvaluateWord", [IsPartialPermCollection, IsList]);
-DeclareOperation("EvaluateWord",
-                 [IsReesZeroMatrixSemigroupElementCollection, IsList]);
+DeclareOperation("EvaluateWord", [IsMultiplicativeElementCollection, IsList]);
+DeclareOperation("EvaluateExtRepObjWord",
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
You can leave the PR open when you're adding tests and doc.

james-d-mitchell:
Nice one, looks good to me.

james-d-mitchell:
Some tests would be nice, I'm not completely sure that doc is required, since there wasn't any before :)

james-d-mitchell:
Any comments @wilfwilson or @mtorpey ?

ChristopherRussell:
Regarding documentation, if we have an fp semigroup `F` and an element `x` then

`EvaluateExtRepObjWord(Generators(F), ExtRepOfObj(x)) = x`

will return true. However `ExtRepOfObj` has different implementations for different types of semigroups and sometimes no implementation. Furthermore the implementation for finitely presented semigroups is not documented.

james-d-mitchell:
Ok, I don't really see any of that as a problem, if you want please add doc.

james-d-mitchell:
@ChristopherRussell I'm happy with this, but since I merged your last PR, you'll have to rebase this one onto `stable-3.0` again.

ChristopherRussell:
@james-d-mitchell thanks. I have just rebased now.

PR #455, opened 2018-02-08
Add numbers option to DotString for congruence posets

mtorpey:
Currently, if we create a congruence poset (e.g. by using `LatticeOfCongruences`) and we call `DotString` on it, the numbers of the congruences are displayed in the diagram unless there are more than 40, in which case we just use dots instead, for simplicity.

@ffloresbrito has a few large lattices he wants to explore, and wants to be able to put a number to each dot on the diagram.  Doing this manually obviously takes a long time, and it would be useful for him to have a simple option to override the "less than 40" rule.

This PR introduces a new `DotString` record option, `show_numbers`, which forces the function to include or exclude labelled circles in the output.  Here's an example of it in use:

```gap
gap> W := WreathProduct(Group((1, 2)), FullTransformationMonoid(3));;
gap> latt := LatticeOfCongruences(W);;                         
gap> Splash(DotString(latt, rec(show_numbers := true)));
```

We already have the option called `info`, so I've made it work in the same way; I also slightly modified the way these options are initialised.  Both of these options are hidden (undocumented) for now.


mtorpey:
Changed to `numbers` as you suggested (good call!) and rebased.  I'm leaving it on master rather than stable-3.0 since it's a new feature (even though it's not advertised).

ffloresbrito:
Thank you. I greatly appreciate it.

PR #457, opened 2018-02-14
Smaller generating pair sets for RMS congruences

mtorpey:
If `cong` is a congruence on a finite (0-)simple semigroup described by a linked triple (i.e. if it satisfies `IsR(Z)MSCongruenceByLinkedTriple`) then we might wish to convert it to a congruence defined by generating pairs (e.g. by using `AsSemigroupCongruenceByGeneratingPairs`).  This uses a special method for `GeneratingPairsOfMagmaCongruence`, which is installed in `congrms.gi`.

It turns out that the current method is really bad in terms of the number of pairs it spits out.  It iterates over every element of the normal subgroup `n`, and doesn't combine this in any way with the column and row operations.  This PR replaces it with an all-new, improved algorithm which avoids unnecessary work and condenses the information into as few pairs as possible.  The number of generating pairs produced is drastically reduced, as displayed in the following example:

```
gap> g := SymmetricGroup(4);;
gap> mat := [[(1, 3), (1, 2)(3, 4)],
>            [(1, 4, 3, 2), ()],
>            [(1, 3)(2, 4), (1, 3, 4, 2)]];;
gap> S := ReesMatrixSemigroup(g, mat);;
gap> congs := CongruencesOfSemigroup(S);;
gap> List(congs, c -> Size(GeneratingPairsOfSemigroupCongruence(c)));
```
The output of this example is currently
```[ 1, 4, 6, 12, 14, 14, 16, 14, 15, 17, 17, 19, 17, 24, 26, 26, 28, 26, 27, 29, 29, 31, 29 ]```
but after this PR it is improved to
```[ 1, 2, 2, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2 ]```

The idea behind the algorithm is to create three lists:
* pairs of columns to be identified
* pairs of rows to be identified
* group elements to be put into the normal subgroup `n`

and then to make pairs of RMS elements for the generating set, where each pair of RMS elements encompasses three pieces of information, one from each list of these three lists.  I'm currently TeXing a proper description of the algorithm and a proof of its validity for my thesis (I can show this to anyone who's interested when it's done).  Anyway, I'm confident it is correct, and it certainly passes the suite of tests we have, along with any more I've thrown at it.

Note that currently `congrms` has different for RMS and RZMS congruences.  These methods are different but similar, causing a certain amount of code duplication.  This duplication should be removed eventually, but this PR doesn't make it any worse.

### DIFF
--- a/doc/cong.xml
+++ b/doc/cong.xml
@@ -573,7 +573,7 @@ gap> S := ReesZeroMatrixSemigroup(SymmetricGroup(3),
 gap> cong := CongruencesOfSemigroup(S)[3];;
 gap> AsSemigroupCongruenceByGeneratingPairs(cong);
 <semigroup congruence over <Rees 0-matrix semigroup 2x2 over 
-  Sym( [ 1 .. 3 ] )> with 3 generating pairs>]]></Example>
+  Sym( [ 1 .. 3 ] )> with 1 generating pairs>]]></Example>
     </Description>
   </ManSection>
 <#/GAPDoc>
--- a/gap/congruences/congrms.gi
+++ b/gap/congruences/congrms.gi
@@ -1260,48 +1260,59 @@ InstallMethod(GeneratingPairsOfMagmaCongruence,
 "for Rees matrix semigroup congruence by linked triple",
 [IsRMSCongruenceByLinkedTriple],
 function(cong)
-  local S, g, m, pairs, x, bl, j, rowNo, i, colNo;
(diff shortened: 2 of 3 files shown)


james-d-mitchell:
Looks good in principal, I'd like to see the write up before merging. And I think this can go into `stable-3.0` since it does not introduce any new functionality, only improves existing.

PR #460, opened 2018-02-19
prop: resolve Issue 459 IsTrivial

james-d-mitchell:
This PR resolves Issue 459, as indicated in the issue itself.


james-d-mitchell:
There appear to be some further issues with the GAP tests themselves, I think because some new tests have been added there which have different output when Semigroups is loaded.

PR #474, opened 2018-04-23
Remove IndecomposableElements immediate method

wilfwilson:
I'm not sure why I created an immediate method for `IndecomposableElements`. This PR removes the immediate method, because of https://github.com/gap-system/gap/issues/2377. It doesn't seem to serve an important purpose. I think I just assumed that setting an empty list as an attribute was cheap.


james-d-mitchell:
I think this is good, I'll rerun the travis once PR #475 is merged.

PR #484, opened 2018-05-28
bipart: simplify BipartitionByIntRep

james-d-mitchell:


james-d-mitchell:
@wilfwilson could you take a quick look at the changes made to the tests here? They relate to `IsMonogenicSemigroup` and `MaximalSubsemigroups` (1 line each). Thanks.

tst/standard/maximal.tst
```diff
@@ -601,9 +601,8 @@ gap> MaximalSubsemigroups(U);
   <subsemigroup of 3x4 Rees 0-matrix semigroup with 5 generators> ]
 
 # maximal: MaximalSubsemigroups, for a permutation group, 1
-gap> MaximalSubsemigroups(SymmetricGroup(3));
-[ <group with 1 generator>, <group with 1 generator>, <group with 1 generator>
- , <group with 1 generator> ]
+gap> List(MaximalSubsemigroups(SymmetricGroup(3)), IsGroupAsSemigroup);
```
wilfwilson: Fine, groups just seem to print differently now.

tst/standard/properties.tst
```diff
@@ -941,7 +941,7 @@ gap> IsMonogenicSemigroup(S);
 true
 gap> x := MinimalSemigroupGeneratingSet(S)[1];;
 gap> S := Semigroup(x, x, x);;
-gap> HasIsMonogenicSemigroup(S) and IsMonogenicSemigroup(S);
+gap> IsMonogenicSemigroup(S);
```
wilfwilson: This isn't going to cause any problems, but I can't work out why the immediate method on `gap/attributes/properties.gi:771` no longer sets `IsMonogenicSemigroup` to be true for `S`. (Whether that immediate method is a good idea is a separate issue).

james-d-mitchell: Me neither, but as you say it is unlikely to cause any problems. Thanks!

PR #494, opened 2018-06-08
Fix building Semigroups on windows

ChrisJefferson:
This (along with a patch to libsemigroups) fixes semigroups in cygwin


james-d-mitchell:
@ChrisJefferson The tests fail because of some changes in GAP `master`, I've pushed a fix to this in PR #495, and I'll merge this as soon as the tests are back to normal. Thanks for this!

PR #498, opened 2018-06-15
Fix issue 496

james-d-mitchell:


src/bipart.cc
```diff
@@ -1082,7 +1093,7 @@ Obj BLOCKS_INV_LEFT(Obj self, Obj blocks_gap, Obj x_gap) {
   for (u_int32_t i = 0; i < blocks->degree(); i++) {
     (*out_blocks)[i] = blocks->block(i);
     u_int32_t j      = fuse_it(x->at(i) + blocks->nr_blocks());
-    if (j > blocks->nr_blocks() || tab[j] == (size_t) -1) {
+    if (j >= blocks->nr_blocks() || tab[j] == (size_t) -1) {
```
flsmith: Is there a reason you've used `static_cast` above in the asserts but left this C-style? I think it would be nice to be consistent (although there are a lot more C-style ones...).

james-d-mitchell: Good point, I'll update this to remove the c-style casts.

PR #500, opened 2018-06-19
Add method to find all the ideals of a semigroup

mtorpey:
This pull request adds a new attribute for a semigroup, called `Ideals`.  An attribute with the same name already exists for rings, so I chose this name for consistency.  I've only added a method for _finite_ semigroups – I'm not sure how we would approach infinite semigroups.

An ideal is a downward closed union of D-classes.  So to find all the ideals, we first find all the D-classes, then find every possible set of pairwise incomparable D-classes.  For each set, we take an element from each class, and use these as generators for an ideal (they generate all the D-classes that lie below them).  In this way, we use the minimum possible number of generators for each ideal, since we only use one generator from each of the ideal's maximal D-classes.

It turns out the Digraphs package is quite useful for implementing this, so we use several Digraphs methods, including `DigraphCliques`.

Doc and tests are included.  Since this is a new feature, I've put it on `master`.

### DIFF
--- a/doc/ideals.xml
+++ b/doc/ideals.xml
@@ -152,3 +152,38 @@ false]]></Example>
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="Ideals">
+  <ManSection>
+    <Attr Name = "Ideals" Arg = "S" Label = "for a semigroup"/>
+    <Returns>
+      An list of ideals.
+    </Returns>
+    <Description>
+      If <A>S</A> is a finite non-empty semigroup, then this attribute returns a
+      list of all the ideals of <A>S</A>. <P/>
+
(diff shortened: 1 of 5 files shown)


doc/ideals.xml
```diff
@@ -152,3 +152,38 @@ false]]></Example>
...
+
+<#GAPDoc Label="Ideals">
+  <ManSection>
+    <Attr Name = "Ideals" Arg = "S" Label = "for a semigroup"/>
+    <Returns>
+      An list of ideals.
+    </Returns>
+    <Description>
+      If <A>S</A> is a semigroup, then this attribute returns a list of all the
+      ideals of <A>S</A>. <P/>
```
james-d-mitchell: Maybe add "non-empty" here!

mtorpey:
Modified as suggested.  I also mentioned "finite".

PR #501, opened 2018-07-09
Use MarkBag instead of MARK_BAG

fingolfin:
In GAP 4.9, the latter is an alias for backwards compatibility for the former.

Alternatively, one could replace `TBipartObjMarkSubBags` by `MarkTwoSubBags` if one was willing to reorder the content of `T_BIPART` blocks so that the two subobject pointers come first, and the C++ pointer last (instead of first).


james-d-mitchell:
Thanks @fingolfin, the second solution would probably be better, but is more work! I'll do it on the next refactoring of the kernel code.

PR #502, opened 2018-08-28
Two C code tweaks: don't use InfoBags; use AlwaysNo for IsMutableObjFuncs

fingolfin:
I based this PR on stable-3.0, let me know if I should do something different.


james-d-mitchell:
This looks good to me, the test fail because the changes make one line in `pkg.cc` longer than 80 characters, and because it would appear that our travis setup is broken. Could you please fix the line length issue in `pkg.cc` @fingolfin. Thanks!

fingolfin:
Done

wilfwilson:
Thanks @fingolfin

PR #505, opened 2018-08-29
Use full names instead of initials in PackageInfo.g

### LINKED ISSUE #504: Use full names in PackageInfo.g
Would it be possible to use the first full name in `FirstNames` component in PackageInfo.g file, like the majority of packages doing? GAP website content generating scripts will still be able to convert them to initials in some places, but will display it in full in a package overview page?

This request actually comes not from the website maintenance, but from https://github.com/gap-system/gap-distribution/issues/49. Extracting names from PackageInfo.g files,  I am trying to get a list of all package authors without duplicates, and using different forms for the same person makes it more difficult. For example, we have both

 - M. Delgado, Manuel Delgado, 
 - C. Jefferson, Christopher Jefferson, 
 - J. D. Mitchell, James Mitchell, 
 - M. Pfeiffer, Markus Pfeiffer, 
 - C. Russell, Christopher Russell,

alex-konovalov:
Closes #504

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -68,7 +68,7 @@ ArchiveURL      := Concatenation(~.SourceRepository.URL,
 Persons := [
   rec(
     LastName      := "Mitchell",
-    FirstNames    := "J. D.",
+    FirstNames    := "James",
     IsAuthor      := true,
     IsMaintainer  := true,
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Thanks Alex

PR #507, opened 2018-09-05
Make McAlister triple semigroups acting

ChristopherRussell:
This PR integrates McAlister triple semigroups with the acting semigroups framework. It also includes various fixes, and improvements to:

- IsomorphismSemigroup when used to find an isomorphism from an E-unitary inverse semigroup to a McAlisterTripleSemigroup.
- McAlisterTripleSemigroup - replaced a check involving IsSubgroup and AutomorphismGroup by one which checks if generators are automorphisms.
- EUnitaryInverse cover - now uses DirectProduct functionality which has recently been added to Semigroups.

It also adds various attributes of the form MTSX which are synonyms for McAlsiterTripleSemigroupX.  FInally there is now a McAlisterTripleSubsemigroup attribute and functionality for these objects.

### DIFF
--- a/gap/main/setup.gi
+++ b/gap/main/setup.gi
@@ -51,7 +51,8 @@ InstallMethod(IsGeneratorsOfActingSemigroup,
 "for a McAlister triple element collection",
 [IsMcAlisterTripleSemigroupElementCollection],
 function(coll)
-  return IsPermGroup(McAlisterTripleSemigroupGroup(MTSEParent(coll[1])));
+  return
+  IsPermGroup(McAlisterTripleSemigroupGroup(MTSEParent(Representative(coll))));
     # and McAlisterTripleSemigroupAction(MTSEParent(coll[1])) = OnPoints;
 end);
 
@@ -102,11 +103,10 @@ function(x)
   return NrMovedPoints(x![2]) + 1;
 end);
 
-# TODO correct?
 InstallMethod(ActionDegree, "for a McAlister semigroup element",
 [IsMcAlisterTripleSemigroupElement],
 function(x)
-  return NrMovedPoints(x[2]);
+  return 0;
 end);
 
 InstallMethod(ActionDegree, "for a matrix over finite field object",
@@ -169,14 +169,10 @@ function(R)
   return 0;
 end);
 
-# TODO check this is ok?
-
 InstallMethod(ActionDegree, "for a McAlister triple subsemigroup",
 [IsMcAlisterTripleSubsemigroup],
 function(S)
-  local rep;
-  rep := Representative(S);
-  return NrMovedPoints(McAlisterTripleSemigroupGroup(MTSEParent(rep)));
+  return 0;
 end);
 
 InstallMethod(ActionDegree, "for a matrix over finite field semigroup",
@@ -248,8 +244,6 @@ function(R)
   end;
 end);
 
-# TODO: ok?
-
 InstallMethod(ActionRank, "for a McAlister triple semigroup element and int",
 [IsMcAlisterTripleSemigroupElement, IsInt],
 function(f, n)
@@ -259,8 +253,6 @@ function(f, n)
   id      := McAlisterTripleSemigroupComponents(MTSEParent(f)).id;
   return Position(DigraphTopologicalSort(digraph), id[f[1]]);
 end);
-
-# TODO: ok?
 
 InstallMethod(ActionRank, "for a McAlister triple subsemigroup",
 [IsMcAlisterTripleSubsemigroup],
@@ -784,11 +776,12 @@ end);
 InstallMethod(RhoBound, "for a Rees 0-matrix semigroup",
 [IsReesZeroMatrixSubsemigroup], LambdaBound);
 
-# TODO improve the upper bound here
 InstallMethod(LambdaBound, "for a McAlister subsemigroup",
 [IsMcAlisterTripleSubsemigroup], S ->
 function(r)
-  return Size(McAlisterTripleSemigroupGroup(MTSEParent(Representative(S))));
+  local G;
+  G := McAlisterTripleSemigroupGroup(MTSEParent(Representative(S)));
+  return Size(Stabilizer(G, r));
 end);
 
 InstallMethod(RhoBound, "for a McAlister subsemigroup",
--- a/gap/main/setup.gi
+++ b/gap/main/setup.gi
@@ -47,6 +47,14 @@ function(coll)
   return IsPermGroup(UnderlyingSemigroup(R)) and IsRegularSemigroup(R);
 end);
 
+InstallMethod(IsGeneratorsOfActingSemigroup,
+"for a McAlister triple element collection",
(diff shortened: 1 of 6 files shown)


wilfwilson:
Travis seems to be failing nowing because of some manual examples displaying differently, and because of the code coverage of `setup.gi` being only 80%.

james-d-mitchell:
@ChristopherRussell can you attempt to fix this?

ChristopherRussell:
I don’t understand why the manual examples are failing. They fail when I run SEMIFROUPS.TestManualExamples() however they work when I run through them manually in gap. I also thought I covered all the code relating to McAlister triples in setup.gi except a 3-4 lines inside the function IdempotentCreator. I am not sure what this function is meant to do and I don’t understand the code so I’m wondering if it’s wrong.

ChristopherRussell:
Oh okay, I will be able to get the setup.gi test coverage up then.

wilfwilson:
Like you, on my computer I get the same diffs in the manual examples as Travis does, when running `SEMIGROUPS.TestManualExamples();`. There is probably therefore some randomness involved in your methods, and since GAP is in a different state when you run them separately as opposed to when you run everything together as part of `SEMIGROUPS.TestManualExamples();`, then you're getting different results.

I suggest that either you simply use the output that you get from `SEMIGROUPS.TestManualExamples();`, which would fix the problem, or perhaps more appropriate would be to suppress the output or even remove the two steps that fail; since they're not guaranteed to give consistent answers then, perhaps it could be slightly misleading or confusing to a user if they get different results.

Or you could redesign the manual example somehow so that you get consistent results.

ChristopherRussell:
So I deleted the manual examples because they didn't seem important. I have now increased the test coverage for setup.gi and I'm no longer worried about the function `IdempotentCreator`. However I'm having trouble testing `SchutzGpMembership`.

```
gap> M := McAlisterTripleSemigroup(SymmetricGroup([2, 3, 4]), Digraph([[1], [1, 2], [1, 3], [1, 4]]), [1, 2, 3]);;
gap> M := Semigroup(Elements(M));;
gap> o := LambdaOrb(M);; Enumerate(o);;
gap> schutz := LambdaOrbStabChain(o, 2);;
gap> SchutzGpMembership(M)(schutz, ());
Error, IsBound: <rec> must be a record (not a boolean or fail) in
  while IsBound( S.stabilizer ) and g <> S.identity do
    bpt := S.orbit[1];
    img := bpt ^ g;
    if IsBound( S.transversal[img] ) then
        while img <> bpt do
            g := g * S.transversal[img];
            img := bpt ^ g;
        od;
        S := S.stabilizer;
    else
        return g;
    fi;
od; at /Users/crussell/gap/lib/stbc.gi:1350 called from
SiftedPermutation( stab, x
 ) at /Users/crussell/gap/pkg/semigroups/gap/main/setup.gi:1132 called from
<function "unknown">( <arguments> )
 called from read-eval loop at *stdin*:8
you can replace <rec> via 'return <rec>;'
```

This seems to be because `schutz` is equal to `true` and not a stabiliser chain record. I noticed that the test for RZMS, which I was trying to base mine on, should also fail in this way except there is a typo. It is:

```
# SchutzGpMembership, for an RZMS
gap> R := ReesZeroMatrixSemigroup(Group((1, 2, 3)), [[()]]);;
gap> R := Semigroup(Elements(R));;
gap> o := LambdaOrb(S);; Enumerate(o);;
gap> schutz := LambdaOrbStabChain(o, 2);;
gap> SchutzGpMembership(R)(schutz, ());
true
```

and the third line has an `S` where there should be an `R`.

wilfwilson:
Hmm, well done sorting these things out Chris, I think you should alert @james-d-mitchell to the problems you're having with `SchutzGpMembership`.

james-d-mitchell:
@ChristopherRussell `LambdaOrbSchutzGp` returns a group except if that group would be the symmetric group or the trivial group. So, when you are seeing the return value of `LambdaOrbSchutzGp` being `true` or `false` this indicates that the corresponding group is the symmetric group or the trivial group, respectively. 

I think to do the test you want, you just need to find an example where this group is not the symmetric group nor the trivial group. Please let me know if this does/doesn't make sense.

ChristopherRussell:
I believe I have fixed this now. I also fixed the `SchutzGpMembership` test for Rees zero matrix semigroups which I mentioned had a typo.

gap/main/setup.gi
```diff
@@ -94,6 +103,13 @@ function(x)
   return NrMovedPoints(x![2]) + 1;
 end);
 
+# TODO correct?
```
wilfwilson: I've no idea if this is correct, but I feel like the TODO should be resolved before merging.

ChristopherRussell:
Thanks @wilfwilson, I have now resolved the TODOs in `setup.gi`.

ChristopherRussell:
There are some recursion depth traps appearing when my tests are run in examples where `Print` and `String` are called on McAlister triple semigroups when the GAP version is master. Did something change about how strings/printing is being handled recently in master?

wilfwilson:
No, however something has changed which means that the ranks of certain filters are different than they were. That explains all but one of your test failures: for a `McAlisterTripleSemigroup`, you are expecting to use the `String` method installed on `semieunit.gi:218`. But the filter of this is `IsMcAlisterTripleSubsemigroup`, which only has rank `25` in my GAP installation; whereas there is a method in the GAP library which applies to semigroups that are known inverse semigroups and which have known `GeneratorsOfSemigroup`: this has rank `28`, and is getting called instead, and is leading to an infinite recursion. You can fix this by artificially increasing the rank of your special method.

So, after line 219, you could add the line:
```
RankFilter(IsInverseSemigroup and HasGeneratorsOfSemigroup),
```
And this should fix the problem.

The other test failure I think is that an isomorphism is now printed a tiny bit correctly. You should just suppress this output with double semicolons, but I recommend that you actually verify that the mapping is a homomorphism (i.e. a bijective homomorphism).

ChristopherRussell:
Thank you @wilfwilson, I don't think I would have been able to figure out the filter ranks stuff by myself! I have also suppressed the output for the lines in the tests which printed isomorphism functions and added `BruteForceIsoCheck` and `BruteForceInverseCheck` to verify that they are isomorphisms.
Tests seem to be passing now.

gap/main/setup.gi
```diff
@@ -763,7 +929,20 @@ InstallMethod(LambdaConjugator, "for a bipartition semigroup",
 InstallMethod(LambdaConjugator, "for a Rees 0-matrix subsemigroup",
 [IsReesZeroMatrixSubsemigroup], S ->
 function(x, y)
-  return ();
+  return ();  # FIXME is this right???? This is not right!!
```
james-d-mitchell: This FIXME troubles me a bit, I will play around with this on Monday, and give you some feedback @ChristopherRussell. This PR looks pretty good though, thanks for persevering.

james-d-mitchell:
Thanks @ChristopherRussell I'm going to merge this now.

Issue #508, opened 2018-09-06
`\=` for Semigroup and an Ideal crashes if given infinite inputs


markuspf:
```
gap> M := Magma(0,1,2);;
gap> SetIsFinite(M, false);
gap> J := MagmaIdealByGenerators(M,[2]);;
gap> SetIsFinite(J, false);
gap> M = J;
```

makes GAP run out of memory, because the method executed (`=` for an ideal with generators and a semigroup with generators https://github.com/gap-packages/Semigroups/blob/f431ea45eeefcf7dd5495b18084f3cef3b6281d7/gap/ideals/ideals.gi#L164) tries to enumerate the ideal for a membership test.

Either `Semigroups` assumes these objects are finite (in which case the filters should reflect this, but of course could still run out of memory...), or there is an intrinsic reason why there are always membership tests for these objects in semigroups (in which case i would like to know what they are).

I should note that the test in gap (`tst/testinstall/domain.tst`) that exposes this behaviour is a bit of a hack to create domains that are infinite and are supposed to be caught to be incomparable.

james-d-mitchell:
Thanks for the report. It looks to me that probably the filters for the `\=` method for an ideal and generators are perhaps too permissive. 

On the other hand, how does GAP know that they are or are not equal @markuspf ?

wilfwilson:
In GAP, without Semigroups, you get:
```
gap> M := Magma(0,1,2);;
gap> SetIsFinite(M, false);
gap> J := MagmaIdealByGenerators(M,[2]);;
gap> SetIsFinite(J, false);
gap> M = J;
Error, no method found for comparing two infinite domains
```

wilfwilson:
As the Travis tests for #510 show, this is now the only thing stopping our Travis tests from passing in the master branch.

fingolfin:
AFAIK this is a blocker for GAP 4.10. So if it helps, we could simply disable this test in GAP, at least for now (of course it'd still be nice if for semigroups which are known to be infinite, tests which may not terminate were linited... but at least from my POV that's not particularly important; so let's be pragmatic

alex-konovalov:
I am for this. Then this plus https://github.com/gap-system/gap/pull/2796 plus new release of Semigroups will  make us ready for branching off stable-4.10.

alex-konovalov:
@wilfwilson will you be able to submit another PR to disable this test in GAP then, please?

wilfwilson:
[...]

PR #509, opened 2018-09-06
Add TraceMat method for finite field matrices

markuspf:


james-d-mitchell:
Can we have at least one test please @markuspf, for code coverage if nothing else?

james-d-mitchell:
What say you @wilfwilson ?

james-d-mitchell:
@markuspf can you also please rebase onto stable-3.0? Thanks!

wilfwilson:
I second your comments James, it would indeed be nice to have at least one test, and moreover thanks for doing this Markus.

markuspf:
Not entirely sure why github got its knickers in a twist over the second commit there?

markuspf:
I rebased on `stable-3.0` and added a test. I briefly thought about putting the `TraceMat` method in for semiring matrices in general, but then saw that that was more work than I thought...

wilfwilson:
Assuming that James can fix #508, then yes ;-)

Just want to reiterate that we really appreciate your help Markus. We try to be good `gap-system` citizens but you definitely make that easier.

james-d-mitchell:
Of course, I *can* fix it, but will I???

wilfwilson:
We can always just remove the ideals functionality from the Semigroups package ;-)

markuspf:
For reference: https://github.com/gap-system/gap/pull/2784

PR #510, opened 2018-09-06
Reinstate testing again GAP master branch in Travis

wilfwilson:
I've made this PR so that we can (hopefully) easily see what is causing the Travis tests to fail when we use the GAP master branch. I will aim to keep rebasing PR as and when things are changed in the package; when the tests pass we can merge this!

I've also updated the 'description matrix' of what tests we do; it was out of date.

I've made the PR against `stable-3.0`, because for us at the moment I think that's the most important branch: that's where we're going to be making the next releases from, including (I presume) a release that makes Semigroups work completely with GAP 4.10.


wilfwilson:
@james-d-mitchell Would you be happy for this to be merged now? (For me, this page is reporting that the Travis build is still pending, but if you click the 'details' button then you see that the Travis tests have passed successfully).

In particular, if we're going to do anything large-scale like renaming `Matrix`, then I think it would be a good idea to be including tests against the `master` branch of GAP.

alex-konovalov:
@wilfwilson build in progress = https://travis-ci.org/gap-packages/Semigroups/builds/426788950

PR #515, opened 2018-09-13
Install libsemigroups in Travis with a separate script

wilfwilson:
A compatible version of libsemigroups in the right location is required for the GitHub version of Semigroups to compiled and used.  In our Travis setup, there is a stage where we check the file `.LIBSEMIGROUPS_VERSION`, and clone the corresponding release of libsemigroups into the directory `src/libsemigroups`.

I have moved this step in our Travis setup into a separate file called `prerequisites.sh`. This will make it easier for someone to quickly get a working version of a development version of Semigroups. In particular, @alex-konovalov has a system for testing the master versions of packages against GAP, in some way. This currently fails for the Semigroups package because it doesn't know about the requirement to get libsemigroups, and so it doesn't compile.

Since I've already written code which does this, I thought (as suggested by @ChrisJefferson I think) that we should provide a script that can be called before `./autogen.sh && ./configure && make`. With this PR, checking out any fresh version of Semigroups and running `./prerequisites.sh` and then the previous commands should produce a working installation of Semigroups.

I'm not sure if there are other packages that require some additional setup step like ours, but maybe there will be in the future. @alex-konovalov's testing setup could perhaps search for (and execute) `prerequisite.sh` files before looking for `autogen.sh`, `configure`, etc.

I'm open to suggestions for a name other than `prerequisites.sh`.


alex-konovalov:
That will be nice - if there would be a standard `./prerequisites.sh` file, I can achieve https://github.com/gap-packages/gap-docker-pkg-tests-master-devel/issues/7 by checking for its presence and run it, if it is present - there may be some other packages with similar needs.

prerequisites.sh
```diff
@@ -0,0 +1,20 @@
+set -e
+
+if [ -d src/libsemigroups ]; then
+  echo -e "a version of libsemigroups is already present; no action required"
```
james-d-mitchell: Wouldn't it be safer to check if the correct version of Libsemigroups is installed? This is what happens in the configuration step for Semigroups. This is done by `tr -d '\n' < src/libsemigroups/.VERSION` (after checking that the `.VERSION` file exists.

alex-konovalov:
@james-d-mitchell  although not all packages have `autogen.sh` ...

wilfwilson:
Good suggestion about checking the version.

I’ll stick with the full version of the name for now, although I agree it’s not brilliant. If other packages seem to settle on some other name we can adopt that.

wilfwilson:
Sorted!

alex-konovalov:
Good! When this will appear in the master branch of Semigroups?

wilfwilson:
@alex-konovalov This is now in the `master` branch of Semigroups.

alex-konovalov:
@wilfwilson excellent, that works: https://travis-ci.org/gap-packages/gap-docker-pkg-tests-master-devel/jobs/428393254

```
/bin/mkdir -p ./bin/x86_64-pc-linux-gnu-default64 ./bin/lib
cp -RL src/libsemigroups/.libs/* ./bin/lib/
cp .libs/semigroups.so /home/gap/Semigroups/bin/x86_64-pc-linux-gnu-default64/semigroups.so
SUCCESS!
make[1]: Leaving directory '/home/gap/Semigroups'
#I  method installed for Matrix matches more than one declaration
Semigroups package: testinstall.tst
msecs: 2720
#I  No errors detected while testing package semigroups version 3.0.18
#I  using the test file `/home/gap/Semigroups/gaproot/pkg/Semigroups/tst/testi\
nstall.tst'
```

Are you happy with only testinstall.tst being used as a default test for Semigroups? Adding several ore minutes of testing wouldn't be a problem.

wilfwilson:
That's good to hear @alex-konovalov, thanks for accommodating this so quickly.

For Semigroups, I think the most appropriate tests to run would be those contained in `tst/standard`. There are two ways to do this:

1. (Not the best way): find the files in that directory and run each one (this takes about a minute on my computer).
2. (Better): call the Semigroups package function `SemigroupsTestStandard();`, which runs all of the standard test files in one mode, and then in another mode (so about two minutes total on my computer; presumably longer on Travis). We have two fundamentally different ways of computing with finite semigroups, and `SemigroupsTestStandard` tests both of them.

These files give almost complete code coverage for the Semigroups package, and so hopefully any problems with compatibility with GAP will be highlighted quickly.

james-d-mitchell:
I second this, way 2 is more thorough.

Issue #519, opened 2018-09-17
Incorrect (?) use of InstallTrueMethod


fingolfin:
Semigroups sets the following implication. which seem wrong to me:
```gap
InstallTrueMethod(IsCommutativeSemigroup, IsCommutative);
```
That implies that every commutative magma is a semigroup, which I think is false.

Perhaps this was meant?
```gap
InstallTrueMethod(IsCommutativeSemigroup, IsCommutative and IsSemigroup);
```

wilfwilson:
Hi @fingolfin, thanks for this report. It is true that the meaning that we intend is the second one that you list: `IsCommutativeSemigroup` should be `true` for something that is a semigroup and satisfies `IsCommutative`.

This is technically currently how it works, since `IsCommutativeSemigroup` is declared in `lib/semigrp.gd` as:
```
DeclareProperty("IsCommutativeSemigroup", IsSemigroup);
```
Therefore the `TrueMethod` only applies to things that satisfy `IsSemigroup`. However, if this declaration is ever generalised, which it could conceivably be, then this would lead to problems. So I'll add the `IsSemigroup` to be safe 🙂

fingolfin:
Unfortunately that analysis is incorrect, implications do not (and can not, on a technical level) take into account whether a property was only declared for a subset of all filters. Indeed, create a commutative magma that is not a semigroup, and you'll see this implication was applied to it

wilfwilson:
Ah right, thanks for the correction. Maybe this will discouraging me from making assumptions (and encourage me to verify my statements)!

wilfwilson:
Resolved by #520.

PR #520, opened 2018-09-17
Require IsSemigroup in the filters for the IsCommutative TrueMethod

wilfwilson:
It make sense to add `IsSemigroup` to `InstallTrueMethod(IsCommutativeSemigroup, IsCommutative);` to be safe; see #519.


wilfwilson:
I am going to merge this; the tests against the GAP master branch fail because of #522; everything else is fine.

PR #524, opened 2018-09-17
Accommodate changes in GAP affecting method ranks

### LINKED ISSUE #522: Test failures due to method re-ordering
See https://github.com/gap-system/gap/issues/2818. Basically, a recent fix made to the GAP master branch means that the tests of the Semigroups package fail, because some ranks of filters are now different.

I'll open this issue so that people know that we're aware of it. Hopefully this won't be too painful to resolve! 🙂

wilfwilson:
Some changes in GAP `master` branch mean that the ranks of certain filters have changed (see https://github.com/gap-system/gap/issues/2818), which means that some rank adjustments related to the Semigroups package are not correct. To ensure compatibility with all versions of GAP I've made some changes.

If I have done everything correctly, then the only failure in Travis will be GAP's `tst/testinstall/triviso.tst`; this should be fixed if https://github.com/gap-system/gap/pull/2819 is merged into GAP's master branch.

This will therefore basically resolve #522.

### DIFF
--- a/gap/ideals/ideals.gi
+++ b/gap/ideals/ideals.gi
@@ -82,7 +82,7 @@ end);
 InstallMethod(ViewString,
 "for a semigroup ideal with ideal generators",
 [IsSemigroupIdeal and HasGeneratorsOfSemigroupIdeal],
-1,  # to beat the library method
+4,  # to beat the library method
 _ViewStringForSemigroupsIdeals);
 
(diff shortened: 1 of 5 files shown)


wilfwilson:
https://github.com/gap-system/gap/pull/2819 is merged and so this is ready to go

PR #529, opened 2018-09-20
Create semilattice from semilattice digraph

ChristopherRussell:
This PR implements a function `SemilatticeFromDigraph` which takes a digraph satisfying either `IsJoinSemilatticeDigraph` (or `IsMeetSemilatticeDigraph`) and returns a semigroup of partial perms isomorphic to this digraph with the join (or meet) operation on the vertices.

I have a few question marks over what I have done:
- There might be a better way of choosing a generating set (or a better implementation of the way I have chosen one) but I couldn't see any online or think of one.
- There is only one function for both join and meet semilattice digraphs. The function is written for join semilattices but if the input is a meet semilattice AND not a join semilattice then the digraph is reversed then the function is called again. This could be confusing in the case a user specifies a lattice intending for it to be interpreted as a meet semilattice. Should I write different functions for join and meet semilattices or just be clear in the documentation about the behaviour that I have described?

One thing that seems good about the implementation is that when I call `SmallerDegreePartialPermRepresentation` on examples it has always returned the identity mapping. I created some 'random' large examples of semilattices with the following function when testing if anyone is interested in playing around with it:

```
RandomSemilattice := function(n, k)
  local digraph, V;
  digraph := Digraph(Combinations([1 .. n]), {i, j} -> IsSubset(i, j));
  V := Union(Set([1 .. k],
             i -> OutNeighboursOfVertex(digraph, Random([1 .. 2 ^ n]))));
  return InducedSubdigraph(digraph, V); 
end;  
```


wilfwilson:
Sorry @ChristopherRussell, I've currently broken things in GAP so Semigroups doesn't load with the GAP master branch: this will cause your tests to fail at the moment. I'll re-run any ones that fail once GAP is fixed.

doc/semicons.xml
```diff
@@ -299,3 +299,32 @@ true]]></Example>
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="SemilatticeFromDigraph">
+  <ManSection>
+    <Func Name="SemilatticeFromDigraph" Arg="digraph"/>
+    <Returns>
+      A semilattice of partial perms <A>n</A>.
```
wilfwilson: What is the point of the `<A>n</A>` here?

doc/semicons.xml
```diff
@@ -299,3 +299,32 @@ true]]></Example>
...
+      If <A>digraph</A> is a
+      <Ref Prop="IsJoinSemilatticeDigraph" BookName="Digraphs"/> or a
+      <Ref Prop="IsMeetSemilatticeDigraph" BookName="Digraphs"/>, then this
+      function returns a semigroup of partial perms which is isomorphic to
+      the semigroup whose elements are the vertices of <A>digraph</A> with
+      the binary operation
+      <Ref Prop="PartialOrderDigraphJoinOfVertices" BookName="Digraphs"/>
+      or <Ref Prop="PartialOrderDigraphJoinOfVertices" BookName="Digraphs"/>,
+      respectively.
+      <Example><![CDATA[
```
wilfwilson: I definitely think you should have separate functions `SemilatticeFromJoinSemilatticeDigraph` and `SemilatticeFromMeetSemilatticeDigraph`, to avoid the confusion to give the user more control.

gap/semigroups/semicons.gi
```diff
@@ -785,6 +785,58 @@ function(filter, G, n)
...
+    # When there is only one neighbour, we must add a point to the image to
+    # distinguish the two idempotent partial perms.
+    if Length(OutNeighboursOfVertex(red, top[i])) = 1 then
+      Add(im[top[i]], max);
+      max := max + 1;
+    fi;
+  od;
+
+  # Add the generators one by one
+  S := Semigroup(PartialPerm(im[top[n]], im[top[n]]));
```
wilfwilson: You should be able to exactly work out a minimal generating set a priori. Specifically, the generators should be the sources of the `DigraphReflexiveTransitiveReduction`, right? Something like that.

james-d-mitchell:
I haven't look at this in detail @ChristopherRussell but I think that the functions you wrote should be methods for `AsSemigroup` rather than introducing a new function `SemilatticeFromDigraph`.

james-d-mitchell:
Otherwise I think this is a good idea!

wilfwilson:
But in that case I'm not sure the best way of resolving this for: should `AsSemigroup(IsWhateverKind, gr);` for a lattice return the semilattice defined by the meet semilattice or the join semilattice?

gap/semigroups/semicons.gi
ChristopherRussell: A line digraph will have only one source but every element is in a minimal generating set. E.g.
`Digraph([[1], [1, 2], [1, 2, 3]])`. I think you are right that it is possible to work out a generating set without using `ClosureSemigroup` to do the checking though.

wilfwilson: Good thinking! Yeah the correct property is something like 'join-irreducibility' or 'meet-irreducibility' or something. If you can't be bothered working it out then no worries, it can always be improved later, maybe leave a comment to say so.

doc/semicons.xml
```diff
@@ -299,3 +299,32 @@ true]]></Example>
...
+    </Returns>
+    <Description>
+      If <A>digraph</A> is a
+      <Ref Prop="IsJoinSemilatticeDigraph" BookName="Digraphs"/> or a
+      <Ref Prop="IsMeetSemilatticeDigraph" BookName="Digraphs"/>, then this
+      function returns a semigroup of partial perms which is isomorphic to
+      the semigroup whose elements are the vertices of <A>digraph</A> with
+      the binary operation
+      <Ref Prop="PartialOrderDigraphJoinOfVertices" BookName="Digraphs"/>
+      or <Ref Prop="PartialOrderDigraphJoinOfVertices" BookName="Digraphs"/>,
```
james-d-mitchell: Also this says `PartialOrderDigraphJoinOfVertices` or `PartialOrderDigraphJoinOfVertices`, the options are identical.

doc/semicons.xml
ChristopherRussell: Ooops, I have no idea. Probably something I forgot to deleted after copying another manual entry... I will get rid of it.

doc/semicons.xml
ChristopherRussell: Oops, I will change this.

ChristopherRussell:
I have two further questions. 
-If I were to implement this as a method for AsSemigroup does that mean there will be no specific documentation for this method? 
-I was not really sure about my choice of putting this method in `semicons.gi` and I am not sure where it should actually go.

james-d-mitchell:
For the first question, why not add documentation for it, for the specific use case? 

I'm not sure about the second question, what do you think @wilfwilson?

james-d-mitchell:
Also it occurs to me, do we have an `AsDigraph` method for semigroups? The inverse of this method? If not, that'd be good to have too, but not necessarily in this PR.

ChristopherRussell:
It's not quite an inverse to this method but if `digraph` is a join semilattice digraph then `IsIsomorphicDigraph(Digraph(NaturalPartialOrder(SemilatticeFromDigraph(digraph))), digraph)` will return a true.

wilfwilson:
Radical suggestion: but perhaps this could even go in the Digraphs package? I think there's no reason why this couldn't work without the Semigroups package.

james-d-mitchell:
Good suggestion @wilfwilson I wondered about this too! You're right of course, because partial perms are defined in the GAP library.

ChristopherRussell:
Shall I open a PR for Digraphs and instead have two functions called `SemilatticeFromJoinSemilatticeDigraph` and `SemilatticeFromMeetSemilatticeDigraph`? (I'm open to suggestions about the names.)

james-d-mitchell:
They should both be methods for `AsSemigroup` as I said previously.

ChristopherRussell:
I have remade this as a PR for the digraphs package

james-d-mitchell:
Thanks Chris, can we close this PR then?

ChristopherRussell:
We can close it

Issue #532, opened 2018-09-21
Invalid (?) assumption in freeinverse.gi that an Iterator for a free group has !.word bound


wilfwilson:
Currently, in the `stable-4.10` branch of GAP, the `tst/standard/freeinverse.tst` file fails when all packages are loaded. This is because (due to very long-standing a bug in GAP) a method for `Iterator` in RCWA that applies to certain groups ends up having higher rank than the special GAP library `Iterator` method for free groups.

The code in `Iterator` for a free inverse semigroup assumes that `Iterator` for a free group has a component `!.word`, and this presumably records something. I don't see this documented anywhere, so we probably shouldn't rely on this behaviour to exist forever. The test file currently fails because the Iterator for a free group returned by the RCWA method does not have a `!.word` component.

In the first instance, I will attempt to fix this by adding `SUM_FLAGS` to the GAP library method for `Iterator`, even though this won't be required any more when the `master` branch of GAP ends up being the version required by Semigroups.

In the second instance, we should try to make this code more robust.

james-d-mitchell:
Just to mention that the most important point here is that we should not depend on the implementational details of free groups but rather on their documented public methods.

wilfwilson:
Resolved by #540.

james-d-mitchell:
Resolved in v3.0.20.

PR #540, opened 2018-09-26
freeinverse: use non-impl. dep.  free group iterator

james-d-mitchell:
This PR resolves Issue #532 by not using implementation dependent features of the free group iterator. 

Disclaimer: I haven't checked if this really resolves the issue when RCWA is loaded.

### DIFF
--- a/gap/fp/freeinverse.gi
+++ b/gap/fp/freeinverse.gi
@@ -45,8 +45,11 @@ function(iter)
   FG_NextIterator := function(iter)
     local word, output, i;
 
-    NextIterator(iter);
-    word := iter!.word;
+    word := ExtRepOfObj(NextIterator(iter));
+    if IsEmpty(word) then
+      word := ExtRepOfObj(NextIterator(iter));
+    fi;
+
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Also there are probably more issues in the free inverse semigroup iterator code, that I will also try to resolve just now.

wilfwilson:
Resolve them in this PR or in another PR?

james-d-mitchell:
In a separate PR

gap/fp/freeinverse.gi
```diff
@@ -45,8 +45,11 @@ function(iter)
   FG_NextIterator := function(iter)
     local word, output, i;
 
-    NextIterator(iter);
-    word := iter!.word;
+    word := ExtRepOfObj(NextIterator(iter));
+    if IsEmpty(word) then
+      word := ExtRepOfObj(NextIterator(iter));
```
wilfwilson: We simply skip the word corresponding to the identity of the free group; makes sense.

Issue #543, opened 2018-09-28
Semigroups test crashes if all packages are loaded.


alex-konovalov:
As can be seen at https://travis-ci.org/gap-system/gap-docker-pkg-tests-stable-4.10/jobs/433451928, it first goes into some errors, and then suddenly terminates. 

This is the test using stable-4.10 branch and http://www.gap-system.org/pub/gap/gap4pkgs/packages-required-stable-4.10.tar.gz packages archive. 

Note that this is a new test for Semigroups - previously the package had a single `.tst` file, and it recently switched to a more comprehensive standard tests, so we had no chance to observe this before.

```
testing: /home/gap/inst/gap-stable-4.10/pkg/semigroups-3.0.19/tst/standard/fre\
einverse.tst
######## > Diff in:
/home/gap/inst/gap-stable-4.10/pkg/semigroups-3.0.19/tst/standard/freeinverse.\
tst:91
# Input is:
for i in [1 .. 10] do
NextIterator(iter);
od;
# Expected output:
# But found:
Error, Record Element: '<rec>.word' must have an assigned value
########
######## > Diff in:
/home/gap/inst/gap-stable-4.10/pkg/semigroups-3.0.19/tst/standard/freeinverse.\
tst:94
# Input is:
NextIterator(iter);
# Expected output:
a*b
# But found:
Error, Record Element: '<rec>.word' must have an assigned value
########
######## > Diff in:
/home/gap/inst/gap-stable-4.10/pkg/semigroups-3.0.19/tst/standard/freeinverse.\
tst:101
# Input is:
for i in [1 .. 1000] do 
NextIterator(iter);
od;
# Expected output:
# But found:
Error, Record Element: '<rec>.word' must have an assigned value
########
######## > Diff in:
/home/gap/inst/gap-stable-4.10/pkg/semigroups-3.0.19/tst/standard/freeinverse.\
tst:106
# Input is:
for i in [1 .. 100] do 
Add(list1, NextIterator(iter));
od;
# Expected output:
# But found:
Error, Record Element: '<rec>.word' must have an assigned value
########
######## > Diff in:
/home/gap/inst/gap-stable-4.10/pkg/semigroups-3.0.19/tst/standard/freeinverse.\
tst:111
# Input is:
for i in [1 .. 100] do 
Add(list2, NextIterator(iter));
od;
# Expected output:
# But found:
Error, Record Element: '<rec>.word' must have an assigned value
########
######## > Diff in:
[...]

Issue #536, opened 2018-09-22
Iterator for a free inverse semigroup is broken


wilfwilson:
It seems that the iterator for free inverse semigroup only returns products of generators that correspond to reduced products in the free group (minus the identity element). In particular, the iterator produces no words in which a generator is adjacent to its inverse, such as a a^-1. See the first 25 elements produced:

```
gap> iter := Iterator(FreeInverseSemigroup(["a", "b"]));;
gap> for i in [1 .. 25] do Print(NextIterator(iter), "\n"); od;
a
a^-1
b
b^-1
a*a
a^-1*a^-1
b*a
b^-1*a
b*a^-1
b^-1*a^-1
a*b
a^-1*b
a*b^-1
a^-1*b^-1
b*b
b^-1*b^-1
a*a*a
a^-1*a^-1*a^-1
b*a*a
b^-1*a*a
b*a^-1*a^-1
b^-1*a^-1*a^-1
a*b*a
a^-1*b*a
a*b^-1*a
```
Ideally, the iterator should iterate over reduced words ordered by ascending length. I'm not sure what the feasibility of this is, but in any case, it seems like the current iterator will not actually iterate over all reduced words. Yes, the semigroup is infinite, but I would expect that for any particular word, given sufficient time, the word will appear in the iterator. The current behaviour doesn't seem a whole lot more useful than the iterator producing "a, a^2, a^3, ..." as the list of elements.

Furthermore, the code doesn't seem to be doing what is intended. The iterator has record components `seq`, `words`, and `iter_list`, but these seem to remain unchanged after iterations:
```
gap> iter := Iterator(FreeInverseSemigroup(["a", "b"]));;
gap> iter!.seq; iter!.words; iter!.iter_list;
[ 1 ]
[ a ]
[ <iterator> ]
gap> for i in [1 .. 1000] do NextIterator(iter); od;
gap> iter!.seq; iter!.words; iter!.iter_list;
[ 1 ]
[ a ]
[ <iterator> ]
```
Something is clearly not right here.

Furthermore, there is the issue mentioned before, where the code accesses undocumented record components of an iterator of a free group that are not guaranteed to be there. There is a fix for this: we change the `!.word` access to a `ExtRepOfObj` call for the element that we are interested in. But I think it's more important to get the iterator working correctly in a mathematical sense before changing this.

james-d-mitchell:
Resolved in v3.0.20.

PR #544, opened 2018-09-28
Adjust for compatibility with all packages loaded

wilfwilson:
With this PR ~and an adjustment to a filter in `smallsemi` that I mentioned in https://github.com/gap-packages/Semigroups/issues/543#issuecomment-425429322~, the next release of the Semigroups package should pass in Alexander's tests where all the packages are loaded.

Not that this required me to adjust an iterator test for free inverse semigroups; @james-d-mitchell has adjusted this code and presumably the tests, so it would be best to merge this after that fix has been pushed (in case this will need rebasing).

I added `SUM_FLAGS` to the method rank of `IsOrthodoxSemigroup` so that Semigroups beats smallsemi. Ideally `1` should be enough of an adjustment to make this happen, but until the current GAP master branch, method rank stuff was broken in GAP, and this is the simplest way to avoid the brokenness of GAP.

Doing this avoids needing to make changes to `smallsemi`.

### DIFF
--- a/gap/attributes/properties.gi
+++ b/gap/attributes/properties.gi
@@ -1066,7 +1066,7 @@ end);
 # same method for ideals
 
 InstallMethod(IsOrthodoxSemigroup, "for a semigroup",
-[IsSemigroup], 1,  # to beat the Smallsemi method
+[IsSemigroup], SUM_FLAGS,  # to beat the Smallsemi method
 function(S)
   local e, m, i, j;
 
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
Looks good, can you remind me (or do it) about the commented out test? I'm working on the fix for Issue #536 now.

james-d-mitchell:
Also if this resolves Issue #543, then could you please indicate that?

PR #545, opened 2018-09-28
Fix iterators for free inverse semigroups (issue #536) and some new methods for free inverse semigroups

james-d-mitchell:
I think that this resolves Issue #536.

This PR removes the previous iterator code for free inverse semigroups. The code did not function as it was intended. If I understand it correctly it was attempting to enumerate canonical forms for elements of the free inverse semigroup, which are words of the form:

> w_1 w_1 ^ -1 w_2 w_2 ^ -1 ... w_k w_k ^ -1 u 

where w_1, w_2, ..., w_k, u are words in the free group with the same number of generators as the free inverse semigroup. The previous approach (when fixed) essentially iterated through all such expressions. Unfortunately, distinct expressions of the above form do not yield distinct elements of the free inverse semigroup. In fact, of the first 10,000 values iterated through, there were only 119 distinct elements.

In this PR, I have:

1. Fixed (I think) the iterator code, then realised it doesn't work.
2. Commented the iterator code out, in case it can be fixed later.
3. Added methods for Random, \<, and, ChooseHashFunction for free
   inverse semigroups and their elements.
4. Make `FreeInverseSemigroup`s belong to `IsEnumerableSemigroupRep`
4. Tweaked the doc and comments slightly.
5. Add some missing HasIsFinite and not IsFinite to Green's classes.
6. FreeInverseSemigroups know they are infinite now.
7. Do some other minor cleaning up.

### DIFF
--- a/doc/z-chap10.xml
+++ b/doc/z-chap10.xml
@@ -45,9 +45,10 @@
     inverse semigroup: a unique canonical form.  Suppose an element of a free
     inverse semigroup is given as a Munn tree. Let <M>L</M> be the set of words
     corresponding to the shortest paths from the start vertex to the leaves of
-    the tree. Also let <M>w</M> be a word corresponding to the shortest path
-    from start to terminal vertices. The word <M>vv ^ {-1}</M> is an idempotent
-    for every <M>v</M> in <M>L</M>.  The canonical form is given by multiplying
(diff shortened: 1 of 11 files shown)


james-d-mitchell:
@wilfwilson please review when you have a moment

PR #551, opened 2018-10-04
Make method selection-related tweaks to accommodate GAP fixes

wilfwilson:
I've made some adjustments so that GAP's `testinstall` tests in GAP `master` pass when all packages are loaded.

In particular,

* I've changed it so that the direct product code does not apply when the arguments all satisfy `IsGroup`.
* I've added `GreensXClassOfElement` methods that include the filter `IsFinite`, to beat the GAP library ones that we do not expect to be applied when the Semigroups package is loaded
* Sometimes, with all packages loaded, the methods for `IsSemigroupIdeal and IsRegularActingSemigroupRep` would have higher ranks than those for `IsInverseActingSemigroupRep`, even though we would want the latter to apply to ideals in that representation. I've added some `TryNextMethod` calls to avoid this problem.

Overall, the work to accommodate the fixed method ordering in GAP is not complete. There are still diffs in GAP's `testbugfix` suite when all packages are loaded. Furthermore, the Semigroups package tests do not run. Hopefully it'll only require small changes like these to fix everything.

### DIFF
--- a/gap/greens/gree.gi
+++ b/gap/greens/gree.gi
@@ -244,9 +244,8 @@ InstallMethod(GreensJClassOfElementNC,
 # Green's classes of an element of a semigroup
 
 InstallMethod(GreensRClassOfElement,
-"for a finite semigroup and multiplicative element",
+"for a possibly finite semigroup and multiplicative element",
 [IsSemigroup, IsMultiplicativeElement],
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
Nice, thanks for this.

PR #552, opened 2018-10-05
gh-pages: use the standard format for "citing"

mtorpey:
You can see the result of this pull request at https://mtorpey.github.io/Semigroups/

We first merge the latest version of https://github.com/gap-system/GitHubPagesForGAP, which includes the desired "Citing" section, then we run `update.g` (1aa9003) to generate the pages properly.  This produces a not-so-good bibtex entry, which we then overwrite with your preferred style in 83adcb9.  Finally we remove the pre-existing "How to cite" section (23034af).

Merging this without squashing or rebasing might make it easier to merge from https://github.com/gap-system/GitHubPagesForGAP in the future.


_data/package.yml
```diff
@@ -51,7 +51,7 @@ needed-pkgs:
       url: "https://gap-packages.github.io/io"
     - name: "digraphs"
       version: ">=0.12.0"
-      url: "https://gap-packages.github.io/Digraphs"
+      url: "https://gap-packages.github.io/digraphs"
```
wilfwilson: This gives a 404 without the capital letter. How can we avoid this happening?

_data/package.yml
```diff
@@ -79,28 +79,20 @@ keywords: |
...
-      year =             {2018},
-      note =             {GAP package},
-      howpublished =     {\href                    {https://gap-packages.github.io/Semigroups}
-                          {\texttt{https://gap-packages.github.io/}\discretionary
-                          {}{}{}\texttt{Semigroups}}},
-      keywords =         {transformation semigroups; partial permutations; inverse semigroups;
-                          Green's   relations;   free  inverse  semigroup;  partition  monoid;
-                          bipartitions; Rees matrix semigroups},
-      printedkey =       {Mit18}
+    @manual{ Mitchell2018aa,
```
wilfwilson: Will this be updated automatically with new version numbers and release dates etc?

_data/package.yml
```diff
@@ -79,28 +79,20 @@ keywords: |
     transformation semigroups, partial permutations, inverse semigroups, Green's relations, free inverse semigroup, partition monoid, bipartitions, Rees matrix semigroups.
 citeas: |
     <p class='BibEntry'>
-    [<span class='BibKey'>Mit18</span>]   <b class='BibAuthor'>Mitchell, J.</b>,
-     <i class='BibTitle'>Semigroups, A package for semigroups and monoids,
-             Version 3.0.20</i>
-     (<span class='BibYear'>2018</span>)<br />
-    (<span class='BibNote'>GAP package</span>),
-    <span class='BibHowpublished'><a href="https://gap-packages.github.io/Semigroups">https://gap-packages.github.io/Semigroups</a></span>.
+    [<span class='BibKey'>Mit18</span>]   <b class='BibAuthor'>J. D. Mitchell et al,</b>,
```
wilfwilson: (I'm not sure whether `Mit18` is James's preferred cite key.)

mtorpey:
Disclaimer: I don't know much about this stuff, and I've just used it in an automatic way in other packages.  But I think the idea is that you can customise it a lot, so I've tried to do that here.

`_data/package.yml` is generated by running `update.g`, but the idea seems to be that you run it and then manually commit the changes it makes.  So I think we can just change it manually as we please.  In order to keep the citation up to date, a lot of it would have to be updated manually or as part of a release script (as I believe already happens with the existing citation).

_data/package.yml
mtorpey: This can be changed manually, and we should!  I'll add a commit now...

mtorpey: (to be clear, this was my mistake and not the result of a script)

_data/package.yml
mtorpey: Nope.  I tried using `{ Mitchell{{site.data.package.date | date: "%Y"}}aa,` as in the citation bit I removed, but the `{{...}}` bit wasn't interpreted.  Not sure if there's a better way, but at the moment this would require manual updating each time.

_data/package.yml
mtorpey: I'm not sure either.  There currently isn't one, so I just went for the default.

wilfwilson: Okay :) My guess would be 'no bib key'... but we'll see!

_data/package.yml
wilfwilson: What about the rest of the information in this `@manual` section? Previously it was all updated automatically in the bit you removed, but if needs to be done manually now, then I don't see how that's an improvement.

mtorpey: I'm starting to think you're right!

James asked if I could update the website to look like [the PackageManager one](https://gap-packages.github.io/PackageManager/), so I figured I'd do it in the same way – but I'm starting to realise the setup used there is geared towards using ReleaseTools to update everything, and it's not all that applicable to this.  Perhaps a different approach is required...

wilfwilson: That's very nice of you to do this. Hopefully James will be able to adjust this to get automatic updating into his release process.

mtorpey: I'm a very nice man! :D

But I'm starting to think that it'd be better just to change how the thing looks on the surface rather than by changing all these processes (or worse, trying to change the behaviour of `BibEntry`)...

fingolfin: Just to clarify: normally with GitHubPagesForGAP, this whole file is autogenerated by the `update.g` script. I have not checked if that is what you are doing, but I hope it is, as that is one of the key ideas behind the whole thing.

wilfwilson: Our `index.md` contains the following at the bottom to generate the bib entry:
```
<pre style="white-space: pre-wrap;">@manual{ Mitchell{{site.data.package.date | date: "%Y"}}aa,
         Author = { Mitchell, J. D. and others },
         Month  = { {{site.data.package.date | date: "%b"}} },
         Title  = { Semigroups - GAP package, Version {{site.data.package.version}} },
         Doi    = { 10.5281/zenodo.592893 },
         Url    = { http://dx.doi.org/10.5281/zenodo.592893 },
         Year   = { {{site.data.package.date | date: "%Y"}} }
}</pre>
```
Is that not a good idea?

fingolfin: Seems fine to me. What I mean above was: "I hope you are indeed running `update.g` to update `_data/package.yml`, instead of, say, manually updating it". Whether you use the bib information in it or not is of course up to you (that bib information is generated by GAP's `Cite` command, and one could certainly think about improving that; all packages using `Cite` to help generated their gh-pages would then immediately benefit from it). But of course in the meantime, if the output of `Cite` does not suite you, I think it's perfectly reasonable to roll your own.

fingolfin: I.e., I thought in the discussion above, you guys were discussing whether this `@manual` entry here had to be manually updated; my answer is: "no, it is generated for you". That's all I wanted to say -- sorry if I misunderstood the conversation here :/

wilfwilson:
@mtorpey It seems like, at the moment, this would require manually editing the `.yml` file to get our BibTex to appear the way we want, once `update.g` is run. So I don't propose merging this as-is.

(To demonstrate, with your PR, when I run `update.g`, I get the following diff:
```
 bibtex: |
-    @manual{ Mitchell2018aa,
-      Author = { Mitchell, J. D. and others },
-      Month  = { Oct },
-      Title  = { Semigroups - GAP package, Version 3.0.20 },
-      Doi    = { 10.5281/zenodo.592893 },
-      Url    = { http://dx.doi.org/10.5281/zenodo.592893 },
-      Year   = { 2018 }
+    @misc{ Semigroups3.0.20,
+      author =           {Mitchell, J.},
+      title =            {{Semigroups},  A  package  for semigroups and monoids,
+                          {V}ersion 3.0.20},
+      month =            {Oct},
+      year =             {2018},
+      note =             {GAP package},
+      howpublished =     {\href      {https://gap-packages.github.io/Semigroups}
+                          {\texttt{https://gap-packages.github.io/}\discretionary
+                          {}{}{}\texttt{Semigroups}}},
+      keywords =         {transformation   semigroups;   partial   permutations;
+                          inverse  semigroups;  Green's  relations; free inverse
+                          semigroup; partition monoid; bipartitions; Rees matrix
+                          semigroups},
+      printedkey =       {Mit18}
     }
```
and we don't want to have to change that back the way we want manually each time).

PR #557, opened 2018-11-09
Use 'xenial' in Travis rather than 'trusty'

wilfwilson:
Travis have released their new `xenial` environment, which is a build of Ubuntu from 2016, rather than `trusty`, which is from 2014. By using this, we don't have to install loads of extra updates. In particular, `xenial` includes `gcc-5.4.0`. This makes the `.travis.yml` file a lot simpler. And the builds should now be a bit quicker because of not having to do these updates any more.

I also stop installing the profiling package in the non-coverage tests, to speed them up.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,11 @@
 dist:
-  trusty
+  xenial
 
 language:
   cpp
 
 compiler:
   gcc
-
-env:
-  global:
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Nice! Thanks @wilfwilson

PR #558, opened 2018-11-09
Rename 'IsDecomposableSemigroup' to 'IsSurjectiveSemigroup'

wilfwilson:
In my viva, Nik Ruskuc pointed out that decomposable is not a good name for a semigroup `S` that satisfies `S^2 = S`. Instead, he suggested `surjective' as a better name (along with several others). It makes sense because a semigroup satisfies S^2 = S if its multiplication is surjective. This is the name that I have moved to in my PhD thesis.

Since `IsDecomposableSemigroup` was never included in a released version of the package, there should be no problems with renaming it.

### DIFF
--- a/doc/attr.xml
+++ b/doc/attr.xml
@@ -1206,7 +1206,7 @@ gap> GeneratorsSmallest(T);
     <A>S</A>. An element of <A>S</A> is <E>indecomposable</E> if it is not
     decomposable. <P/>
     
-    See also <Ref Prop="IsDecomposableSemigroup" />. <P/>
+    See also <Ref Prop="IsSurjectiveSemigroup" />. <P/>
 
     Note that any generating set for <A>S</A> contains each indecomposable
     element of <A>S</A>. Thus <C>IndecomposableElements(<A>S</A>)</C> is a
--- a/doc/properties.xml
+++ b/doc/properties.xml
@@ -1133,38 +1133,38 @@ false]]></Example>
   </ManSection>
 <#/GAPDoc>
 
-<#GAPDoc Label="IsDecomposableSemigroup">
+<#GAPDoc Label="IsSurjectiveSemigroup">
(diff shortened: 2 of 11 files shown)


james-d-mitchell:
Hmm, I was going to review this, I don't much like the name `IsSurjectiveSemigroup` but I can live with it...

wilfwilson:
Sorry @james-d-mitchell, I know it's not the done thing to merge it myself.

My excuse: I made my final changes to my PhD thesis today. As part of the changes, I now name and  describe the functions that I've implemented in the Semigroups package related to my PhD thesis. I wanted to make sure that my description was accurate and consistent. So now the name of the function in the package matches the description of the function in my thesis, which matches the name of the property in my thesis (surjective). I need to resubmit my thesis soon, which is why I merged this – otherwise of course this is neither an important nor urgent change and I would have waited longer for your comment.

I don't really like the name either. But I had an interesting discussion about possible names with Nik in my viva. Of the options we talked about, I preferred surjective. It has the benefit of seeming a bit more specific, and being used in the literature (according to Nik). Nik's problem with decomposable is that the idea of a semigroup being "decomposable" is very general, and in some sense it should mean that the semigroup can be broken up into parts, and that there should be some kind of product that lets you create the whole from those parts.

But using `IsDecomposableSemigroup` for this specific instance of `S = S^2` maybe isn't the best idea, especially since it's not an interesting decomposition: the decomposition is just of one semigroup into two copies itself (so there aren't "smaller" components that it decomposes into), and the product is just set product. A semigroup might not be "surjective", as I now define it, but it could have many other interesting decompositions (for instance, it could decompose as a direct product), and yet in the Semigroups package it would have `IsDecomposableSemigroup` return `false`.

I hope that helps you feel a bit better about moving away from the name `IsDecomposableSemigroup`.

PR #566, opened 2018-11-25
Update semidp.tst for compatibility with future GAP version

wilfwilson:
I have created PR https://github.com/gap-system/gap/pull/3053 for GAP, which adds additional properties that are transferred from the factors of a direct product to the direct product, at creation.

This means that, once this is merged, many direct products will know more about themselves at creation than they currently do. This means the test output in `semidp.tst` and `semidp.xml` needs to be updated. However, since we still need to support previous versions of GAP, we can't just update the files with the new output. We have to make sure that the properties are definitely computed before viewing the direct product, and showing the new output. That's why there are lots of new calls to `IsRegularSemigroup`, etc.

The tests on the `GAP stable-4.10, 64-bit` Travis job will fail until #565 is fixed and merged.

### DIFF
--- a/doc/semidp.xml
+++ b/doc/semidp.xml
@@ -43,9 +43,9 @@ gap> D := DirectProduct(S, T);
 <commutative inverse partial perm monoid of rank 5 with 1 generator>
 gap> Elements(D);
 [ <identity partial perm on [ 1, 2, 3, 4, 5 ]>, (1,2)(3)(4)(5) ]
-gap> S := PartitionMonoid(3);;
-gap> D := DirectProduct(S, S, S);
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
@wilfwilson rerunning the CI now.

wilfwilson:
The corresponding PR for GAP has been merged, faster than I expected, so I think I have to merge this now to stop the Travis tests going crazy.

alex-konovalov:
@wilfwilson this should NOT go to Semigroups release intended for GAP 4.10.1, isn't it?

alex-konovalov:
@wilfwilson thanks, that's reassuring!

PR #567, opened 2018-11-25
The empty semigroup is not monogenic

wilfwilson:
Currently, an empty semigroup is set to be monogenic by an immediate method. But I don't think that it should count as monogenic: its rank isn't 1, and there exists no element inside the empty semigroup that generates the empty semigroup.

I know that we normally avoid empty semigroups, and for certain properties it's probably not very clear whether the empty semigroup has that property. But here I think it's clear so we may as well fix it.

### DIFF
--- a/gap/attributes/properties.gi
+++ b/gap/attributes/properties.gi
@@ -777,7 +777,7 @@ function(S)
   gens := GeneratorsOfSemigroup(S);
   if Length(gens) <= 1 then
     SetMinimalSemigroupGeneratingSet(S, gens);
-    return true;
+    return Length(gens) = 1;
   elif CanEasilyCompareElements(gens)
       and ForAll([2 .. Length(gens)], i -> gens[1] = gens[i]) then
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
@wilfwilson this seems to cause some of the GAP tests to fail. Any thoughts?

wilfwilson:
@james-d-mitchell the test failures were unrelated to this PR, and were caused by first this https://github.com/gap-system/gap/pull/3052 and then this https://github.com/gap-system/gap/pull/3053 being merged into GAP master. For the first, this caused our `testinstall.tst` to fail, which I had forgotten to check, whoops. I fixed this by pushing directly to the branch (`testinstall: stop printing an empty semigroup`). For the second, I knew this PR would break the Semigroups tests, but I had already made the fix in https://github.com/gap-packages/Semigroups/pull/566 - it just needed to be merged 🙂 

So (having changed nothing about this PR), the tests on this should pass now.

james-d-mitchell:
Thanks @wilfwilson !

PR #574, opened 2019-02-04
Avoid using EQ_INTOBJS

fingolfin:
This is the only place which currently is using EQ_INTOBJS, among
all GAP packages and the whole GAP kernel. Not using it here allows
to either remove EQ_INTOBJS from GAP, or to refactor it.


wilfwilson:
Thanks @fingolfin. Happy to adopt the change.

PR #575, opened 2019-02-06
Bib update

MTWhyte:
This pull request updates the bibliography, for example updating entries which suggest that a paper is in progress, when in fact the paper has been published a couple of years ago.

In an effort to present the bibliography nicely in HTML, PDF and text formats, the format of the bibliography format has been changed from BibTeX to BibXMLext, as suggested in the [GAPDoc documentation](https://www.gap-system.org/Manuals/pkg/GAPDoc-1.6.2/doc/chap7.html#X7A4126EC7BD68F64).

Note that in the first commit, the change I made to the file main.xml is actually relevant to the second commit, as it changes the file recognised as the bibliography.


flsmith:
@james-d-mitchell 

opinions on this XML format?

james-d-mitchell:
I’m happy with this, many thanks to @mtwhyte for doing this!

PR #584, opened 2019-03-06
Conditionally install libsemigroups, and drop src/ prefix for its headers

dimpase:
conditionally install libsemigroups; nothing changes if there is no external libsemigroups
known to pkg-config and matching the version test; there is an extra option `--with-included-libsemigroups` to force the use of the included library (it is needed to `make dist`
including libsemigroups). The included libsemigroups will also be used if no external one is good enough.

Dealing with libsemigroups (version check, etc) is moved into m4/ subdir macro.

`src/` prefix is removed from for libsemigroups headers `#include`'s, and `src/libsemigroups` is moved one level up the directory tree.

### DIFF
--- a/m4/ax_check_libsemigroup.m4
+++ b/m4/ax_check_libsemigroup.m4
@@ -1,0 +1,52 @@
+dnl handle libsemigroups checks
+dnl
+dnl if --with-included-libsemigroups is supplied, always use the included version
+dnl otherwise, use an external version if it is known to pkg-config and is new enough
+dnl failing the latter, use the included version
+dnl
+AC_DEFUN([AX_CHECK_LIBSEMIGROUPS], [
+  AC_ARG_WITH([included-libsemigroups],
+	      [AC_HELP_STRING([--with-included-libsemigroups],
+			      [use the libsemigroups included here])])
+  REQUI_LIBSEMIGROUPS_VERSION="$(cat .LIBSEMIGROUPS_VERSION)"
+  need_included_libsemigroups=yes
+  PKG_CHECK_MODULES([LIBSEMIGROUPS], [libsemigroups >= $REQUI_LIBSEMIGROUPS_VERSION], 
+  	[need_included_libsemigroups=no],
+  	[need_included_libsemigroups=yes])
+
+  if test "$need_included_libsemigroups" = yes; then
+     AC_MSG_NOTICE([Using included libsemigroups.])
+     if test "$with_included_libsemigroups" = no; then
+	with_included_libsemigroups=yes
+     fi
+  fi
+
+  if test "$with_included_libsemigroups" = yes;  then
+  	AC_CHECK_FILE(
+   		[libsemigroups/src/semigroups.h],
+   		[],
+   		[AC_MSG_ERROR([libsemigroups is required, clone or download the repo from
+    			https://github.com/james-d-mitchell/libsemigroups into this directory])])
+
+	AC_CHECK_FILE(
+   		[libsemigroups/VERSION],
+   		[],
+   		[AC_MSG_ERROR([libsemigroups version 0.6.1 or higher is required])])
+
+	AC_MSG_CHECKING([libsemigroups version])
+	FOUND_LIBSEMIGROUPS_VERSION="$(cat libsemigroups/VERSION)"
+	AC_MSG_RESULT([$FOUND_LIBSEMIGROUPS_VERSION])
+	AX_COMPARE_VERSION($FOUND_LIBSEMIGROUPS_VERSION,
+                   [ge],
+                   $REQUI_LIBSEMIGROUPS_VERSION,
+                   [],
+                   [AC_MSG_ERROR([libsemigroups version $REQUI_LIBSEMIGROUPS_VERSION or higher is required])]
+                  )
+	LIBSEMIGROUPS_CFLAGS = -I$(srcdir)/libsemigroups
+	LIBSEMIGROUPS_LIBS = libsemigroups/libsemigroups.la
+
+  fi
+  AC_CONFIG_SUBDIRS([libsemigroups])
+
+  AM_CONDITIONAL([WITH_INCLUDED_LIBSEMIGROUPS], [test "$with_included_libsemigroups" = yes])
+])
--- a/m4/ax_check_libsemigroup.m4
+++ b/m4/ax_check_libsemigroup.m4
@@ -38,7 +38,7 @@ AC_DEFUN([AX_CHECK_LIBSEMIGROUPS], [
                    [AC_MSG_ERROR([libsemigroups version $REQUI_LIBSEMIGROUPS_VERSION or higher is required])]
                   )
 	AC_SUBST(LIBSEMIGROUPS_CFLAGS, ['-I./bin/include'])
-	AC_SUBST(LIBSEMIGROUPS_LIBS, ['libsemigroups/libsemigroups.la'])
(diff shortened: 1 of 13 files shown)


dimpase:
I've dropped the commit 129bb5e for the time being, perhaps it should be dealt with on another PR.

james-d-mitchell:
This is very welcome @dimpase. The CI has failed though, looks like the compiler isn't finding the headers `libsemigroups/whatever.h`.

m4/ax_check_libsemigroup.m4
```diff
@@ -0,0 +1,52 @@
...
+  	AC_CHECK_FILE(
+   		[libsemigroups/src/semigroups.h],
+   		[],
+   		[AC_MSG_ERROR([libsemigroups is required, clone or download the repo from
+    			https://github.com/james-d-mitchell/libsemigroups into this directory])])
+
+	AC_CHECK_FILE(
+   		[libsemigroups/VERSION],
+   		[],
+   		[AC_MSG_ERROR([libsemigroups version 0.6.1 or higher is required])])
```
james-d-mitchell: It'd be better if the version number wasn't hard coded here, but I appreciate that this is the same as it was previously (and not a change that you are introducing).

dimpase:
The CI error is my m4 error, to be fixed, sorry.

m4/ax_check_libsemigroup.m4
dimpase: So what should it be, instead of 0.6.1?

james-d-mitchell: The value in the file `.LIBSEMIGROUPS_VERSION` I think.

dimpase:
Now I am confused - first of all, how does one make tarfiles for GAP here? Does `make dist` mean to work?

dimpase:
OK, it's just the headers of libsemigroups that are in the wrong place...

james-d-mitchell:
No, `make dist` is not used here, I'll make the archive when the time is right. We need the PR to pass the CI first.

m4/ax_check_libsemigroup.m4
dimpase: fixed by bab804e

james-d-mitchell:
Thanks @dimpase. Just to be clear, in this PR:
* the system `libsemigroups` is used if available and recent enough
* if the system `libsemigroups` is not available, then the included one is `make install`ed, and then used. 

If I'm understanding this properly, then I prefer the current setup in `Semigroups` to be the default, and that the included `libsemigroups` is not `make install`ed. More precisely, I'd rather see:
* the included `libsemigroups` used if available and recent enough by default
* if the included `libsemigroups` is not available, then use the system version (if available and recent enough). 

One way to overcome the difficulties with the header includes requiring  different prefixes, i.e
`#include "libsemigroups/semigroups.h"` versus `#include "libsemigroups/src/semigroups.h"`, we could define a macro that would go into `semigroups-config.h` (like `SEMIGROUPS_DEBUG`), say `SEMGROUPS_USE_SYSTEM_LIBSEMIGROUPS` and then include `libsemigroups` headers as:
~~~~
#ifdef SEMIGROUPS_USE_SYSTEM_LIBSEMIGROUPS
#include <libsemigroups/semigroups.h>
#else
#include "libsemigroups/src/semigroups.h"
#endif 
~~~~
Maybe there are other better solutions.

I don't think many of our current users are using a system `libsemigroups`, and the approach outlined above is more compatible with the current approach.

dimpase:
You explicitly agreed with getting rid of `src/` in https://github.com/gap-packages/Semigroups/issues/580#issuecomment-469226293 and this is done here.
And I don't get why one wants to unnecessarily pollute the sources with `#ifdefs`.

dimpase:
If you are worried that if you touch a header in libsemigroups/ and it does not get updated when you rebuild, please rest assured that it's taken care of.

Perhaps ask @fingolfin to review this, he's a Make expert...

dimpase:
The bottom line is that the only difference between not having ifdefs and having ifdefs is bin/include directory (which may be made to vanish at the end of the build, if you like).

james-d-mitchell:
@fingolfin any comments?

dimpase:
By the way, calling `make -C libsemigroups` makes these `cp <..> libsemigroups/.libs/*` not needed, as it does the same thing, only cleaner. (@fingolfin probably agrees here, as he prefers `install` to `cp`, in general :-))

(Perhaps on CYGWIN one still needs more, I don't know, this needs testing, on a platform I don't have easy access to.)

james-d-mitchell:
Maybe I misunderstood what is going on, I just tried this, and it seems that this is not doing a `make install` in the sense I understood.  Ok, so the only change I'd request is that the logic is reversed, use the included `libsemigroups` by default and the system one optionally. With the flag`--with-system-libsemigroups` or something similar. Does that seem reasonable?

james-d-mitchell:
And I agree it's better to remove the `cp`'s.

dimpase:
> flag`--with-system-libsemigroups`

I will do this - but first I really would like to hear from Max, I trust him more than myself on all make-related things. Perhaps more should be done on ensuring that headers of libsemigroups really stay in sync with the contents of libsemigroup/.

james-d-mitchell:
In the absence of any further comments, I'll be happy enough to merge this with the changes we agreed @dimpase/

dimpase:
0beef21 does the promised `--with-system-libsemigroups` (well, I called it `--with-external-libsemigroups` as a better fit)

james-d-mitchell:
Apologies for the delay, in looking at this. This seems to work, but I have one remaining concern. When I have `libsemigroups` installed, and I do:
~~~~
./configure --with-external-libsemigroups
~~~~
then the configure script of the included `libsemigroups` is run. Is that the expected behaviour? It seems if the included `libsemigroups` directory is not present at all, then everything works ok.

dimpase:
> When I have `libsemigroups` installed, and I do:
> 
> ```
> ./configure --with-external-libsemigroups
> ```
> then the configure script of the included `libsemigroups` is run. Is that the expected behaviour?

yes, it is as expected. And then nothing gets built in/installed from  `libsemigroups`.

james-d-mitchell:
Ok, thanks, but what's the point in running `./configure` in the included `libsemigroups` and then not using anything in the included `libsemigroups`? Can't this be avoided by moving 

~~~~
 AC_CONFIG_SUBDIRS([libsemigroups])
~~~~

in `ax_check_libsemigroups` into the `if` clause two lines up? This seems to work on my computer.

dimpase:
OK, I was under false (according to the [manual](https://www.gnu.org/software/autoconf/manual/autoconf-2.60/html_node/Subdirectories.html)) impression that one should not put AC_CONFIG_SUBDIRS into an `if` block, but apparently it's the way to go.

dimpase:
Would you like me to change this, or can you do this change yourself?

james-d-mitchell:
Either way is fine by me!

dimpase:
They [say](https://stackoverflow.com/questions/15230696/best-practice-for-building-with-out-included-libraries) that the reason for unconditional AC_CONFIG_SUBDIRS is to make `make dist` more robust. But as here `make dist` is broken (perhaps another PR  to fix it?) it does not matter too much.

Perhaps a remark in the docs that before trying `make dist` one has to `./configure` without  `--with-external-libsemigroups` would suffice.

-----

Anyhow, please go ahead and change this as you see fit.

PR #590, opened 2019-03-14
Fix issue 588 - check if PKG_CHECK_MODULES is defined

james-d-mitchell:


james-d-mitchell:
Thanks @dimpase, I'll wait until the CI finishes, then I'll merge this.

m4/ax_check_libsemigroup.m4
```diff
@@ -11,17 +11,19 @@ AC_DEFUN([AX_CHECK_LIBSEMIGROUPS], [
...
   if test "$with_external_libsemigroups" = yes;  then
-	PKG_CHECK_MODULES([LIBSEMIGROUPS], [libsemigroups >= $REQUI_LIBSEMIGROUPS_VERSION],
-	[need_included_libsemigroups=no],
-	[need_included_libsemigroups=yes])
+	m4_ifdef([PKG_CHECK_MODULES], [
+	PKG_CHECK_MODULES([LIBSEMIGROUPS], 
+                          [libsemigroups >= $REQUI_LIBSEMIGROUPS_VERSION],
+			  [need_included_libsemigroups=no],
+			  [need_included_libsemigroups=yes])],
+	[AC_MSG_NOTICE([ignoring flag --with-external-libsemigroups, pkg-config is not available])])
```
dimpase: This error message is not strictly correct. This merely checks that the pkg.m4 macros are present. Yet, if they are present, one probably would also want to check that `pkg-config` itself is there and working. This can be done with PKG_PROG_PKG_CONFIG() [macro](https://people.freedesktop.org/~dbn/pkg-config-guide.html#using) and error out, or  print a warning, if it's not the case.

dimpase: The error message should say `Semigroups ./configure was created on a system without m4 macros for pkg-config available.`

Otherwise, in particular if one distributes a tarball built on such a system, its users, with all the pkg-config tools present, will be misled by the present message.

james-d-mitchell:
ok will do.

dimpase:
IMHO t would also be an improvement if you check for presence of pkg-config, in the case of 
`--with-external-libsemigroups`, and output a warning if it's not there. But OK, it's minor compared to my other point.

james-d-mitchell:
I'll do both just now.

m4/ax_check_libsemigroup.m4
james-d-mitchell: Yup, agreed.

james-d-mitchell:
I think I've made the change you suggested @dimpase, is this what you meant?

dimpase:
The warning message fix is fine.

But `PKG_PROG_PKG_CONFIG()` should not be where it is now --- otherwise it's back to almost the same error as before this PR. Let's just remove it---sorry for noise. I think it's clear enough without it.

james-d-mitchell:
I've removed it, no worries, thanks for the review!

PR #591, opened 2019-03-14
travis: configure with --with-external-libsemigroups

james-d-mitchell:
In this PR I add a travis test for when Semigroups is configured with the flag `--with-external-libsemigroups`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,6 +75,25 @@ matrix:
       - PACKAGES=required
       - THRESHOLD=95
 
+    - env:
+      - SUITE=external-libsemigroups
+      - ABI=64
+      - GAP=stable-4.10
+      - PACKAGES=required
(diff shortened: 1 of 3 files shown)


dimpase:
Just in case: is conda playing with `pkg-config` right? I.e., to me there is no guarantee that the config of libsemigroup lands in the place where pkg-config picks it up without an extra parameter, as I describe here: https://github.com/james-d-mitchell/libsemigroups/pull/113 (I'm not a conda user, so perhaps my suspicion here is misplaced).

james-d-mitchell:
I'm going to close this for now.

dimpase:
Have you got it working?

james-d-mitchell:
Yes, in PR #593, there were various complications including the one you mentioned.

PR #593, opened 2019-03-14
travis: configure with --with-external-libsemigroups

james-d-mitchell:
Another attempt at PR #591.


james-d-mitchell:
So, @wilfwilson if you don't know that it's not working, then it must be working, right?

james-d-mitchell:
I just put the travis test back in their usual order, and then I think this can be merged.

wilfwilson:
Fixed my typo ;)

PR #605, opened 2019-06-05
Add IsSemigroupWithClosedIdempotents

wilfwilson:
This is a sensible feature to have, in my opinion, especially since the code for it already exists pretty much as part of `IsOrthodoxSemigroup`. There is a problem: this already exists in `smallsemi`, and `IsSemigroupWithClosedIdempotents` is also declared there, so loading both packages gives the warning:
```
#I  method installed for `IsSemigroupWithClosedIdempotents` matches more than one declaration
```
What should we do about this?
1. Ignore it.
2. Move the declaration of `IsSemigroupWithClosedIdempotents` to the GAP library.
3. Not have this feature in Semigroups.

### DIFF
--- a/doc/properties.xml
+++ b/doc/properties.xml
@@ -844,12 +844,37 @@ true]]></Example>
   </ManSection>
 <#/GAPDoc>
 
+<#GAPDoc Label="IsSemigroupWithClosedIdempotents">
+  <ManSection>
+    <Prop Name = "IsSemigroupWithClosedIdempotents" Arg = "S"/>
+    <Returns><K>true</K> or <K>false</K>.</Returns>
+    <Description>
+      <C>IsSemigroupWithClosedIdempotents</C> returns <K>true</K> if the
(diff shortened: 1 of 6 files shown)


james-d-mitchell:
You missed an option: by now Smallsemi is on github, so we could patch it also?

wilfwilson:
Since neither package requires the other, they either both need to declare `IsSemigroupWithClosedIdempotents`, or it needs to be declared in the GAP library.

james-d-mitchell:
Can't they each check if the other is loaded already and declare the method if not?

wilfwilson:
Yeah, I think so. In that case we'll need to update smallsemi to declare the method for `IsSemigroup` rather than just `IsSmallSemigroup`.

james-d-mitchell:
Sounds good to me.

james-d-mitchell:
See https://github.com/gap-packages/smallsemi/pull/3

wilfwilson:
Cool, all updated now 🙂

james-d-mitchell:
Thanks @wilfwilson !

Issue #556, opened 2018-11-08
There should be no implications to `IsEnumerableSemigroupRep`


fingolfin:
As discussed in https://github.com/gap-system/gap/issues/2818, the following code in Semigroups is problematic:
```gap
InstallTrueMethod(IsEnumerableSemigroupRep, IsGroup and IsFinite);
```
It is problematic because it means that now GAP consider every finite group to be in the filter `IsEnumerableSemigroupRep`, and hence in the filter `IsComponentObjectRep` -- even objects which are in positional rep! Clearly, that's wrong.

IMHO, this extends to all implications to `IsEnumerableSemigroupRep`. A possible fix would be to add a new filter/category/property `IsEnumerableSemigroup`; then change `IsEnumerableSemigroupRep` to imply that, and change the above implications to only imply `IsEnumerableSemigroup`. You might also want or need to change some more places to use `IsEnumerableSemigroup` instead of `IsEnumerableSemigroupRep`.

james-d-mitchell:
Thanks for the report @fingolfin, this is obviously a bad idea. I intend to get rid of `IsEnumerableSemigroupRep` in the near future anyway, and will certainly fix it at that point. Can the fix wait, or should I do it more quickly?

fingolfin:
Thank you! Resolving this can wait, the immediate issue it caused also revealed bad code on the GAP side, which is now fixed, so the immediate problem there is  resolved now.

fingolfin:
Just for the record, it would be nice if there was some progress on this in time for GAP 4.11, which is hope we can get out in April/May or so. But if that's not feasible, well, so be it, we'll manage.

james-d-mitchell:
I'll try my best to meet the May deadline, and will update whatever progress I make.

fingolfin:
My apologies for nagging, but: I just see this was added to the 4.0.0 milestone (though it also has the 3.* label). This sounds as if the bigger change of getting rid of `IsEnumerableSemigroupRep` might be farther away. Perhaps in the meantime, a smaller fix could be implemented, which just gets rid of the bad implication, as outlined by me above?

I'd be happy to discuss this tomorrow, if needed :-)

james-d-mitchell:
No need to apologise, my (perhaps baseless) thoughts were that this is not so easy to fix and could (probably) be most easily resolved when I reorganise things for 4.0.0. I'm happy to try to find a resolution sooner than that, I'll try to figure out what to do today.

[...]

PR #607, opened 2019-06-10
 Fix issue 556

james-d-mitchell:
This PR partially fixes Issue #556, by removing the implication `IsFinite and IsGroup` => `IsEnumerableSemigroupRep`.  Other such implications remain, and should be removed. The remaining implications will be removed in a more major clean up (hopefully) in the coming months.


james-d-mitchell:
Not trying to hurry you @wilfwilson but in case you didn't see this earlier, I requested your review of this PR.

Issue #608, opened 2019-06-21
Seg fault in bipartitions code


james-d-mitchell:
I get a seg fault when I do:

~~~~
test := function(S)
  local SS, XX, map1, map2, Y, T, R, z, r, xx;

  SS := DirectProduct(S, S);
  XX := GeneratorsOfSemigroup(SS);
  map1 := Embedding(SS, 1);
  map2 := Embedding(SS, 2);
  Y := List(GeneratorsOfSemigroup(S), x -> x ^ map1 * x ^ map2);
  T := Semigroup(Y);

  R := [RClassNC(T, One(SS))];

  for r in R do 
    for xx in XX do 
      z := xx * Representative(r);
      if not ForAny(R, C -> z in C) then 
        Add(R, RClassNC(T, z));
        Print("found ", Length(R), " so far\n");
      fi;
    od;
  od;
  return R;
end;
test(UniformBlockBijectionMonoid(6));
~~~~~

This is with GAP 4.10.1 and Semigroups `master` branch. It appears to intermittently seg fault in `BIPART_PERM_LEFT_QUO`, I haven't had the chance to investigate further at present.

james-d-mitchell:
I think this may be a non-issue, I've rebuilt  everything, and can't seem to reproduce this. @flsmith also couldn't reproduce it.

james-d-mitchell:
Now, I'm not sure, further investigation is required.

james-d-mitchell:
This does indeed appear to be a bug.

The-Beast:
I also got a seg fault.

james-d-mitchell:
Resolved in v3.1.4

PR #609, opened 2019-07-01
kernel: fix issue 608

james-d-mitchell:
This PR fixes Issue #608. For future reference, I only had this issue when Semigroups is compiled with optimisations on (`-O3`) and is compiled with clang. As far as I can see, the code before this change is valid and does the same thing as after this change.  It might be that the bug is elsewhere in the code, or that some part of the code has undefined behaviour, so that the optimised code is broken, while the non-optimised code is not.


james-d-mitchell:
@wilfwilson I can't any longer stop the builds for Semigroups on travis-ci, I have no idea why. Any ideas?

james-d-mitchell:
Ok, I can edit travis again, must have been a temporary glitch.

james-d-mitchell:
Yeah, I'm not sure this isn't hiding something somewhere else either. I'm going to merge, for now since it seems to fix the issue.

PR #610, opened 2019-07-03
Add DigraphCore method

MTWhyte:
This pull request adds a method `DigraphCore`, which takes a digraph and returns its core.

The core of a digraph _D_ is the (unique up to isomorphism) minimal subdigraph which is a homomorphic image of _D_.

Note that I have rebased to James's `Update for Digraphs v1.0` branch.


wilfwilson:
Thanks for this. I’ve not looked at it in detail, but my initial impression that this was weird that this is in the semigroups package and not the digraphs package. Then I saw that it used `RepresentativeOfMinimalIdeal` and `EndomorphismMonoid`. Are these the only uses of Semigroups? I’ll have a think about whether these could be avoided so that this doesn’t require a semigroups.

MTWhyte:
Yeah, they’re the only uses is Semigroups as is. I’d thought about this, and I figured that since previous versions I’d made had drawn more from Semigroups that it’d be useful to keep it here, just so if anyone might find a way to make it faster using Semigroups then it isn’t too difficult.

wilfwilson:
The problem is that the Digraphs package doesn't require the Semigroups package in order for it to run, and it is reasonable to expect a user of the Digraphs package to not necessarily have Semigroups installed. Therefore, it seems suboptimal that this is in the Semigroups package rather than the Digraphs package, since it's very much a function about digraphs, and one would problem expect it to belong in the Digraphs package, and yet a user of Digraphs might not be able to get their hands on it.

This is essentially the code that is Semigroups-dependent:
```gap
if HasGeneratorsOfEndomorphismMonoidAttr(digraph) then
     rep := RepresentativeOfMinimalIdeal(EndomorphismMonoid(digraph));
     return ImageSetOfTransformation(rep, N);
fi;
```
I know that `RepresentativeOfMinimalIdeal` pretty much immediately throws away the semigroup that it's given, and just cares about the generators; therefore isn't a fundamental need to call `EndomorphismMonoid` (modulo the details). Removing the dependence on `RepresentativeOfMinimalIdeal` is more tricky. At its heart, it's digraph algorithm, and so maybe its main functionality could be moved into the Digraphs package... but let's not do that for now.

Note that this code is really just an optimisation: you're saying "if I know the endomorphisms of my digraph, I can just grab something from the minimal ideal". So maybe this special case should be in a new method. So I think I have a better solution: move all of this code to the Digraphs package, except for the lines I quoted above. Then, you can add an extra method to the Semigroups package, which gets called when the endomorphism monoid is known:

```gap
InstallMethod(DigraphCore, "for a digraph with generators of endomorphism monoid"
[IsDigraph and HasGeneratorsOfEndomorphismMonoid],
function(D)
  local x;
  x := RepresentativeOfMinimalIdeal(EndomorphismMonoid(digraph));
  return ImageSetOfTransformation(x, DigraphNrVertices);
end);
```
Therefore, with the Semigroups package loaded, the default `DigraphCore` can take advantage of this optimisation; when Semigroups is not available, `DigraphCore` still works reasonably well.

MTWhyte:
Thanks for the suggestion - I'll push the changes now.

I'm almost ready to push to Digraphs too, but I'm a bit stuck on what to do about the documentation.
I would be inclined to include `DigraphCore` in chapter 7 of the Digraphs documentation, but I'm not sure how to do this since there isn't a `z-chap7.xml` file. If you agree it should go there, which file should I edit to add to chapter 7?

wilfwilson:
Great. The homomorphism documentation chapter in Digraphs (i.e. Chapter 7) is in... `z-chap6.xml`.

wilfwilson:
I'll just note that this won't work, and shouldn't be merged, until the main `DigraphCore` stuff is available in a released version of the `Digraphs` package (presumably Digraphs 1.0), and when we require that version (which will presumably be as soon as it is released).

james-d-mitchell:
Thanks @wilfwilson and @MTWhyte for your work on this!

james-d-mitchell:
@MTWhyte any chance you could rebase this onto the `stable-3.2` branch. I think it can have the `do-not-merge` label removed then, right @wilfwilson ?

wilfwilson:
Yep, please rebase and then I'll review soon, and hopefully merge.

wilfwilson:
Actually @james-d-mitchell do you want it rebased on (and merged into) `stable-3.2` or into `master`? I'm not bothered, but I wondered if you'd prefer new features to be in `master` rather than a `stable` branch.

james-d-mitchell:
Yes @wilfwilson you're right, onto the `master` branch please!

PR #616, opened 2019-09-16
IsOperation implies IsFunction

fingolfin:


wilfwilson:
Thanks, nice simplification.

PR #618, opened 2019-09-18
Avoid using attributes Matrix, Rows, Columns

fingolfin:
and also their testers and setters. Instead use one of these:
- RowsOfReesZeroMatrixSemigroup
- RowsOfReesMatrixSemigroup
- ColumnsOfReesZeroMatrixSemigroup
- ColumnsOfReesMatrixSemigroup
- MatrixOfReesZeroMatrixSemigroup
- MatrixOfReesMatrixSemigroup

Right now, in GAP 4.11 and master branch, there are just synonyms for
the attributes they replace, but this may change in the future (namely
in order to ensure that Matrix on a single argument is a plain operation,
not an attribute, as for use as a matrix *constructor* function, it is not
desirable to have it an attribute).

My hope is that this can be merged soon. Then once a release of Semigroups with this change was made, we can change Matrix to not be an attribute anymore in the GAP master branch, for GAP 4.12 down the road; see https://github.com/gap-system/gap/pull/3664 for that.

CC @ThomasBreuer

### DIFF
--- a/gap/main/fropin.gi
+++ b/gap/main/fropin.gi
@@ -68,7 +68,8 @@ IsGeneratorsOfSemigroup);
 # Rees(Zero)MatrixSemigroup code.
 
 InstallImmediateMethod(IsGeneratorsOfEnumerableSemigroup,
-IsReesZeroMatrixSubsemigroup and HasRows and HasColumns, 0,
+IsReesZeroMatrixSubsemigroup and HasRowsOfReesZeroMatrixSemigroup
+    and HasColumnsOfReesZeroMatrixSemigroup, 0,
 function(R)
   return IsGeneratorsOfEnumerableSemigroup([Representative(R)]);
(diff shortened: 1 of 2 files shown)


wilfwilson:
So just to check my understanding: in the future (for example) `MatrixOfReesZeroMatrixSemigroup` would be the attribute in GAP, and there'd be a method installed for the operation `Matrix` with argument `IsReesZeroMatrixSemigroup` that delegates to this, i.e. we'd still be able to call `Matrix` on a Rees 0-matrix semigroup in the future?

fingolfin:
@wilfwilson yes; see https://github.com/gap-system/gap/pull/3664 for details

wilfwilson:
Ah great (I wrote my message before seeing https://github.com/gap-system/gap/pull/3664). Thanks!

PR #624, opened 2019-10-07
Don't use info argument for InstallGlobalFunction

fingolfin:
This third (middle) argument is undocumented, and in fact does nothing.

This PR has a commit branching from stable-3.1, so you can choose freely which branch the PR should target -- I pointed it at stable-3.2 for now.


fingolfin:
Yes we did it once before. But no worries. I totally see how it would slip in; it's also hard to detect; which is why I now made a PR for GAP itself which will print a warning when it finds similar instances. Then we can either "fix" all packages until GAP 4.12, and change the warning to an error; or just keep it as a warning forever. 

Anyway, this is really a minor thing :-). And as always, Semigroups maintainers are super nice to deal with, which I really appreciate!

Issue #623, opened 2019-10-06
Migrate OnPosIntSetsTrans from GAP to here?


fingolfin:
The GAP kernel implements an undocumented function `OnPosIntSetsTrans` which nothing in GAP itself uses, except some tests for it. The only place using it that we could find is in the Semigroups package.

Perhaps this function and its tests should be migrated to Semigroups then? Of course a little care has to be taken in order to get a smooth migration; I imagine that it would be implemented in the Semigroups kernel code under a different name, and then in the GAP code, there'd be something like `if not IsBound(OnPosIntSetsTrans) then BindGlobal("OnPosIntSetsTrans", WHATEVEr_THE_NEW_NAME_IS); fi;`. That way, 100% performance plus 100% compatibility with both old and new GAP versions.

james-d-mitchell:
@fingolfin I'm happy to do this. The code only exists in the GAP library because at some point in the distant past Semigroups had no kernel module, and this function is critical to the performance of some methods in Semigroups, and so it was agreed at the time to put it into the GAP kernel.

james-d-mitchell:
I will do this tomorrow afternoon unless someone beats me to it.

wilfwilson:
We've decided not to do this for now, see especially #625.

PR #625, opened 2019-10-09
Fix issue 623

james-d-mitchell:
This PR fixes Issue 623 by adding OnPosIntSetsTrans to the kernel module of the Semigroups package.

### DIFF
--- a/src/.clang-format
+++ b/src/.clang-format
@@ -81,10 +81,6 @@ PenaltyBreakString: 1000
 PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 600
 PointerAlignment: Left
-RawStringFormats: 
-  - Delimiter:       pb
-    Language:        TextProto
-    BasedOnStyle:    google
 ReflowComments:  true
 SortIncludes:    true
 SpaceAfterCStyleCast: true
--- a/src/pkg.cc
+++ b/src/pkg.cc
@@ -176,7 +176,9 @@ static Obj ON_POS_INT_SETS_TRANS(Obj self, Obj set, Obj f, Obj n) {
             "transformation (not a %s)",
             (Int) TNAM_OBJ(f),
             0L);
-  return 0L;
+  SORT_PLIST_CYC(res);
+  REMOVE_DUPS_PLIST_INTOBJ(res);
+  return res;
 }
 
 // Function to print a T_SEMI Obj.
--- a/src/pkg.cc
+++ b/src/pkg.cc
@@ -130,7 +130,7 @@ static Obj ON_POS_INT_SETS_TRANS(Obj self, Obj set, Obj f, Obj n) {
     return set;
   }
 
-  if (LEN_LIST(set) == 1 && INT_INTOBJ(ELM_LIST(set, 1)) == 0) {
+  if (LEN_LIST(set) == 1 && ELM_LIST(set, 1) == INTOBJ_INT(0)) {
     return FuncIMAGE_SET_TRANS_INT(self, f, n);
   }
 
--- a/src/pkg.cc
+++ b/src/pkg.cc
@@ -60,7 +60,7 @@ UInt T_BIPART = 0;
 UInt T_BLOCKS = 0;
 
 // Forward decl
-Obj FuncIMAGE_SET_TRANS_INT(Obj, Obj, Obj);
+extern Obj FuncIMAGE_SET_TRANS_INT(Obj, Obj, Obj);
 
 static inline Int IS_MUTABLE_PLIST(Obj list) {
   return !((TNUM_OBJ(list) - T_PLIST) % 2);
--- a/src/pkg.cc
+++ b/src/pkg.cc
@@ -24,6 +24,8 @@
 
 #include <iostream>
 
+#include "src/trans.h"
+
 #include "bipart.h"
 #include "congpairs.h"
 #include "converter.h"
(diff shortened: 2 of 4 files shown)


james-d-mitchell:
I need to add the tests from GAP to this PR too.

james-d-mitchell:
Might eventually resolve Issue #623. 

Genius idea to rename `IS_MUTABLE_PLIST` -> `IS_PLIST_MUTABLE`.

src/pkg.cc
```diff
@@ -57,6 +59,126 @@ UInt T_SEMI   = 0;
 UInt T_BIPART = 0;
 UInt T_BLOCKS = 0;
 
+// Forward decl
+Obj FuncIMAGE_SET_TRANS_INT(Obj, Obj, Obj);
+
+static inline Int IS_MUTABLE_PLIST(Obj list) {
+  return !((TNUM_OBJ(list) - T_PLIST) % 2);
+}
```
fingolfin: That function is a bad idea, it will break in future GAP versions, as encoding the (im)mutability in TNUMs has a good chance of going away at some point.

You can avoid this by conditionally compiling this code only when GAP is new enough, e.g using an `#if GAP_KERNEL_API_VERSION >= ...` statement. See also https://github.com/gap-system/gap/pull/3700 which I made just for that purpose :-)

src/pkg.cc
```diff
@@ -57,6 +59,126 @@ UInt T_SEMI   = 0;
...
+  const Obj*   ptset;
+  UInt         deg;
+  Obj *        ptres, res;
+  UInt         i, k;
+
+  if (LEN_LIST(set) == 0) {
+    return set;
+  }
+
+  if (LEN_LIST(set) == 1 && INT_INTOBJ(ELM_LIST(set, 1)) == 0) {
```
fingolfin:
```suggestion
  if (LEN_LIST(set) == 1 && ELM_LIST(set, 1) == INTOBJ_INT(0)) {
```
This is (a) more efficient and (b) slightly safer, as it'll work correctly even if the first element of the list is not an intobj.

Of course then the code later on would run into issues, but that's a different concern :-)

src/pkg.cc
```diff
@@ -57,6 +59,126 @@ UInt T_SEMI   = 0;
...
+    }
+    SORT_PLIST_CYC(res);
+    REMOVE_DUPS_PLIST_INTOBJ(res);
+    return res;
+  }
+  ErrorQuit("OnPosIntSetsTrans: the argument must be a "
+            "transformation (not a %s)",
+            (Int) TNAM_OBJ(f),
+            0L);
+  return 0L;
```
fingolfin: If you were to put the final `ErrorQuit` into an `else` block, then you could factor out some common code, i.e.:
```suggestion
  SORT_PLIST_CYC(res);
  REMOVE_DUPS_PLIST_INTOBJ(res);
  return res;
```

src/pkg.cc
```diff
@@ -57,6 +59,126 @@ UInt T_SEMI   = 0;
...
+          ADDR_OBJ(res)[k] = CONST_ADDR_OBJ(res)[k - h];
+          k -= h;
+        }
+        ADDR_OBJ(res)[k] = tmp;
+      }
+      h = h / 3;
+    }
+    CHANGED_BAG(res);
+  }
+}
```
fingolfin: If you check for the kernel version as suggested above, then you could also remove this function and just call `SortPlistByRawObj`

src/pkg.cc
```diff
@@ -57,6 +59,126 @@ UInt T_SEMI   = 0;
 UInt T_BIPART = 0;
 UInt T_BLOCKS = 0;
 
+// Forward decl
```
fingolfin: Maybe this?
```suggestion
// Forward decl of a function imported from the GAP kernel
```
I wonder how the code even loads with this in older GAP versions, though?!? Anyway, with the `#if` check for kernel versions, this would be a non-issue.

src/.clang-format
```diff
@@ -81,10 +81,6 @@ PenaltyBreakString: 1000
 PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 600
 PointerAlignment: Left
-RawStringFormats: 
-  - Delimiter:       pb
-    Language:        TextProto
-    BasedOnStyle:    google
```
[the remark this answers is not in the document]
james-d-mitchell: I mean I did this on purpose, in a separate commit, but it doesn't really fit with the rest of the PR I agree.

src/pkg.cc
james-d-mitchell: I know, I wasn't going to leave this in, I was just trying to make this work.

fingolfin: Ah OK, sorry then -- and: excellent :-). Of course it's important to first get it to work (and thanks for working at this at all, much appreciated)

james-d-mitchell:
Thanks for your comments @fingolfin, I should have removed the request for you to review until I got this to work, I somehow thought this would be easier than it turned out to be... I'll try to update the PR soon.

src/pkg.cc
james-d-mitchell: More efficient because `INTOBJ_INT(0)` is faster than `INT_INTOBJ(ELM_LIST(set, 1))`, or why? Not important, but I'd like to understand why...

fingolfin: `INTOBJ_INT(0)` is a compile time constant, so the compiler can turn the comparision into a single machine instruction. Whereas for the old code, it has to first shift the value returned by ELM_LIST, and then compare (so two instructions). Not that this really matters much in practice ;-).

fingolfin:
@james-d-mitchell please take your time

src/pkg.cc
james-d-mitchell: Makes sense, thanks for the explanation!

src/pkg.cc
james-d-mitchell: Right you are, deleted!

james-d-mitchell:
Ok, I've spent another afternoon trying to figure out how to do this, and not break backwards compatibility for Semigroups, and I'm ready to give up. 

The problem is to make `OnPosIntSetsTrans` work in GAP 4.9 we have to take large parts of `trans.c[c]` and put them into the Semigroups kernel module. This doesn't seem worth it to me. 

We could stop supporting older versions of GAP, but doing that just to move this one function from the GAP kernel doesn't seem worth it to me. Another solution would be to remember that we want to do this and do it the next time it makes sense to update the version of GAP required by Semigroups. 

If you have a better suggestion @fingolfin, I'm happy to hear it.

fingolfin:
@james-d-mitchell I am sorry that you wasted so much time on this :-(. Please don't bother doing any more on this. I'll take a look if I can see an "easy" (whatever that means) way to pull this off, and then I'll submit a PR; but otherwise, it really won't hurt us to leave `OnPosIntSetsTrans` in GAP itself.

Issue #632, opened 2019-12-15
Failures in GAP master branch


alex-konovalov:
https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master/jobs/624910497 shows this:
```
testing: /home/gap/inst/gap-master/pkg/semigroups-3.2.0/tst/standard/attr.tst
######## > Diff in:
/home/gap/inst/gap-master/pkg/semigroups-3.2.0/tst/standard/attr.tst:1210
# Input is:
S := Monoid(Matrix(GF(2 ^ 2),
[[Z(2 ^ 2), 0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2)],
 [Z(2 ^ 2), 0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2)],
 [0 * Z(2), Z(2) ^ 0, 0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2)],
 [0 * Z(2), 0 * Z(2), Z(2) ^ 0, 0 * Z(2), 0 * Z(2), 0 * Z(2)],
 [0 * Z(2), 0 * Z(2), 0 * Z(2), Z(2) ^ 0, 0 * Z(2), 0 * Z(2)],
 [0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2), 0 * Z(2)]]));;
# Expected output:
# But found:
Error, no method found! For debugging hints type ?Recovery from NoMethodFound
Error, no 2nd choice method found for `OneMutable' on 1 arguments
########
```

wilfwilson:
Thanks. Before looking into this further, I’ll note that Semigroups 3.2.1 is the most recent version (not 3.2.0, as you are testing there), so it’s possible that the problem is fixed already. If the problems still exist with Semigroups 3.2.1, then we’ll definitely look into it properly 🙂

alex-konovalov:
Ok - a new test is now running at https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master/builds/625265651, it has Semigroups 3.2.1. Let's watch it.

alex-konovalov:
@wilfwilson ok now - sorry for the false alarm: https://travis-ci.org/gap-infra/gap-docker-pkg-tests-master/jobs/625265734

alex-konovalov:
Still something is wrong here:
[...]

PR #633, opened 2019-12-23
io: close files when erroring

james-d-mitchell:
This is a partial fix for Issue #632, I think that this will resolve the broken test in:

https://travis-ci.org/gap-infra/gap-docker-pkg-tests-stable-4.11/jobs/627234005

Which I think was caused by several functions in `gap/tools/io.gi` failing to close files that they open when an error was given. This PR resolves this issue by closing such files.

### DIFF
--- a/gap/tools/io.gi
+++ b/gap/tools/io.gi
@@ -260,6 +260,9 @@ function(arg)
   fi;
 
   if not IsList(collcoll) or IsEmpty(collcoll) then
+    if IsString(name) then
+      IO_Close(file);
+    fi;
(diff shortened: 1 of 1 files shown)


wilfwilson:
Quick Q @james-d-mitchell : do you think Digraphs has similar issues?

james-d-mitchell:
@wilfwilson on quick inspection Digraphs has the same issue as this, I detected this by running `DigraphsTestStandard` and then doing `IO.OpenFiles`, which ought to be an empty list but isn't. I didn't look into it further just yet.

PR #635, opened 2020-01-01
Make sure to include GAP headers before libsemigroups

fingolfin:
This way, #defines in libsemigroups don't affect GAP headers. For example, GAP may typedef BOOL in the future, and right now that would break Semigroups because libsemigroups #defines BOOL to something. With this patch, that problem goes away.

It also seems in general useful to have a fixed order; this way, *if* things break in either GAP or libsemigroups due to a change in the respective other, at least they will break consistently.
Motivated by https://github.com/gap-system/gap/pull/3814

### DIFF
--- a/src/bipart.cc
+++ b/src/bipart.cc
@@ -28,10 +28,10 @@
 #include <utility>
 #include <vector>
 
+#include "src/compiled.h"
+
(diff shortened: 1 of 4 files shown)


james-d-mitchell:
Thanks for this @fingolfin, happy to merge

PR #641, opened 2020-01-20
Accept both libsemigroups before and after PR 172: Includes in libsem…

nthiery:
…igroups/


james-d-mitchell:
Thanks for this @nthiery

PR #645, opened 2020-02-07
kernel: fix compile warnings and leak

james-d-mitchell:


src/semigrp.cc
```diff
@@ -42,13 +42,13 @@ using libsemigroups::UNDEFINED;
...
 #ifdef SEMIGROUPS_KERNEL_DEBUG
-#define ERROR(obj, message)                               \
-  char buf[128];                                          \
-  strncpy(buf, __func__, sizeof(buf));                    \
-  strncat(buf, ": ", (sizeof(buf) * 2));                  \
-  strncat(buf, message, (sizeof(buf) * strlen(message))); \
-  strncat(buf, " not a %s,", (sizeof(buf) * 9));          \
-  ErrorQuit(buf, (Int) TNAM_OBJ(obj), 0L);
+#define ERROR(obj, message) \
+  static std::string buf;   \
```
fingolfin: For various reasons I just stumbled over this PR, and so here's some unasked for feedback; only good intentions here, though! And of course, feel free to ignore me!

But I was wondering: Any particular reason to make this `static`? This is not thread safe (i.e., will hurt HPC-GAP). Also, you are adding a lot of static global variables this way (in each place calling `ERROR`, which means every place using one of your `CHECK_*` macros). You'll get a smaller binary and smaller code by leaving it on the stack (and this is an exceptional code path anyway, so in the end its performance and memory usage don't matter.

Actually, I would recommend to turn `ERROR` into an actual function. This way, the calling code (the one containing the checks) gets smaller (and so is more likely to fit into the cache), and the code gets easier to edit (no dealing with line continuation). The only thing "preventing" this is the use of `__func__`, but that has a simply solution: Just do this:
```C
#define ERROR(obj, message)  ERROR_INTERN(obj, message, __func__)
static void ERROR_INTERN(Obj obj, const char * message, const char * func)
{
 ....
}
```

src/semigrp.cc
```diff
@@ -42,13 +42,13 @@ using libsemigroups::UNDEFINED;
...
+  buf += __func__;          \
+  buf += ": ";              \
+  buf += message;           \
+  buf += " not a %s,";      \
```
fingolfin: Why does the error message end with a trailing comma?

Also, since you are creating buf here anyway, you could also replace this:
```suggestion
  buf += " not a ";      \
  buf += TNAM_OBJ(obj);      \
  buf += ",";      \
```
and then don't pass TNAM_OBJ to ErrorQuit

james-d-mitchell: Fair enough, the trailing comma is standard in the Semigroups package error messages, to indicate where the error message ends and the stuff output by GAP begins, if you see what I mean. This might not any longer be necessary/desirable, in the (ever more distant past) error messages in GAP were not very good/accurate and so we used a particular format in the error message strings as a hack to make up for the shortcomings.

src/semigrp.cc
james-d-mitchell: Thanks for the suggestion, I'm happy to implement this change. The variable is static just to avoid allocating and deallocating the required memory repeatedly, I'm aware that this isn't thread safe, but so far that hasn't been an issue. FYI I made this change because of some compiler warnings that the old code emitted.

src/semigrp.cc
james-d-mitchell: PS: the reason for this PR is that the travis tests are failing. So far, I've been unable to figure out what is causing these failures, and I also haven't been able to reproduce them locally.

Issue #646, opened 2020-02-19
Option to remove -march=native even if it is supported


isuruf:
For binary distributions like conda, we need a way to avoid build machine specific instructions.

james-d-mitchell:
Resolved in 3.2.4

PR #651, opened 2020-02-28
build: add --without-march-native flag

james-d-mitchell:


james-d-mitchell:
@isuruf does this resolve issue #646 ?

isuruf:
Yes, it does. Thanks.

PR #653, opened 2020-03-02
Fix bug in IsRowTrim / IsColTrim for boolean matrices

flsmith:

PR #654, opened 2020-03-04
Fix bug in IsRowTrim / IsColTrim for boolean matrices

flsmith:
Fix #653

### DIFF
--- a/gap/elements/boolmat.gi
+++ b/gap/elements/boolmat.gi
@@ -251,7 +251,8 @@ function(x)
   n := Length(x![1]);
   for i in [1 .. n - 1] do
     for j in [i + 1 .. n] do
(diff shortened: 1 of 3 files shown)


gap/elements/boolmat.gi
```diff
@@ -265,8 +266,8 @@ function(x)
   local n, contained, row, i, j, k;
 
   n := Length(x![1]);
-  for i in [1 .. n - 1] do
-    for j in [i + 1 .. n] do
+  for i in [1 .. n] do
+    for j in Concatenation([1 .. i - 1], [i + 1 .. n]) do
```
james-d-mitchell: I don't know if this'll make much difference, but might it not be better to just check if j = i inside the loop here?

PR #659, opened 2020-04-03
Fix header includes

ChrisJefferson:
This PR fixes the root directory of semigroups getting included in the include path.

The first commit fixes two included headers -- technically these were always wrong, but worked because lots of extra things were in the include path. The second commit stops extra include paths appearing.

Now you have renamed VERSION anyway, this might not make things any better, but it is (I hope) "neater" in the long term. it would also be nice to see if it breaks any semigroups tests, to see if this is the right idea for a fix.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,7 +23,7 @@ AC_CANONICAL_HOST
 dnl ##
 dnl ## Setup automake
 dnl ##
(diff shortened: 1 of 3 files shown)


james-d-mitchell:
Thanks @ChrisJefferson very helpful!

PR #663, opened 2020-04-19
Change #include "src/compiled.h" to just "compiled.h"

fingolfin:
This will allow future GAP versions to simplify their GAP_CPPFLAGS,
which currently need to contain `-I$(abs_srcdir)` just to support this
old use case. Also, this in turn makes it easier to have package
compatible with future installed version of GAP, which have headers in
PREFIX/include/gap/ so there is no `src` there (unless downstream
packagers jump through hoops to ensure that, e.g. by adding a special
symlink).

Note: we could also #include "gap_all.h" but that would require GAP >= 4.11.


james-d-mitchell:
Thanks @fingolfin, in principle I'm happy to accept this, except that it breaks backwards compatibility (as can be seen in the failing travis-ci tests). If this could be updated to remain compatible with older versions of GAP, then I'd be happy to merge.

PR #683, opened 2020-07-01
Add EqualInFreeBand method

MTWhyte:
This PR adds an operation `EqualInFreeBand`, written by @tomcontileslie, @reiniscirpons and myself, which is an implementation of the algorithm described in Jakub Radoszewski and Wojciech Rytter's _Efficient Testing of Equivalence of Words in a Free Idempotent Semigroup⋆_.

https://link.springer.com/chapter/10.1007/978-3-642-11266-9_55

PR #661, opened 2020-04-08
Added function to write words as strings

mariatsalakou:

PR #686, opened 2020-07-03
Add StandardiseWord and StringToWord methods

MTWhyte:
This PR adds two more string and word related functions to the `word` family.

`StandardiseWord` takes a word, consisting of `n` distinct positive integers, and changes it in-place to a word over the alphabet `[1 .. n]`, where the characters of the new word correspond to the characters of the input, in order of first appearance.

`StringToWord` takes a string `s`, consisting of `n` distinct characters, and returns a word over `[1 .. n]` whose characters correspond to the characters of `s` in order of first appearance.

### DIFF
--- a/doc/word.xml
+++ b/doc/word.xml
@@ -44,3 +44,60 @@ gap> RandomWord(8, 4);
 </Description> 
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="StandardiseWord" >
+<ManSection>
+  <Oper Name = "StandardiseWord" Arg = "w"/>
+  <Oper Name = "StandardizeWord" Arg = "w"/>
+  <Returns>
+    A list of positive integers.
+  </Returns>
+  <Description>
+    This function takes a word <A>w</A>, consisting of <C>n</C> distinct
+    positive integers and returns a word <C>s</C> where the characters of
+    <C>s</C> correspond to those of <A>w</A> in order of first appearence.<P/>
+
+    The word <A>w</A> is changed in-place into word <C>s</C>.
+  <Example><![CDATA[
+gap> w := [3, 1, 2];
+[ 3, 1, 2 ]
+gap> StandardiseWord(w);
+[ 1, 2, 3 ]
+gap> w;
+[ 1, 2, 3 ]
+gap> w := [4, 2, 10, 2];
+[ 4, 2, 10, 2 ]
+gap> StandardiseWord(w);
+[ 1, 2, 3, 2 ]]]></Example> 
+</Description> 
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="StringToWord" >
+<ManSection>
+  <Oper Name = "StringToWord" Arg = "s"
+    Label="for a string"/>
+  <Returns>
+    A list of positive integers.
+  </Returns>
+  <Description>
+    This function takes a string <A>s</A>, consisting of <C>n</C> distinct
+    positive integers and returns a word <C>w</C> (i.e. a list of positive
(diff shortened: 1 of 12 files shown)


MTWhyte:
As far as the relationship of this branch goes with others, I first branched off of #661 and then rebased on to #683, and the commits from these appear in this PR. I have tried to split up my changes commit-wise as much as possible, since I think some of them are likely to clash with future changes to #661.

MTWhyte:
Since it contains all the relevant changes, this PR supersedes #661.

doc/word.xml
```diff
@@ -0,0 +1,104 @@
...
+
+<#GAPDoc Label="WordToString" >
+<ManSection>
+  <Oper Name = "WordToString" Arg = "alphabet, word"
+    Label="for a string and a list"/>
+  <Returns>
+    The <A>word</A> in the form of a string.
+  </Returns>
+  <Description>
+    Returns the word <A>word</A> that is given in the form of a list, in
```
tomcontileslie, line 18: I would recommend adding a comment here on the fact that `alphabet` has to be a string. I don't think the `Label` that you've added works for all platforms; it doesn't seem to appear anywhere in the HTML documentation!

doc/word.xml
```diff
@@ -0,0 +1,104 @@
...
+ </ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="RandomWord" >
+<ManSection>
+  <Oper Name = "RandomWord" Arg = "length, number_letters"
+    Label="for two integers"/>
+  <Returns>
+    A random word of length <A>length</A> in the form of a list using 
+    <A>number_letters</A> letters.
```
tomcontileslie, line 33: I don't know what you think about this, but I think it's easier to read if you give shorter names to the arguments?
```suggestion
  <Oper Name = "RandomWord" Arg = "length, n"
    Label="for two integers"/>
  <Returns>
    A random word of length <A>length</A> in the form of a list using 
    <A>n</A> letters.
```

doc/word.xml
```diff
@@ -0,0 +1,104 @@
...
+  </Returns>
+  <Description>
+	  Returns a random word of length <A>length</A>
```
tomcontileslie, line 36: This line effectively repeats the contents of `<Returns>`. Maybe you could either move some of the stuff you wrote  in the "returns" section to the description section, or you could add some additional comments in the description regarding how the random word is generated, just to avoid repetition?

doc/z-chap10.xml
```diff
@@ -179,4 +180,15 @@ x1x2x2^-1x1^-1x1x2]]></Example>
 
     <#Include Label = "GreensDClassOfElement" >
   </Section>
+
+
+  <Section>
+    <Heading> Words and strings in finitely presented semigroups
+    </Heading>
+
```
tomcontileslie, line 188: Do you think some explanation of what GAP "words" are and how they are used could be added here, for context? E.g. specify that by a "word", you mean a list of pos ints, and that there are methods for converting arbitrary words to "nice" GAP lists.

tomcontileslie (commented):
I haven't commented on the EqualInFreeBand stuff because I understand this PR should still be merged after #683, but I've added a couple comments for clarity in the documentation - otherwise all looks good!

MTWhyte:
@tomcontileslie thanks for the suggestions! I've just pushed all of these now.

I'm not actually sure what the `<Label>` tag does -- I also can't see it come up anywhere -- but it seems ubiquitous in the xml files so I'll keep it in. Perhaps @james-d-mitchell could explain it?

james-d-mitchell:
I think the changes in this PR are also contained in PR #752 (merged), but maybe some of your comments still apply @MTWhyte  and @tomcontileslie?

PR #687, opened 2020-07-06
method for small generating set 

Peter-Ing:
…trong orientation and data strucutres

	modified:   doc/semitrans.xml documentation for small gnereating set
	modified:   gap/semigroups/semitrans.gi declared a new method for the small generating set function for singular closures of transformation and partial transformation monoids
	modified:   tst/standard/semitrans.tst added test for each 4 cases of function


james-d-mitchell:
I tried in 2022 to fix this PR up so that it could be merged, but basically gave up. I don't foresee having the time to do this in the near future, and it might be quicker to do it again from scratch since the original author is no longer available,.

PR #684, opened 2020-07-02
Add ToddCoxeterBand method

MTWhyte:
This PR adds a band-specific Todd-Coxeter algorithm, written by @tomcontileslie, @reiniscirpons and myself. This is still a work in progress, until we are more certain that the algorithm is proper.

I plan on adding tests and documentation examples once we have decided what the output ought to be.

PR #691, opened 2020-07-09
Add ToddCoxeterBand method

tomcontileslie:
**Edit: for an update, see enormous comment below**

This PR builds on @MTWhyte's PR #684 adding a ToddCoxeterBand method, also joint work with @reiniscirpons under @james-d-mitchell. This version is consistent with the properness proof we have written, the changes from 684 being:

- `new_coset` now checks pre-existence of words via paths (i.e. checks value of `tau(i, canon(wa))` rather than `canon(wa) in words`) 
- Double the implicit relations are now pushed: when working on word n, active cosets in {1, ..., n-1} are pushed through `words[n]words[n] = words[n]`.
- The `canon` function now ensures the canonical form is shortlex-minimal.

This is WIP until output format is decided and tests and docs are written. This function should eventually be implemented in libsemigroups, preferably with an analogous Felsch implementation (the strategy here is broadly HLT).

### DIFF
--- a/doc/freeband.xml
+++ b/doc/freeband.xml
@@ -190,3 +190,23 @@ gap> ContentOfFreeBandElementCollection([x, y]);
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="ToddCoxeterBand">
+  <ManSection>
+    <Oper Name="ToddCoxeterBand"  Arg="n, R"/>
+    <Description>
+      This operation takes band presentation, where <A>n</A> is the size
+      of alphabet <C>A = [1 .. n]</C> and <A>R</A> is a list of lists of
+      words over <C>A</C>, representing the relations. It computes the
+      band defined by this band presentation via a band-specific version
+      of the Todd-Coxeter algorithm. If <A>R</A> is the empty list, then
+      the free band over <C>A</C> is computed.
+      
+      <!--
+      TODO: I think we should write a little more about the algorithm here
+      once we've finalised it. We should also add some examples, once we've
+      decided on exactly what the output should be.
+      -->
+    </Description>
+  </ManSection>
+<#/GAPDoc>
--- a/doc/z-chap10.xml
+++ b/doc/z-chap10.xml
@@ -154,6 +154,7 @@ x1x2x2^-1x1^-1x1x2]]></Example>
     <#Include Label = "IsFreeBandElementCollection">
     <#Include Label = "IsFreeBandSubsemigroup">
     <#Include Label = "ContentOfFreeBandElement">
+    <#Include Label = "ToddCoxeterBand">
     
   </Section>
 
--- a/gap/fp/freeband.gd
+++ b/gap/fp/freeband.gd
@@ -18,3 +18,5 @@ DeclareGlobalFunction("FreeBand");
 DeclareAttribute("ContentOfFreeBandElement", IsFreeBandElement);
 DeclareAttribute("ContentOfFreeBandElementCollection",
                  IsFreeBandElementCollection);
+
+DeclareOperation("ToddCoxeterBand", [IsPosInt, IsList]);
--- a/gap/fp/freeband.gi
+++ b/gap/fp/freeband.gi
@@ -550,3 +550,186 @@ function(x, hashlen)
   return rec(func := SEMIGROUPS.HashFunctionForFreeBandElements,
              data := hashlen);
 end);
+
+InstallMethod(ToddCoxeterBand, "for a pos int and list of lists of words",
+[IsPosInt, IsList],
+function(N, R)
+  local new_coset, tauf, canon, push_relation, process_coincidences,
+  A, F, G, k, active_cosets, table, coincidences, words, n, word,
+  pair, char, coset;
+
+  for pair in R do
+    if not IsList(pair) then
+      ErrorNoReturn("expected a list of lists as second argument");
+    elif Length(pair) <> 2 then
+      ErrorNoReturn("expected a list of lists with length 2 as second argument");
+    fi;
+    for word in pair do
+      for n in word do
+        if n < 1 or n > N then
+          ErrorNoReturn("expected a list of lists containing 2 words over ",
+                        "[1 .. N] as second argument");
+        fi;
+      od;
+    od;
+  od;
+
+  new_coset := function(coset, char)
+    local new_word, pos;
+    # new_coset for bands is smart. If the word created, once reduced,
+    # is already somewhere else in the list, then it just sets
+    # table[coset][char] to be that coset.
+    if table[coset][char] = 0 then
+      new_word := canon(Concatenation(words[coset], [char]));
+      pos      := Position(words, new_word);
+
+      if pos = fail then
+        # in this case the word is genuinely new and we make a new coset
+        table[coset][char] := k;
+        active_cosets[k]   := true;
+        Add(table, ListWithIdenticalEntries(Length(A), 0));
+        Add(words, new_word);
+        k := k + 1;
+
+      else
+        # word already exists
+        table[coset][char] := pos;
+
+      fi;
+    fi;
+  end;
+
+  tauf := function(coset, word)
+    local char;
+    # forced tau. This creates new cosets as necessary.
+    if Length(word) = 0 then
+      return coset;
+    fi;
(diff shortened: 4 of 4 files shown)


tomcontileslie:
## Update on the status of this pull request

File which I reference later: [Band-Todd-Coxeter.pdf](https://github.com/semigroups/Semigroups/files/6585465/Band-Todd-Coxeter.pdf)

Most of the work here was done a while ago, so memory is a bit rusty, but the stuff here is still relevant although it needs a bit more work before being justified. I'm just going to write out what work led us here, and I'll explain what I'm still unsure about.

The code written in this PR is a first attempt at answering the question:
> Given a FINITE band S given by a band presentation <A,R> (i.e. a list of generators and a list of relations, where it is implicitly assumed that all words are idempotent), can we devise an algorithm which will output some description of this band S - a coset table, Cayley digraph, etc...?

For general presentations (not band presentations), we can use the Todd-Coxeter (TC) algorithm to answer this question. The rigorous backing for this that I have familiarised myself with is the work of @james-d-mitchell, @flsmith, @mariatsalakou and Tom Coleman, which describes coset tables as labelled digraphs (called "R-digraphs"). The node set of an R-digraph is a subset of A* (these are the "cosets" of coset tables), its edges are labelled by elements of A, and the restriction is that if two paths over the graph lead to the same node, then the elements given by those paths must be equal in the presentation. They explain that any TC-style algorithm - referred to as a _congruence enumeration process_ - can be implemented as a combination of three basic digraph manipulations, starting with the digraph at step i, G(i), and outputting a new digraph at step i+1, G(i+1). The algorithm also keeps track of a set of _coincidences_ K(i), a set of pairs of words which are known to be equal in the presentation, but currently give different paths in G(i). (the idea is that eventually in the algorithm, any coincidences should be eliminated so that the R-digraph is consistent with the presentation is is computing).

The three steps are, roughly:
- **(TC1: add a node).** If a node _w_ has no out-edge _a_, define a new node _wa_ and an arrow labelled _a_ from _w_ to _wa_.
- **(TC2: push a relation)** For a node _w_ and a relation _(u,v) in R_, require that following paths _u_ and _v_ from _w_ lead to the same node; adding coincidences if not.
- **(TC3: process coincidences)** Clear the set _K_ by merging any nodes in the coincidence set and appropriately redefining paths through the digraph.

With some additional assumptions, the idea is that congruence enumeration processes for finite semigroups always terminate on a description of the semigroup. Our idea with @reiniscirpons and @MTWhyte was then to adapt this to work for bands too.

How do we do this? If we're given an alphabet _A_ of size _n_ and a set of relations _R_, we want to compute the smallest band which satisfies the relations _R_. The first thing we could try is to add "the band relations" to _R_, and then run the usual Todd-Coxeter algorithm on _A_ and _R_. If, say, we have _A = {x,y}_ and _R = {xy=yx}_, then the "band relations" (a list of relations which give a presentation for the free band on _A_) are _{xx=x, yy=y, xyxy=xy, yxyx=yx}_. So, running TC on _<A, {xx=x, yy=y, xyxy=xy, yxyx=yx, xy=yx}>_ will give the result we're looking for. However, for _n_ equal to 4 and above, the list of band relations becomes incredibly long. The idea is then to see whether we can tweak the TC algorithm so that we don't have to _explicitly_ pass the band relations, but still get the right result.

My first idea (see section 2.1 of attached doc) was to change the definition of TC1 so that it labels nodes in the free band canonical form. This can actually be generalised to work in any variety of semigroups, so let V be any variety (with the aim of later taking V = variety of bands). Suppose we have a function which, for any word _w_ in _A*_, returns the shortlex-smallest word in _A*_ which is equal to _w_ in (the free object of) the variety _V_ (call this its _canonical form_ and denote it _bar(w)_). Then our new TC1 (call it TC1V) does something like:
- **(TC1V: cleverly add a node in the variety _V_).** If a node _w_ (in canonical form) has no out-edge _a_, then check whether there is a node _bar(wa)_. If there is, draw an edge from _w_ to _bar(wa)_ labelled _a_. If not, define it (and all of its (also canonical) prefixes - some technicalities here) and draw an edge from _w_ to the newly defined _bar(wa)_.

This was the first thing we naturally tried to do when devising a band Todd-Coxeter (BTC) algorithm: only label cosets using some canonical form. In the attached document I prove that this is valid. More specifically:

**Result from Section 2.1.** Suppose we have a presentation _<A,R>_ implicitly given in a variety _V_. Let _R'_ be the union of the explicit relations _R_ with the set of all implicit relations generating the free object in the variety _V_ (remember, as mentioned before, this set of relations may be significantly larger than _R_, so we want to avoid computing _R'_ and/or passing it as input to the Todd-Coxeter algorithm). Then, any sequence of TC1V, TC2 and TC3 has exactly the same properties as a congruence enumeration process for R': namely, these three steps output R'-digraphs when given R'-digraphs, and any sequence of applications eventually stabilises.

The next step in the original paper is to devise specific combinations of TC1, TC2 and TC3 (_algorithms_ or _strategies_) which not only stabilise, but stabilise exactly at a description of the semigroup. It is shown that this always happens provided the algorithm satisfies three conditions, in which case we call it a _proper_ congruence enumeration process. These conditions essentially require that everything happens as you would hope; they roughly say:
1. Every path in the digraph is eventually defined.
2. Every possible coincidence is eventually noticed.
3. Coincidences are regularly processed.

Switching now to the band-specific setting (variety _V=B_), we have a canonical labelling function for free band elements and a clever band step TC1B from above. I have managed to show that a similar algorithm to the well-known HLT strategy satisfies properness conditions 1 and 3, but the issue I am having is that I am not sure whether it also satisfies properness condition 2.

The reason for this is as follows. We have a presentation _<A,R>_, and a larger presentation _R'_ which contains all of _R_ as well as all the free band relations. The HLT approach guarantees that coincidences involving pairs in _R_ are noticed for every coset, since these relations are passed explicitly. However, I am not sure this necessarily means all the implicit relations are also noticed (since they are never explicitly pushed through the cosets). This can be reduced to the following problem: given a canonical node _w_ and a free band element _u_ in canonical form, can we show that eventually either _w_ is not a node or _e(w,uu) = e(w,u)_?

I don't think we ever found a counterexample - but I haven't managed to show that the implicit relations will always end up being somehow implicitly pushed so as to satisfy condition 2. So, in section 2.2 of the attached document, I detail a modified HLT strategy where, as words _v_ are added to the node set, we also push the relation _vv=v_ quite a few times through all the nodes. This is a shoddy fix because it means that actually, we end up explicitly treating band relations which we were hoping we could keep implicit. However, this is still better than explicitly passing all the band relations from the outset. At the end of the attached document I seem to have been close to proving that with these additional relations gradually becoming explicit, the process is indeed proper (I did this last summer so I'm not sure how far we are from a full proof). This algorithm reflects the state of the document, in that we have a coset creation function which cleverly creates new table entries, with the relation pusher and coincidence processor functions relatively unchanged. The actual loop that implements the strategy currently contains, as explained, some extra steps which push band relations.

Future work on this PR (likely not done by me since I am graduating and therefore my brain will be in decline) should aim to either prove that this shoddy approach is valid, then implement it in GAP and then C++ code - or, ideally, prove that the strategy is valid even if we never push implicit relations, and then implement that. There is also a Felsch analogue to be developed. Good luck!

### Description of functions implemented in this PR

- `SEMIGROUPS.PrefixTupleOfFreeBandElement(word, n)`: returns the largest prefix of `word` with content size `n-1`.
- `SEMIGROUPS.ShortCanonicalFormOfFreeBandElement(word)`: returns the shortlex least representative of `word` in the free band (does this work? I'm not sure I ever proved it; @reiniscirpons is the person to ask). The strategy is to recursively take the canonical form of the prefix and suffix (in Howie's words, the _initial_ and _terminal_) and then see whether they can overlap.

Functions inside the main feature, `ToddCoxeterBand`:
- `new_coset`: carries out TC1B
- `tau(coset, word)`: returns, in JDM/FLS/MT/TC notation, the value _e(coset, word)_
- `tauf(coset, word)`: computes _e(coset, word)_, and if edges are undefined, it defines them using `new_coset`

Then as mentioned, `push_relation` and `process_coincidences` do exactly what you'd expect. The only things that may need tweaking are `new_coset`, and then the last 50 or so lines of `ToddCoxeterBand` which currently push a ridiculous number of cosets in an effort to fix the issues in the proof.

PR #694, opened 2020-07-29
Fix fp semi negative size

james-d-mitchell:


src/congpairs.cc
```diff
@@ -210,7 +210,12 @@ static Congruence* cong_obj_get_cpp(gap_cong_t cong) {
 }
 
 Obj CONG_PAIRS_NR_CLASSES(Obj self, gap_cong_t o) {
-  return INTOBJ_INT(cong_obj_get_cpp(o)->nr_classes());
+  auto n = cong_obj_get_cpp(o)->nr_classes();
+  if (n != libsemigroups::POSITIVE_INFINITY) {
```
flsmith, line 214: Does this pattern not occur anywhere else? Seems like there should be other instances, warranting a function which has this logic.

james-d-mitchell: Probably,  I'll double-check

PR #695, opened 2020-07-29
Small generating set partition monoid

Peter-Ing:
…rong orientation

	modified:   doc/semitrans.xml adds documentation for small genertaing set partition monoid
	modified:   gap/semigroups/semitrans.gi installs addition method for small generating set for partition monoid
	modified:   scripts/travis-build-dependencies.sh ensures data strucutres is there so test pass
	modified:   tst/standard/semitrans.tst added test for small generating set partition monoid

### DIFF
--- a/doc/semitrans.xml
+++ b/doc/semitrans.xml
@@ -376,3 +376,19 @@ gap> Size(WW);
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="SmallGeneratingSet">
+   <ManSection>
+    <Oper Name="SmallGeneratingSet" Arg="G"/>
+    <Returns>A small generating set with maximal idempotent elements.</Returns>
+    <Description>
+      When <A>G</A> is a subgroup of the Partition Monoid on <A>n</A> points. 
+	  of the form <A>N</A> union </A>SingularPartitionMonoid(n)</A> where <A>N</A>
+	  is a subgroup of the symmetric group on <A>n</A> points then this
+	  returns the small generating set which contains an equal number of 
+	  idempotent partitions as the idempotent rank of G. This set would be 
+	  minimal if <A>SmallGeneratingSet(N)</A> is of size equal to the rank of
+	  <A>N</A>.
+    </Description>
+  </ManSection>
+<#/GAPDoc>
--- a/doc/semitrans.xml
+++ b/doc/semitrans.xml
@@ -377,17 +377,17 @@ gap> Size(WW);
   </ManSection>
 <#/GAPDoc>
 
-<#GAPDoc Label="SmallGeneratingSet">
+<#GAPDoc Label="PartitionMonoidSmallGeneratingSet">
    <ManSection>
     <Oper Name="SmallGeneratingSet" Arg="G"/>
     <Returns>A small generating set with maximal idempotent elements.</Returns>
     <Description>
       When <A>G</A> is a subgroup of the Partition Monoid on <A>n</A> points. 
-	  of the form <A>N</A> union <A>SingularPartitionMonoid(n)</A> where <A>N</A>
+	  of the form <A>N</A> union <C>SingularPartitionMonoid(n)</C> where <A>N</A>
 	  is a subgroup of the symmetric group on <A>n</A> points then this
 	  returns the small generating set which contains an equal number of 
 	  idempotent partitions as the idempotent rank of G. This set would be 
-	  minimal if <A>SmallGeneratingSet(N)</A> is of size equal to the rank of
+	  minimal if <C>SmallGeneratingSet(N)</C> is of size equal to the rank of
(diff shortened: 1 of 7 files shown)


doc/semitrans.xml
```diff
@@ -376,3 +376,19 @@ gap> Size(WW);
...
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="SmallGeneratingSet">
+   <ManSection>
+    <Oper Name="SmallGeneratingSet" Arg="G"/>
+    <Returns>A small generating set with maximal idempotent elements.</Returns>
+    <Description>
+      When <A>G</A> is a subgroup of the Partition Monoid on <A>n</A> points. 
+	  of the form <A>N</A> union <A>SingularPartitionMonoid(n)</A> where <A>N</A>
```
tomcontileslie, line 386: Use `<A>...</A>` for function arguments, and `<C>...</C>` for code such as `SingularPartitionMonoid(n)`.

gap/semigroups/semitrans.gi
```diff
@@ -691,6 +691,161 @@ function(M, S)
...
+  # special case for if quoteient is a point
+  e := DigraphNrEdges(O);
+  N := [1 .. (2 * e) + g];
+  if v = 1 then
+    for i in [1 .. DigraphNrEdges(O)] do
+      K := o[i][1][1];
+      L := o[i][1][2];
+      N[2 * i] := AsBipartition(Transformation([K], [L]), n);
+    od;
+    # cases are if only 1 non loop and otherweise
```
tomcontileslie, line 756:
```suggestion
    # cases are if only 1 non loop and otherwise
```

gap/semigroups/semitrans.gi
```diff
@@ -691,6 +691,161 @@ function(M, S)
...
+        L := o[i][1][2];
+      elif DigraphEdges(O)[i][1] in DigraphEdges(Digo) then
+        K := o[i][1][1];
+        L := o[i][1][2];
+      else
+        K := o[i][1][2];
+        L := o[i][1][1];
+      fi;
+      # using these start and end point values
+      # construct a transfomarion which is constant on that edge
```
tomcontileslie, line 810:
```suggestion
      # construct a transformation which is constant on that edge
```

doc/semitrans.xml
```diff
@@ -376,3 +376,19 @@ gap> Size(WW);
...
+<#GAPDoc Label="PartitionMonoidSmallGeneratingSet">
+   <ManSection>
+    <Oper Name="SmallGeneratingSet" Arg="G"/>
+    <Returns>A small generating set with maximal idempotent elements.</Returns>
+    <Description>
+      When <A>G</A> is a subgroup of the Partition Monoid on <A>n</A> points. 
+	  of the form <A>N</A> union <C>SingularPartitionMonoid(n)</C> where <A>N</A>
+	  is a subgroup of the symmetric group on <A>n</A> points then this
+	  returns the small generating set which contains an equal number of 
+	  idempotent partitions as the idempotent rank of G. This set would be
```
tomcontileslie, line 389: I think speaking without hypotheticals is clearer when you're talking about how the function works.
```suggestion
	  idempotent partitions as the idempotent rank of G. This set is
```

doc/semitrans.xml
```diff
@@ -376,3 +376,19 @@ gap> Size(WW);
     </Description>
   </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="PartitionMonoidSmallGeneratingSet">
+   <ManSection>
+    <Oper Name="SmallGeneratingSet" Arg="G"/>
+    <Returns>A small generating set with maximal idempotent elements.</Returns>
+    <Description>
+      When <A>G</A> is a subgroup of the Partition Monoid on <A>n</A> points.
```
tomcontileslie, line 385: I think this is a confusing way to word the type of argument this function works on. You explained earlier that, actually, as long as `G` is a semigroup that contains the singular partition monoid on `n` points as well as at least one permutation, that fits the bill. Maybe something like this would be clearer:
```
If <A>G</A> is a bipartition semigroup on <M>n</M> points
which contains every element of <C>SingularPartitionMonoid(n)</C>,
along with at least one permutation, then this function returns...
```
Remember that you should be using:
- `<A>...</A>` tags for **arguments**, i.e. in this case only `G`
- `<C>...</C>` tags for **code**, such as `SingularPartitionMonoid(n)`
- `<M>...</M>` tags for **mathematical expressions**, including, I would say, `n`.

tomcontileslie:
@Peter-Ing I think you may have forgotten to declare the local variable `elements` there!

james-d-mitchell:
I tried in 2022 to fix this PR up so that it could be merged, but basically gave up. I don't foresee having the time to do this in the near future, and it might be quicker to do it again from scratch since the original author is no longer available.

PR #698, opened 2020-08-11
Revise C level access to GAP lists in fropin.cc

fingolfin:
- move `ELM_PLIST2` from fropin.h to fropin.cc
- `SET_ELM_PLIST2` called CHANGED_BAG on the wrong list (resulting in a
  potential GC crash); change it to use `ASS_LIST` which avoids this and in
  general will take care of everything for us at a minuscule performance cost
- in particular, this renders several calls to `RetypeBag` unnecessary
- a call to `RetypeBag(rules, T_PLIST_CYC)` was incorrect (because the
  elements of `rules` actually seem to be plists, not cyclotomics); instead
  use `ASS_LIST` to let the GAP kernel worry about setting suitable TNUMs
- use `ASS_LIST` and `AssPlist` in a few more places
- tiny optimization: instead of `INT_INTOBJ(expr) == 0` it is better to do
  `expr = INTOBJ_INT(0)` as the compiler can precompute the right side
- drop some redundant calls to `SET_LEN_PLIST`: new plists always have
  length 0, no need to set that; and `AssPlist` / `ASS_LIST` automatically
  set the correct list length for us

### DIFF
--- a/src/fropin.cc
+++ b/src/fropin.cc
@@ -29,13 +29,13 @@ using libsemigroups::detail::Timer;
 
 // Macros for the GAP version of the algorithm
 
+#define ELM_PLIST2(plist, i, j) ELM_PLIST(ELM_PLIST(plist, i), j)
+
 #define INT_PLIST(plist, i) INT_INTOBJ(ELM_PLIST(plist, i))
 #define INT_PLIST2(plist, i, j) INT_INTOBJ(ELM_PLIST2(plist, i, j))
 
 inline void SET_ELM_PLIST2(Obj plist, UInt i, UInt j, Obj val) {
-  SET_ELM_PLIST(ELM_PLIST(plist, i), j, val);
-  SET_LEN_PLIST(ELM_PLIST(plist, i), j);
-  CHANGED_BAG(plist);
+  ASS_LIST(ELM_PLIST(plist, i), j, val);
 }
 
 // Fast product using left and right Cayley graphs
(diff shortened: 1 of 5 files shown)


src/fropin.cc
```diff
@@ -29,13 +29,13 @@ using libsemigroups::detail::Timer;
...
 
+#define ELM_PLIST2(plist, i, j) ELM_PLIST(ELM_PLIST(plist, i), j)
+
 #define INT_PLIST(plist, i) INT_INTOBJ(ELM_PLIST(plist, i))
 #define INT_PLIST2(plist, i, j) INT_INTOBJ(ELM_PLIST2(plist, i, j))
 
 inline void SET_ELM_PLIST2(Obj plist, UInt i, UInt j, Obj val) {
-  SET_ELM_PLIST(ELM_PLIST(plist, i), j, val);
-  SET_LEN_PLIST(ELM_PLIST(plist, i), j);
-  CHANGED_BAG(plist);
```
fingolfin, line 38: This should have been `CHANGED_BAG(ELM_PLIST(plist, i));`, so in the above form there is potential for a GC crash. But of course `ASS_LIST` also does that for us now.

james-d-mitchell:
Thanks @fingolfin for the fixes, and the rationale. The CI is failing in one job, but I think this is probably related to the CI setup rather than anything else. We'll try to resolve this today, and then merge your PR.

PR #703, opened 2020-10-07
ci: resolve failing tests

james-d-mitchell:


james-d-mitchell:
The `--with-external-libsemigroups` travis job will fail until the conda feedstock for v1.3.3 is merged and then picked up by conda.

Issue #702, opened 2020-09-30
prerequisites.sh may error if libsemigroups already present


mtorpey:
When installing Semigroups 3.3.1 with PackageManager, by archive, the following error gets reported:
```
-e libsemigroups v1.0.6 is required by this version of Semigroups
-e the /home/mtorpey/.gap/pkg/semigroups-3.3.1/libsemigroups directory exists and is non-empty
-e The installed version of libsemigroups is v1.0.6
/home/mtorpey/.gap/pkg/semigroups-3.3.1/prerequisites.sh: 25: [: 1.0.6: unexpected operator
-e Error, the installed version of libsemigroups is too old
```
This seems to be caused by `prerequisites.sh` being called while a version of libsermigroups is already present (bundled in the Semigroups archive).  It looks like `prerequisites.sh:25` is maybe unsafe bash syntax, but I can never remember how this works.

Any idea what's happening here?

james-d-mitchell:
Thanks @mtorpey I'm pretty sure we can fix this!

james-d-mitchell:
I think this is resolved now @mtorpey and @wilfwilson

PR #704, opened 2020-10-07
prerequisites: fix line 25

james-d-mitchell:
@mtorpey I think this resolves issue #702, but can you please double check before we merge it? 

Also @wilfwilson is it the intended behaviour that every line of output from `prerequisites.sh` has the prefix `-e`?

### DIFF
--- a/prerequisites.sh
+++ b/prerequisites.sh
@@ -22,7 +22,7 @@ if [ -d $LIBS_DIR ] && [ "$(ls -A $LIBS_DIR)" ]; then
   INSTALLED=`tr -d '\n' < $LIBS_DIR/.VERSION`
   echo -e "The installed version of libsemigroups is v$INSTALLED"
   LEAST=`echo -e "$VERS\n$INSTALLED" | sort -V | head -n1`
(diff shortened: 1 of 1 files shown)


prerequisites.sh
```diff
@@ -22,7 +22,7 @@ if [ -d $LIBS_DIR ] && [ "$(ls -A $LIBS_DIR)" ]; then
   INSTALLED=`tr -d '\n' < $LIBS_DIR/.VERSION`
   echo -e "The installed version of libsemigroups is v$INSTALLED"
   LEAST=`echo -e "$VERS\n$INSTALLED" | sort -V | head -n1`
```
wilfwilson, line 24: I think only this one needs to be `echo -e`, to convert the `\n` into a newline character. The rest could all be without the `-e`.

wilfwilson (approved):
Looks like a fix to me, but let's see what @mtorpey says.

james-d-mitchell:
Thanks @wilfwilson !

mtorpey:
Hi folks -- this fixes my problem!  Thanks :)

PR #708, opened 2020-11-09
Replace some uses of IMMUTABLE

fingolfin:
Use of IMMUTABLE is an internal kernel detail, and there are some plans
to eventually move it from a bit in the TNUM to a separate object flag
at some point.

PR #709, opened 2020-11-10
Remove special case for GAP < 4.9, which is no longer supported

wilfwilson:
I noticed while reviewing #708 that we can get ride of the special case here. The minimum version of GAP required by Semigroups is currently 4.9.0.


wilfwilson:
For now I've still got some Travis access, so the CI should show here: https://travis-ci.com/github/wilfwilson/Semigroups/builds/199692590

james-d-mitchell:
Nice thanks @wilfwilson !

PR #717, opened 2020-11-18
ci: add more actions

james-d-mitchell:


james-d-mitchell:
This currently fails because `libsemigroups` isn't present when we try to configure the Semigroups package. I'm not sure how to resolve this at present, so I'm going to leave this PR open until I figure it out, or decide to give up.

.github/workflows/GAP.yml
```diff
@@ -0,0 +1,26 @@
...
+        os: [ubuntu-latest]
+        gap-branch:
+          - master
+          - stable-4.11
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gap-actions/setup-gap-for-packages@v1
+        with:
+          GAP_PKGS_TO_BUILD: "digraphs genss io orb images datastructures"
```
wilfwilson, line 24: You should add `profiling` to this list to get code coverage.

.github/workflows/GAP.yml
```diff
@@ -0,0 +1,26 @@
...
+  test:
+    name: "GAP ${{ matrix.gap-branch }} on ${{ matrix.os }}"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        gap-branch:
+          - master
+          - stable-4.11
```
wilfwilson, line 18: You _should_ be able to add `stable-4.10` here with no extra work. The reason I haven't done this yet for Digraphs is because there's a tiny bit of extra effort that will be needed to install NautyTracesInterface (which is optional, and not needed here).

wilfwilson (commented):
What we need is for [the `pkg-ci-scripts/build_pkg.sh` file](https://github.com/gap-system/pkg-ci-scripts/blob/master/build_pkg.sh), which is ultimately run by the `gap-actions/setup-gap-for-packages` action to compile Semigroups, to run the `prerequisites.sh` file in the repository before compiling the Semigroups package.

I have made a PR to make this change at https://github.com/gap-system/pkg-ci-scripts/pull/9; once it is merged, these actions should work. I hope.

.github/workflows/GAP.yml
james-d-mitchell: Thanks!

james-d-mitchell:
> What we need is for [the `pkg-ci-scripts/build_pkg.sh` file](https://github.com/gap-system/pkg-ci-scripts/blob/master/build_pkg.sh), which is ultimately run by [the `gap-actions/setup-gap-for-packages` action](https://github.com/gap-actions/setup-gap-for-packages/blob/master/action.yml) to compile Semigroups, to first run the `prerequisites.sh` file in the repository before compiling the Semigroups package.
> 
> I have made a PR to make this change at [gap-system/pkg-ci-scripts#9](https://github.com/gap-system/pkg-ci-scripts/pull/9); once it is merged, these actions should work. I hope.

Thanks @wilfwilson that's super helpful. I'll check periodically to see if your PR is merged, and then update this PR!

.github/workflows/GAP.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gap-actions/setup-gap-for-packages@v1
+        with:
+          GAP_PKGS_TO_BUILD: "digraphs genss io orb images datastructures"
+          GAPBRANCH: ${{ matrix.gap-branch }}
+      - uses: gap-actions/run-test-for-packages@v1
```
wilfwilson, line 26: Actually @james-d-mitchell, you _will_ need to do something else.

The action `gap-actions/run-test-for-packages` ultimately runs the file https://github.com/gap-system/pkg-ci-scripts/blob/master/run_tests.sh, which checks the exit code given by GAP when running a particular file.

It runs the file in the variable `GAP_TESTFILE`, if any, but by default it runs `tst/testall.g`. However, the value of `GAP_TESTFILE` is not (yet) passed to the action `gap-actions/run-test-for-packages`. I have opened a PR for this at https://github.com/gap-actions/run-test-for-packages/pull/2; once it is merged, if you wanted, you'd be able to do something like:
```suggestion
      - uses: gap-actions/run-test-for-packages@WHATEVER (probably @master for now)
        with:
          GAP_TESTFILE: "CUSTOM_FILE_PATH"
```
There is a TODO in https://github.com/gap-system/pkg-ci-scripts/blob/master/run_tests.sh suggesting that eventually it will use the value of `TestFile` in the `PackageInfo.g`, if present (which for us would be `tst/teststandard.g`), but that is not yet implemented either.

For now, I think your best bet is to create a new file `tst/testall.g` as I did for Digraphs, and put the tests in there that you want to run. This could be a copy of `tst/teststandard.g`, but you might want to run more stuff too.

james-d-mitchell:
Superseded by #718 !

PR #718, opened 2020-11-19
ci: add more actions (working)

wilfwilson:
I've got a working version of #717, because my PRs there were merged. You can either merge this PR, or close it and take the code for use in #717. I just thought it would be easiest for you to get the code this way.

It wasn't completely simple to add GAP `stable-4.10` support (some of the packages contained in `bootstrap-pkg-full` for GAP 4.10 are too old for this package, so would have to be manually replaced), so I haven't done that.

All this does is run `tst/teststandard.g`, which runs `SemigroupsTestStandard()`. If you want it to do more (e.g. compile the manual, test the manual examples, run `SemigroupsTestInstall`, run extreme tests) then I recommend you make a file `tst/testall.g` analogous to the one in the Digraphs package to do what you want.

### DIFF
--- a/.github/workflows/GAP.yml
+++ b/.github/workflows/GAP.yml
@@ -1,0 +1,28 @@
+name: "GAP tests"
+
+# Trigger the workflow on push or pull request
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: "GAP ${{ matrix.gap-branch }} on ${{ matrix.os }}"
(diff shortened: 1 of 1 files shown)


wilfwilson:
That said, despite the ambitious PR title, perhaps it's not working... We'll see.

james-d-mitchell:
Thanks @wilfwilson, awesome!

PR #721, opened 2020-12-02
CI: don't build pushes to release candidate branches

wilfwilson:
This is basically the same as https://github.com/digraphs/Digraphs/pull/360

Once this is merged, we should merge `stable-3.4` into `master` so that both branches have the up-to-date CI.


james-d-mitchell:
Thanks @wilfwilson i somehow overlooked this until now, this is most useful!

PR #723, opened 2020-12-16
Extend congruence methods to equivalence relations

flsmith:
Allows `EquivalenceRelationLookup` and `EquivalenceRelationCanonicalLookup` to be applied to equivalence relations.

I didn't really examine the current setup very hard for subtleties and there should probably be more functions this is done to, hence the draft. Should possibly `TryNextMethod` instead of `Error` as well.


james-d-mitchell:
Seems like this works just fine, can we remove the draft status?

flsmith:
Sure @james-d-mitchell

PR #725, opened 2020-12-23
Remove uf

james-d-mitchell:
This PR removes from `Semigroups` the use of the union-find datastructures implemented in `libsemigroups` in favour of using the implementation in the `datastructures` package. 

Pros:

1) Simplifies the kernel module
2) `Semigroups` is compatible with newer versions of `libsemigroups`

Possible cons:

1) introduces dependency on `datastructures` package. This dependency already existed, via Digraphs, so this is actually not a con.


wilfwilson:
This PR is branched off `master`, but you're making the PR to `stable-3.4`. These two things should be consistent. Which one do you want?

james-d-mitchell:
I want stable, my bad will up date when I get back in front of the computer

james-d-mitchell:
Updated!

james-d-mitchell:
One of the CI jobs seems to be spending an inordinate amount of time downloading something, any insights @wilfwilson ?

Issue #671, opened 2020-06-02
Implement isomorphism from arbitrary band to a semilattice of rectangular bands


james-d-mitchell:
As described in Howie's book Chapter 4.

james-d-mitchell:
@rcirpons, @MTWhyte, and Tom.

tomcontileslie:
There isn't always an isomorphism from a band to a **strong** semilattice of rectangular bands. Currently, the only semilattice of semigroups-type object we have in GAP is a strong semilattice of semigroups. So, we need to do one of the following to implement decomposition functions.

## 1. Change the restrictions on the source so that the isomorphism can have an SSS object as its range.

By an SSS object, I mean those created in PR #673.

[This paper](https://link.springer.com/content/pdf/10.1007/s10474-015-0504-y.pdf) by Z.-P. Wang references a number of papers that collectively show that the following types of semigroups can be decomposed into **strong** semilattices of semigroups:

- Normal cryptogroups;
- Normal orthogroups;
- Clifford semigroups.

I also think @reiniscirpons may have a different reference showing that an isomorphism always exists from a *normal band* to an SSS of rectangular bands.

Howie's introduction to semigroups provides a construction for the isomorphism from a Clifford semigroup to a strong semilattice of semigroups (of groups, in fact). My first step in creating some computational tools might therefore be to implement an `IsomorphismSemigroup` method from the `IsCliffordSemigroup` filter to the `IsStrongSemilatticeOfSemigroups` filter, if that sounds good? As a bit of a warm-up. I know that's not band decomposition, but there are a load of very similar decomposition methods to be implemented here. More implementations can come for other types of semigroups which decompose into SSSs.

## 2. Be less restrictive than a Strong Semilattice of Rectangular Bands.

As in this issue's title, every band decomposes into a semilattice of rectangular bands. We do not currently have a semilattice of semigroups object in GAP, but the same paper by Wang referenced above contains a general description of a **fundamental semilattice of semigroups**, which I believe generalises the type of semilattice that a general band decomposes into. So, if we can implement a `FundamentalSemilatticeOfSemigroups` object in GAP, we can create the desired isomorphism.

PR #731, opened 2021-01-13
Decomposition of some Completely Regular Semigroups into Strong Semilattices of Semigroups

tomcontileslie:
**May 2021 update: this PR has only ticked one of four boxes, but it works on Clifford semigroups. Implementing IsomorphismSemigroup for other types of completely regular semigroups can be done in a different PR.**

This PR introduces methods for decomposing certain completely regular semigroups into strong semilattices, as suggested in Issue #671. The aim is introduce `IsomorphismSemigroup` methods from the following types of semigroups:

- [X] Clifford semigroups
- [ ] Normal orthogroups
- [ ] Normal cryptogroups
- [ ] Normal bands (?)

to SSS objects.

### DIFF
--- a/gap/attributes/isomorph.gi
+++ b/gap/attributes/isomorph.gi
@@ -332,7 +332,7 @@ InstallMethod(IsomorphismSemigroup,
 "for IsStrongSemilatticeOfSemigroups and a Clifford semigroup",
 [IsStrongSemilatticeOfSemigroups, IsSemigroup and IsFinite],
 function(filt, S)
-  local A, idemps, n, D, N, L, classes, idemp, DC, H, map, SSS, i, j;
+  local A, idemps, n, D, N, L, classes, idemp, DC, H, SSS, i, j, addfunc;
   # decomposes a finite Clifford semigroup S into a strong semilattice of
   # groups and returns an SSS object.
   if not (IsCliffordSemigroup(S) and IsFinite(S)) then
@@ -366,12 +366,13 @@ function(filt, S)
   for i in [1 .. n] do
     idemp := idemps[i];
     Add(H, []);
+    addfunc := function(x, a, b, c)  # horrible namespace hack (credits JDM)
+      Add(x, [a, b, z -> c * z]);
+    end;
     for j in N[i] do
-      map := function(elm)
-        return idemp * elm;
-      end;
-      Add(H[i], MappingByFunction(L[j], L[i], map));
+      addfunc(H[i], L[j], L[i], idemps[i]);
     od;
+    Apply(H[i], x -> MappingByFunction(x[1], x[2], x[3]));
   od;
 
   SSS := StrongSemilatticeOfSemigroups(D, L, H);
--- a/gap/attributes/isomorph.gi
+++ b/gap/attributes/isomorph.gi
@@ -327,3 +327,60 @@ function(S)
   UseIsomorphismRelation(H, G);
   return H;
 end);
+
+InstallMethod(IsomorphismSemigroup,
+"for IsStrongSemilatticeOfSemigroups and a Clifford semigroup",
+[IsStrongSemilatticeOfSemigroups, IsSemigroup and IsFinite],
+function(filt, S)
+  local A, idemps, n, D, N, L, classes, idemp, DC, H, map, SSS, i, j;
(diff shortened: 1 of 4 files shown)


gap/attributes/properties.gd
```diff
@@ -102,3 +102,6 @@ DeclareProperty("IsSurjectiveSemigroup", IsSemigroup);
 InstallTrueMethod(IsSurjectiveSemigroup, IsRegularSemigroup);
 InstallTrueMethod(IsSurjectiveSemigroup, IsMonoidAsSemigroup);
 InstallTrueMethod(IsSurjectiveSemigroup, IsIdempotentGenerated);
+
+DeclareProperty("IsOrthogroup", IsSemigroup);
+DeclareSynonym("IsOrthoGroup", IsOrthogroup);
```
wilfwilson, line 107:
```suggestion
DeclareSynonymAttr("IsOrthoGroup", IsOrthogroup);
```

gap/attributes/isomorph.gi
```diff
@@ -327,3 +327,60 @@ function(S)
...
+  # since the semilattice decomposition is by J-classes, and J = D here
+  A      := Semigroup(Idempotents(S));
+  idemps := Elements(A);
+  n      := Size(idemps);
+
+  # create semilattice
+  D := DigraphReflexiveTransitiveReduction(Digraph(NaturalPartialOrder(A)));
+  # currently wrong way round
+  D := DigraphReverse(D);
+  N := OutNeighbours(D);
```
wilfwilson, line 351: You can directly get this via
```suggestion
  N := ReverseNaturalPartialOrder(A);
```
(although I don't think that necessarily (or at all?) contains `x` in `N[x]` for each `x` - if those are actually necessary, you will still want to add them or modify your code below to acts as if they are there).

gap/attributes/isomorph.gi
```diff
@@ -327,3 +327,60 @@ function(S)
...
+  H := [];
+  for i in [1 .. n] do
+    idemp := idemps[i];
+    Add(H, []);
+    for j in N[i] do
+      map := function(elm)
+        return idemp * elm;
+      end;
+      Add(H[i], MappingByFunction(L[j], L[i], map));
+    od;
```
wilfwilson, line 374: It seems like there is no need to keep re-creating the function `map` inside the `for` loop.
```suggestion
    map := elm -> idemp * elm;
    for j in N[i] do
      Add(H[i], MappingByFunction(L[j], L[i], map));
    od;
    # Add(H[i], MappingByFunction(L[i], L[i], map));  # Add this if you do actually need `i` to be in `N[i]`
```

gap/attributes/isomorph.gi
tomcontileslie: Thanks @wilfwilson, unfortunately I get a "no method found" error when running ReverseNaturalPartialOrder on some inputs - for example:
```
gap> S := Semigroup(Transformation([1, 2, 4, 5, 6, 3, 7, 8]), Transformation([3, 3, 4, 5, 6, 2, 7, 8]), Transformation([1, 2, 5, 3, 6, 8, 4, 4]));
<transformation semigroup of degree 8 with 3 generators>
gap> IsCliffordSemigroup(S);
true
gap> A := Semigroup(Idempotents(S));
<transformation monoid of degree 8 with 3 generators>
gap> NaturalPartialOrder(A);
[ [ 2, 3, 4 ], [ 4 ], [ 4 ], [  ] ]
gap> ReverseNaturalPartialOrder(A);
Error, no 2nd choice method found for `ReverseNaturalPartialOrder' on 1 arguments at /Applications/GAP/lib/methsel2.g:250 called from
<function "HANDLE_METHOD_NOT_FOUND">( <arguments> )
 called from read-eval loop at *stdin*:18
type 'quit;' to quit to outer loop
```
Although you are correct that we don't need `x` in `N[x]` since the homomorphisms in this case are the identity, and the strong semilattice constructor is clever enough to fill that in upon creation.

tomcontileslie: Also, I'm realising that taking the reflexive transitive reduction of `D` is redundant, since the SSS constructor effectively reverts this, so I'll remove that part

PR #736, opened 2021-01-25
Suppress incorrect "1 generators" in test output

wilfwilson:
In a future GAP version, "1 generators" will change to "1 generator", and so to maintain forwards and backwards compatibility we should hide the output. It's inconsequential in these examples anyway.


james-d-mitchell (approved):
Thanks @wilfwilson, happy to have this

PR #738, opened 2021-01-28
Release candidate for v3.4.1

james-d-mitchell:


README.md
```diff
@@ -30,21 +29,23 @@ can be obtained from the main [GAP](http://www.gap-system.org) page.
...
-
-* get the [orb] package version 4.8.2 or higher
+* get the [datastructures](http://gap-packages.github.io/datastructures)
+  package version 0.2.5 or higher 
 
-Both [orb] and [Semigroups] perform better when [orb] is compiled, so compile
-[orb]!
-
-* ensure that the [Digraphs] package version 1.0.0 or higher is available.
+* ensure that the [Digraphs] package version 1.2.0 or higher is available.
```
wilfwilson, line 35: Whoops, for some reason I've had it in my head that links of this kind have to follow the syntax `[keyword][]` (with the extra square brackets at the end. And I've just learned here that that's apparently not required! So we should probably removed the extra `[]`s from the CHANGELOG and README in this package and in Digraphs.

james-d-mitchell: Oh yes, I'm in the same boat precisely! Does remove the extra `[]` really work?

wilfwilson: We should check in the Markdown specification somewhere, but on GitHub at least, it seems to render properly:

https://github.com/semigroups/Semigroups/blob/master/README.md

PR #739, opened 2021-01-28
Various small technical tweaks

wilfwilson:
As described by the commit messages:
* Update some peoples' URLs
* `doc/main.xml`: update some package URLs
* Change `http`→`https` where possible
* `properties.gi`: can not → cannot
* `semigrp.xml`: update Log "1 generators → 1 generator"
* Use standard Markdown syntax for links
* Remove checks concerning GAP versions < 4.10
* Fix weird whitespace issue
    * The linter (locally on my computer) was saying that the second character in this lines was not spaces. Even though it looked like a space. So I replaced it with an actual space.

### DIFF
--- a/gap/attributes/properties.gi
+++ b/gap/attributes/properties.gi
@@ -1024,7 +1024,7 @@ function(S)
   fi;
 
   if not IsFinite(S) then
-    # WW I cannot find an example of an example of an infinite inverse monoid
+    # WW I cannot find an example of an infinite inverse monoid
     # with InverseOp - FreeGroup(1) should work, but it is not in the category
     TryNextMethod();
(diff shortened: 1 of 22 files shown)


gap/attributes/properties.gi
```diff
@@ -1024,7 +1024,7 @@ function(S)
   fi;
 
   if not IsFinite(S) then
-    # WW I can not find an example of an example of an infinite inverse monoid
+    # WW I cannot find an example of an example of an infinite inverse monoid
```
james-d-mitchell, line 1027:
```suggestion
    # WW I cannot find an example of an infinite inverse monoid
```

PR #744, opened 2021-02-07
Release candidate for v3.4.2

james-d-mitchell:


makedoc.g
```diff
@@ -6,12 +6,12 @@
...
 
-PACKAGE := "Semigroups";
 LoadPackage("GAPDoc");
 LoadPackage("semigroups");
 
 SemigroupsMakeDoc();
 CopyHTMLStyleFiles("doc");
-GAPDocManualLab(PACKAGE);
+# The next step loads the Semigroups package!
+GAPDocManualLab("Semigroups");
```
wilfwilson, line 15: The Semigroups package is already loaded on line 10.

james-d-mitchell: Right! But it would be possible to build the doc without loading the package, if this line and line 10 weren't included

james-d-mitchell: I should remove this comment

wilfwilson: Ah I see 🙂

PR #756, opened 2021-03-22
Only declare `DirectProductOp`/`SmallGeneratingSet` when required

wilfwilson:
In GAP versions before 4.12, these are only declared for `IsGroup`.  This means that we have to separately declare them for `IsSemigroup`.  This then gives warnings when further methods are installed for `IsGroup`, (e.g. if you load the polycyclic package after loading Semigroups) because they don't know which declaration they belong to.  Not a huge problem, but one with an easy fix.

The GAP master branch does, and GAP 4.12 will, declare these things for `IsSemigroup` instead – see https://github.com/gap-system/gap/pull/4329. Therefore we should only make these declarations when GAP has not already made them.  Eventually, the package will presumably require GAP 4.12, and so all of this code can be deleted. I have added comments to that effect.

It turns out that the repeated declaration of attributes does not cause a warning to be printed, although repeated declarations of operations do. So when loading Semigroups `stable-3.4` in the GAP `master` branch right now, you get only the warning:
```
 Loading the library and packages ...
#I  equal requirements in multiple declarations for operation `DirectProductOp'
```
That means we only 'need' to make the change in this PR for `DirectProductOp`, but I still think it'd be better to do both.

### DIFF
--- a/gap/attributes/attr.gd
+++ b/gap/attributes/attr.gd
@@ -35,7 +35,11 @@ DeclareAttribute("SmallInverseSemigroupGeneratingSet",
                  IsMultiplicativeElementCollection);
 DeclareAttribute("SmallInverseMonoidGeneratingSet",
                  IsMultiplicativeElementWithOneCollection);
-DeclareAttribute("SmallGeneratingSet", IsSemigroup);
+# SmallGeneratingSet was only declared for IsGroup before GAP 4.12, so the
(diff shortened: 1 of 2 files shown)


wilfwilson:
We should probably wait until https://github.com/gap-system/gap/pull/4329 is merged before merging this, in case I'm doing something silly (which the reviews on that PR might reveal).

james-d-mitchell (approved):
I approve!

PR #759, opened 2021-03-31
Allow ElementOfFpSemigroup to take simpler args

tomcontileslie:
This PR adds a new implementation of [`ElementOfFpSemigroup`](https://www.gap-system.org/Manuals/doc/ref/chap52.html#X847012347856C55E) from the main GAP distribution. The original implementation needs the first argument to be an element family, and the second to be an associative word. For ease of use, I thought it might be helpful to have a shorthand that can take an fp semigroup as the first argument. That way, you get reasonable output rather than an error when running the final line of this example:
```
gap> f := FreeSemigroup("x", "y");;
gap> AssignGeneratorVariables(f);
#I  Assigned the global variables [ x, y ]
gap> s := f / [[x * x, x], [y * y, y]];
<fp semigroup on the generators [ x, y ]>
gap> a := ElementOfFpSemigroup(s, x * y);
x*y
```

I also added the corresponding `ElementOfFpMonoid` function.

### DIFF
--- a/doc/semifp.xml
+++ b/doc/semifp.xml
@@ -41,3 +41,51 @@ gap> f2 / ParseRelations(GeneratorsOfSemigroup(f2), "a = a^2");
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="ElementOfFpSemigroup">
+<ManSection>
+  <Oper Name="ElementOfFpSemigroup" Arg="S, word"/>
+  <Returns>An element of the fp semigroup <A>S</A>.</Returns>
+  <Description>
+    When <A>S</A> is a finitely presented semigroup and <A>word</A> is an
+    associative word in the associated free semigroup (see
+    <Ref Filt="IsAssocWord" BookName="ref"/>), this returns the
+    fp semigroup element with representative <A>word</A>.
+    <P/>
+
+    This function is just a short form of the &GAP; library implementation of
+    <Ref Oper="ElementOfFpSemigroup" BookName="ref"/> which does not require
+    retrieving an element family.
+
+    <Example><![CDATA[
+gap> f := FreeSemigroup("x", "y");;
+gap> AssignGeneratorVariables(f);
+gap> s := f / [[x * x, x], [y * y, y]];
+<fp semigroup on the generators [ x, y ]>
+gap> a := ElementOfFpSemigroup(s, x * y);
+x*y
+gap> b := ElementOfFpSemigroup(s, x * y * y);
+x*y^2
+gap> a in s;
+true
+gap> a = b;
+true]]></Example>
+  </Description>
+</ManSection>
+<#/GAPDoc>
+
+<#GAPDoc Label="ElementOfFpMonoid">
+<ManSection>
+  <Oper Name="ElementOfFpSemigroup" Arg="M, word"/>
+  <Returns>An element of the fp monoid <A>M</A>.</Returns>
+  <Description>
+    When <A>M</A> is a finitely presented monoid and <A>word</A> is an
+    associative word in the associated free monoid (see
+    <Ref Filt="IsAssocWord" BookName="ref"/>), this returns the
+    fp monoid element with representative <A>word</A>.
+    <P/>
+
+    This is analogous to <Ref Oper="ElementOfFpSemigroup"/>.
+  </Description>
+</ManSection>
+<#/GAPDoc>
--- a/gap/semigroups/semifp.gi
+++ b/gap/semigroups/semifp.gi
@@ -43,6 +43,16 @@ function(x)
   return ExtRepOfObj(UnderlyingElement(x));
 end);
 
+InstallMethod(ElementOfFpSemigroup,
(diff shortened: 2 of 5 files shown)


doc/semifp.xml
```diff
@@ -41,3 +41,51 @@ gap> f2 / ParseRelations(GeneratorsOfSemigroup(f2), "a = a^2");
...
+  <Oper Name="ElementOfFpSemigroup" Arg="S, word"/>
+  <Returns>An element of the fp semigroup <A>S</A>.</Returns>
+  <Description>
+    When <A>S</A> is a finitely presented semigroup and <A>word</A> is an
+    associative word in the associated free semigroup (see
+    <Ref Filt="IsAssocWord" BookName="ref"/>), this returns the
+    fp semigroup element with represenative <A>word</A>.
+    <P/>
+
+    This function is just a shorthand for the GAP implementation of
```
james-d-mitchell, line 56:
```suggestion
    This function is just a shorthand for the &GAP; library implementation of
```

doc/semifp.xml
```diff
@@ -41,3 +41,51 @@ gap> f2 / ParseRelations(GeneratorsOfSemigroup(f2), "a = a^2");
...
+
+<#GAPDoc Label="ElementOfFpSemigroup">
+<ManSection>
+  <Oper Name="ElementOfFpSemigroup" Arg="S, word"/>
+  <Returns>An element of the fp semigroup <A>S</A>.</Returns>
+  <Description>
+    When <A>S</A> is a finitely presented semigroup and <A>word</A> is an
+    associative word in the associated free semigroup (see
+    <Ref Filt="IsAssocWord" BookName="ref"/>), this returns the
+    fp semigroup element with represenative <A>word</A>.
```
james-d-mitchell, line 53:
```suggestion
    fp semigroup element with representative <A>word</A>.
```

doc/semifp.xml
```diff
@@ -41,3 +41,51 @@ gap> f2 / ParseRelations(GeneratorsOfSemigroup(f2), "a = a^2");
...
+  <Oper Name="ElementOfFpSemigroup" Arg="M, word"/>
+  <Returns>An element of the fp monoid <A>M</A>.</Returns>
+  <Description>
+    When <A>M</A> is a finitely presented monoid and <A>word</A> is an
+    associative word in the associated free monoid (see
+    <Ref Filt="IsAssocWord" BookName="ref"/>), this returns the
+    fp monoid element with representative <A>word</A>.
+    <P/>
+
+    This is analogous to <Ref Oper="ElementOfFpSemigroup"/> above.
```
james-d-mitchell, line 88:
```suggestion
    This is analogous to <Ref Oper="ElementOfFpSemigroup"/>.
```

gap/semigroups/semifp.gi
```diff
@@ -43,6 +43,20 @@ function(x)
...
+"for an fp semigroup and an associative word",
+[IsFpSemigroup, IsAssocWord],
+function(s, w)
+  return ElementOfFpSemigroup(FamilyObj(s.1), w);
+end);
+
+InstallMethod(ElementOfFpMonoid,
+"for an fp monoid and an associative word",
+[IsFpMonoid, IsAssocWord],
+function(m, w)
```
james-d-mitchell, line 56: This could be `{m, w} -> ElementOfFpSemigroup(FamilyObj(m.1), w);`

gap/semigroups/semifp.gi
```diff
@@ -43,6 +43,20 @@ function(x)
   return ExtRepOfObj(UnderlyingElement(x));
 end);
 
+InstallMethod(ElementOfFpSemigroup,
+"for an fp semigroup and an associative word",
+[IsFpSemigroup, IsAssocWord],
+function(s, w)
+  return ElementOfFpSemigroup(FamilyObj(s.1), w);
```
james-d-mitchell, line 50: Might it not be a good idea to add a check that `S` has at least one generator here? I think it's not possible to create an fp semigroup with 0 generators, but it'd be good to check/assert this explicitly here, maybe?

james-d-mitchell (approved):
Thanks @tomcontileslie I approve, there are a couple of very minor suggestions/comments, but otherwise, I'll be happy to merge this!

gap/semigroups/semifp.gi
tomcontileslie: I'm happy to implement this, hadn't really thought about that. I'm not sure how many ways there are of creating an fp semigroup with 0 generators; I've found this method, but surprisingly `s.1` is still defined and works fine.

```
gap> f := FreeSemigroup([]);
<free semigroup on the generators [  ]>
gap> s := f / [];
<fp semigroup on the generators [  ]>
gap> s.1;

gap> s.1 = ElementOfFpSemigroup(FamilyObj(s.1), f.1);
true
```

The issue will be exactly the same for `ElementOfFpMonoid` by the way, since the generators `m.1`, `m.2` etc. do not include the identity. What do you think we should do, and how should we check that `s.1` will be defined?

PR #760, opened 2021-04-02
Method for AssignGeneratorVariables for fp semigroups and monoids

### LINKED ISSUE #758: AssignGeneratorVariables for fp semigroups
This is a feature request. Currently in `/gap/semigroups/semifp.gi` there are methods to `AssignGeneratorVariables` for a free semigroup and for a free monoid. Would it not also be of interest to be able to do:
```
gap> f := FreeSemigroup("a", "b");
<free semigroup on the generators [ a, b ]>
gap> AssignGeneratorVariables(f);
#I  Assigned the global variables [ a, b ]
gap> s := f / [[a * b, b * a]];
<fp semigroup on the generators [ a, b ]>
gap> AssignGeneratorVariables(s);  # this is the requested feature
#I  Something something...
gap> a * b = b * a;
true
```
I guess this would overwrite the previously assigned free semigroup variables `a` and `b`. But then again, at least based on my experience, if you're assigning generator variables for a free semigroup then creating an fp semigroup, there is little use for the original variables once you've got the fp semigroup, so this doesn't seem unreasonable.

tomcontileslie:
Fixes #758.

### DIFF
--- a/gap/semigroups/semifp.gi
+++ b/gap/semigroups/semifp.gi
@@ -405,6 +405,18 @@ function(S)
   DoAssignGenVars(GeneratorsOfMonoid(S));
 end);
 
+InstallMethod(AssignGeneratorVariables, "for an fp semigroup",
+[IsFpSemigroup],
+function(S)
+  DoAssignGenVars(GeneratorsOfSemigroup(S));
+end);
+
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Thanks @tomcontileslie !

Issue #771, opened 2021-05-18
Why was Semigroups not working properly in GitHub Actions?


wilfwilson:
I realise @james-d-mitchell that you've not been involved with the GitHub Actions setup, but nevertheless I'm raising this issue because something weird is going on in GitHub Actions, even though (as far as I'm aware) Semigroups is neither being built nor being run particularly weirdly there. So I suspect this might be a symptom of something deeper. 

And I was hoping that it might be obvious to you what is happening, and therefore how to fix it. I'll be happy to try to help you to debug this, if I can.

As can be seen on #770, this affects the Ubuntu jobs but thankfully not the new macOS job.

This started happening seemingly out of the blue in the last week or two. Presumably the GitHub Actions virtual machines were updated with slightly different software, and there's now a subtle incompatibility.

Here's an example where it goes wrong: https://github.com/semigroups/Semigroups/runs/2614993060?check_suite_focus=true
```
 ┌───────┐   GAP 4.12dev built on 2021-05-18 22:16:54+0000
 │  GAP  │   https://www.gap-system.org
 └───────┘   Architecture: x86_64-pc-linux-gnu-default64-kv8
 Configuration:  gmp 6.2.0, GASMAN, readline
 Loading the library and packages ...
 Packages:   AClib 1.3.2, Alnuth 3.1.2, AtlasRep 2.1.0, AutoDoc 2020.08.11, 
             AutPGrp 1.10.2, CRISP 1.4.5, Cryst 4.1.23, CrystCat 1.1.9, 
             CTblLib 1.3.1, FactInt 1.6.3, FGA 1.4.0, Forms 1.2.5, 
             GAPDoc 1.6.4, genss 1.6.6, IO 4.7.0, IRREDSOL 1.4.1, 
             LAGUNA 3.9.3, orb 4.8.3, Polenta 1.3.9, Polycyclic 2.15.1, 
             PrimGrp 3.4.1, RadiRoot 2.8, recog 1.3.2, ResClasses 4.7.2, 
             SmallGrp 1.4.2, Sophus 1.24, SpinSym 1.5.2, TomLib 1.2.9, 
             TransGrp 3.0, utils 0.69
 Try '??help' for help. See also '?copyright', '?cite' and '?authors'
#I  Running tests with acting methods enabled . . .
Architecture: x86_64-pc-linux-gnu-default64-kv8

testing: /home/runner/work/Semigroups/Semigroups/gaproot/pkg/Semigroups/tst/st\
[...]

PR #774, opened 2021-05-19
Fix GitHub Actions CI

wilfwilson:
This should address the immediate problem in #771, but we don't really know why...


james-d-mitchell:
Nice thanks @wilfwilson

PR #775, opened 2021-05-20
Merge stable-3.4 into master

wilfwilson:
I recommend making a merge commit for this one, rather than rebase-and-merge.

This should get the CI working again in the master branch! (Which should be helpful for reviewing @tomcontileslie's PRs)


wilfwilson:
Actually, the CI is not set up quite right for this kind of thing (I'll try to fix it eventually):
- The GitHub Actions CI has already run on this version of the `stable-3.4` branch.
- Therefore it's not going to run on the pull request.
- But (as far as I understand it) the PR tests would run on the hypothetical merge result of this branch into `master`, which may well have different results than the tests in `stable-3.4`.

So I need to come up with a better way of skipping duplicate jobs, while enabling proper tests from PRs from the same repository. For now I'll close this and merge `stable-3.4` into `master` manually.

PR #782, opened 2021-05-20
Add colours to the `prerequisites.sh` script

wilfwilson:
I noticed that the `.release` script basically contains a duplicate of the `prerequisites.sh` script. As a first step to allowing `.release` to just call `prerequisites.sh` directly, I'm copying across the colours that are used in `.release`.

### DIFF
--- a/prerequisites.sh
+++ b/prerequisites.sh
@@ -2,6 +2,15 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
Very good!

PR #784, opened 2021-05-20
Update the `.release` script

wilfwilson:
Primarily, this is about taking advantage of some of the newest changes to ReleaseTools (that I added).

For this to work properly, you will need to update your copy of ReleaseTools to the latest commit in the master branch! (I don’t think ReleaseTools itself has versioned releases.)

Here are the specific changes:

* Do not delete `doc/manual.lab`. (ReleaseTools used to delete this too, but this was a bug, and it no longer does.)
* We need to remove fewer files ourselves. (Because more of them are deleted by ReleaseTools.)
* We no longer need to remove `makedoc.g`. (Because ReleaseTools now builds manuals _before_ calling `.release`, and not after. Therefore we no longer have to trick it into not rebuilding the manual after `.release` is run.)
* Use `prerequisites.sh` directly in `.release`. (It's always nice for maintainability to not have duplicate or near-duplicate code).
* Add TODOs about some more code that can be deleted soon. (These TODOs can be acted upon once my PR https://github.com/gap-system/ReleaseTools/pull/84 is finished and merged.)

### DIFF
--- a/.release
+++ b/.release
@@ -18,6 +18,7 @@ error() {
     exit 1
 }
 
+# TODO: This function can be deleted once the code to compile the manual is gone
 run_gap() {
   gap_output=$( \
     (echo 'OnBreak:=function() Print("FATAL ERROR"); FORCE_QUIT_GAP(1); end;;' ; cat - ; echo ; echo "FORCE_QUIT_GAP(0);") \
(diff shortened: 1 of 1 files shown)


wilfwilson:
Thanks @james-d-mitchell 🙂 Just make sure you update your version of ReleaseTools before making the next Semigroups package release!

Issue #777, opened 2021-05-20
Semigroups doesn't pass cpplint-1.5.5


wilfwilson:
The CI started to use cpplint-1.5.5 rather than cpplint-1.5.4 about an hour ago. This caused the tests to fail, see e.g. https://github.com/semigroups/Semigroups/pull/759/checks?check_run_id=2629215056

For now (see #776) I'm forcing the use of 1.5.4. @james-d-mitchell hopes that this issue will resolve itself when #746 is merged, at which point we can revert #776.

Here's an example of what the problem looks like:
```
src/congpairs.cc:58:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/congpairs.cc:122:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/congpairs.cc:132:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/congpairs.cc:141:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/congpairs.cc:163:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/congpairs.cc:188:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/congpairs.cc:364:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/converter.cc:108:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/converter.h:85:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/converter.h:90:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/converter.h:132:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/converter.h:141:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/fropin.cc:108:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/fropin.cc:112:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
src/fropin.cc:214:  Using C-style cast.  Use static_cast<size_t>(...) instead  [readability/casting] [4]
[...]

Issue #634, opened 2019-12-29
Random crashes in CI tests


wilfwilson:
Over the last couple of weeks, I have noticed sporadic but fairly frequent failed Travis jobs for Semigroups. These seem to be various kinds of actual crashes.

I have restarted all of these jobs as they happened, and they passed eventually. This suggests that it’s perhaps going to be difficult to reproduce locally and track down. But I’ll stop doing that and record the problems here, so that maybe we can work towards resolving this.

I haven’t noticed any patterns yet with which jobs are failing, and where. My guess would that be this is libsemigroups 1.x related - but who knows?!

### 32-bit / master / latest packages

- [14/1/2020](https://travis-ci.org/gap-packages/Semigroups/jobs/636907291)
```
testing: /home/travis/gap/pkg/semigroups/tst/standard/semifp.tst
cannot allocate memory for thread-local data: ABORT
```
- [17/1/2020](https://travis-ci.org/gap-packages/Semigroups/jobs/638546899)
```
testing: /home/travis/gap/pkg/semigroups/tst/standard/semitrans.tst
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
scripts/travis-test.sh: line 69: 22425 Done                    echo "LoadPackage(\"semigroups\"); SemigroupsTestStandard(); SEMIGROUPS.TestManualExamples();"
     22426 Aborted                 (core dumped) | $GAPSH -A -x 80 -m 768m -o $MEM -T 2>&1
     22427                       | tee -a $TESTLOG
```
- [25/1/2020](https://travis-ci.org/gap-packages/Semigroups/builds/641688523)
```
testing: /home/travis/gap/pkg/semigroups/tst/standard/semifp.tst
cannot allocate memory for thread-local data: ABORT
```
- [7/2/2020](https://travis-ci.org/gap-packages/Semigroups/jobs/647412550)
```
testing: /home/travis/gap/pkg/semigroups/tst/standard/semifp.tst
cannot allocate memory for thread-local data: ABORT
```

### 32-bit / stable-4.11 / latest packages

- [28/12/2019](https://travis-ci.org/gap-packages/Semigroups/jobs/630324535)
```
testing: /home/travis/gap/pkg/semigroups/tst/standard/grpperm.tst
cannot allocate memory for thread-local data: ABORT
```

### 32-bit / stable-4.9 / required packages

- [8/1/2020](https://travis-ci.org/gap-packages/Semigroups/jobs/634171245)
[...]

PR #785, opened 2021-05-21
Fix warnings from cpplint-1.5.5

james-d-mitchell:
This resolves issue #777

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,5 +25,5 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
-      - run: pip install 'cpplint==1.5.4'
+      - run: pip install cpplint
       - run: cpplint src/*.h src/*.cc
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,4 +26,4 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
       - run: pip install cpplint
-      - run: cpplint src/*.h src/*.cc
+      - run: bash etc/cpplint.sh
--- a/src/congpairs.cc
+++ b/src/congpairs.cc
@@ -55,7 +55,7 @@ static inline congruence_type cstring_to_congruence_t(char const* type) {
 
 static inline word_type plist_to_word_type(gap_list_t plist) {
   word_type word;
-  for (size_t i = 1; i <= (size_t) LEN_PLIST(plist); i++) {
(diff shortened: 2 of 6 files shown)


wilfwilson:
Could you please change this line back to just `pip install cpplint` as part of this PR:
https://github.com/semigroups/Semigroups/blob/9fbb4c9bbb22110e474b359d0b8b35564be578c8/.github/workflows/lint.yml#L28

wilfwilson:
Note that the (new) macOS job seems to crash really quite often with the kinds of crash that we've been talking about for ages, see for instance #634. It just happened for [the tests on this PR](https://github.com/semigroups/Semigroups/pull/785/checks?check_run_id=2638546375):
```
testing: /Users/runner/work/Semigroups/Semigroups/gaproot/pkg/Semigroups/tst/s\
tandard/semimaxplus.tst
pkg-ci-scripts/run_tests.sh: line 38: 35038 Killed: 9               $GAP ${GAP_TESTFILE:-tst/testall.g}
Error: Process completed with exit code 137.
```

Maybe this gives us an opportunity: we could do the same `ssh` thing that we did on Wednesday, to log in to one of these GitHub Actions macOS instances, and try to run the tests in a debugger until there's a crash. And then we can try to see just what the heck is causing the problem... (I've no idea how to debug in this way, though.)

(In this case, [it also happened in an Ubuntu job](https://github.com/semigroups/Semigroups/pull/785/checks?check_run_id=2638546324), but this seems to be a bit rarer:)
```
testing: /home/runner/work/Semigroups/Semigroups/gaproot/pkg/Semigroups/tst/st\
andard/semirms.tst
pkg-ci-scripts/run_tests.sh: line 38: 32469 Killed                  $GAP ${GAP_TESTFILE:-tst/testall.g}
Error: Process completed with exit code 137.
```

.github/workflows/lint.yml
```diff
@@ -25,5 +25,5 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
-      - run: pip install 'cpplint==1.5.4'
+      - run: pip install cpplint
       - run: cpplint src/*.h src/*.cc
```
wilfwilson, line 29: Come to think of it, I also recommend changing this line so that we don't have to manually keep it in sync with `etc/cpplint.sh`. In particular, your `libsemigroups-v2` branch changes `etc/cpplint.sh`, and so if you make this change now, you'll be future-proofing the listing CI against that change.
```suggestion
      - run: bash etc/cpplint.sh
```

wilfwilson (approved):

.github/workflows/lint.yml
james-d-mitchell: Good idea!

PR #787, opened 2021-05-21
Add GitHub Actions Windows tests

wilfwilson:
This is using the same idea and same setup as https://github.com/digraphs/Digraphs/pull/464.

I want to wait for my setup to settle a bit (and become a bit more 'official') before merging it in. For now, as with Digraphs, everything looks fine! The only problem seems to be that it's slow.


wilfwilson:
@james-d-mitchell I've now added a GitHub Actions Cygwin job to replace the old AppVeyor tests.

From my point of view this is now finished, and thankfully the setup is quite small. The tests even seem to pass!

The big drawback is that the new Cygwin job is very slow. The standard tests and manual examples 'only' take 7 minutes to run, but another 25 minutes is required to get to that point (installing Cygwin, compiling GAP, building Semigroups and its dependencies).

In the longer term, it would be great to speed this up with the use of a Docker image, that sort of thing. But I don't have time for that.

So it's up to you to decide whether you'll accept the slowness of this new job. Currently this new Cygwin job is run whenever the other jobs are run, which is the 'safest' option. Would you prefer something else? (The problem is that pull requests are where a slowdown is most noticeable, but a pull request is also the most likely place where Windows support will become broken).

wilfwilson:
@ChrisJefferson If you're feeling super kind and helpful some time, could you please try to compile the newly released Semigroups v4.0.0 in Cygwin?

This PR worked fine when made against the old `stable-3.4` branch. But now that I have rebased it on version 4, it fails when building the package, see e.g. https://github.com/semigroups/Semigroups/runs/5362272006?check_suite_focus=true:
```
[...omitted...]
mv -f src/.deps/libsemigroups_la-cong.Tpo src/.deps/libsemigroups_la-cong.Plo
mv -f src/.deps/libsemigroups_la-freeband.Tpo src/.deps/libsemigroups_la-freeband.Plo
make[1]: Leaving directory '/cygdrive/d/a/Semigroups/Semigroups/libsemigroups'
make: *** [Makefile:1292: bin/include/libsemigroups/libsemigroups.hpp] Error 2
Error: Process completed with exit code 2.
```
Can you make any sense of this/does it work for you?

james-d-mitchell:
Hey @wilfwilson any chance you can post the omitted bit?

james-d-mitchell:
Ah no wait I can see the omitted bit in the log. This is a bug in libsemigroups, which I thought I’d fixed already, I’ll try to post a fix soon!

wilfwilson:
No probs, hope it's an easy fix!

james-d-mitchell:
@wilfwilson I'm trying to setup a CI cygwin job for libsemigroups at:

https://github.com/libsemigroups/libsemigroups/runs/5374124738?check_suite_focus=true

If you have any insights into how to get it to work, I'd find this most useful!

james-d-mitchell:
@wilfwilson I've rebased this onto the most recent `stable-4.0` branch, but it doesn't seem to work. The previous issues are resolved, but something else is going on. If you have the time, can you possibly take a look?

Thanks!

wilfwilson:
Some configuration stuff is missing from the `setup-gap` step: compare with the `setup-gap` step of the Unix job. In particular, you need to be cloning Digraphs; you might also need to do the `LDFLAGS` thing (I don't know).

wilfwilson:
I've added cloning Digraphs; let's see if it works... except that the tests will fail, because it tests against GAP master (I am unsure whether the Cygwin CI setup will work with a GAP version in the `stable-4.11` series or newer, but we could try).

wilfwilson:
It is unsurprisingly slow; but it is even slower than it needs to be, because Digraphs is being built twice; see https://github.com/gap-actions/setup-gap/issues/38.

wilfwilson:
Okay yeah, it failed because of GAP master breaking Semigroups. The Semigroups package took 28 minutes to build! So perhaps we wouldn't want to run this on every build of every PR.

james-d-mitchell:
This is incorporated into and superseded by #912, thanks @wilfwilson

wilfwilson:
You're welcome! Please you got something sorted.

Issue #788, opened 2021-06-14
Failing test in `tst/standard/semiffmat.tst` in GAP master branch


wilfwilson:
The GitHub Actions tests are failing because of this. This might be related to recent changes in GAP `master` branch to do with matrices (which I don't know the details of). I've narrowed it down in the later messages in this thread.

The `AsMonoid` bit is where something is going wrong:
```
gap> S := GLM(2,2);;
gap> AsMonoid(IsMatrixOverFiniteFieldMonoid, GF(4), S);;
gap> Size(Elements(S));
17
gap> Size(S);
16
```

wilfwilson:
The `AsMonoid` bit is where something is going wrong:
```
gap> S := GLM(2,2);;
gap> AsMonoid(IsMatrixOverFiniteFieldMonoid, GF(4), S);;
gap> Size(Elements(S));
17
gap> Size(S);
16
```

wilfwilson:
Narrowed it down a bit:
```
gap> S := GLM(2,2);;
gap> Length(EN_SEMI_AS_LIST(S));
16
gap> S := GLM(2,2);;
gap> List(GeneratorsOfSemigroup(S), x -> Matrix(GF(4), x));;
gap> Length(EN_SEMI_AS_LIST(S));
17
```
Surely the line `List(GeneratorsOfSemigroup(S), x -> Matrix(GF(4), x));;` shouldn't affect `S` at all?!

wilfwilson:
The GAP change which broke this test was https://github.com/gap-system/gap/commit/fd946c31c291f68617b5486a4f40e61eec7d5e57

wilfwilson:
The specific problematic change is that this `TryNextMethod` was removed from this equality method for matrices. Presumably the bit below doesn't always work as we expect for our matrices:
<img width="1218" alt="Screenshot 2021-06-14 at 14 44 55" src="https://user-images.githubusercontent.com/8956868/121902327-1ed6b580-cd1f-11eb-8337-2ab13e0c5fd9.png">
In particular I think the `ConstructingFilter` bit maybe doesn't work...

wilfwilson:
Note that the `ConstructingFilter` of the first generator (which is the identity element of the monoid), which is a representation, changes:
```
gap> S := GLM(2,2);;
gap> List(GeneratorsOfSemigroup(S), x -> ConstructingFilter(x!.mat));
[ <Representation "IsPlistRep">, <Representation "IsGF2MatrixRep">,
[...]

PR #789, opened 2021-06-14
Copy the matrix argument given to `NewMatrixOverFiniteField`

wilfwilson:
`ImmutableMatrix` is a GAP function that can change its argument in place, even if it is given an immutable argument. In particular, `ImmutableMatrix` can change the representation of the matrix that it is given.

For us, this is dangerous since matrices in different representations are considered to be *not* equal in GAP, and they cannot in general be multiplied. Therefore, if a matrix that belongs to a semigroup has its representation changed in-place, then multiplication and equality testing no longer necessarily work correctly in that semigroup. This is what led to the issue in #788, where a semigroup of size 16 had 17 elements, because it had two copies of its identity in different representations.

An even safer version would be to change the line to:
```
  mat := Objectify(PlistMatrixOverFiniteFieldType,
                   rec(mat := MakeImmutable(List(list, ShallowCopy))));
```
i.e. to bypass `ImmutableMatrix` entirely.

### DIFF
--- a/gap/elements/ffmat.gi
+++ b/gap/elements/ffmat.gi
@@ -247,7 +247,7 @@ function(filter, field, list)
   fi;
 
   mat := Objectify(PlistMatrixOverFiniteFieldType,
-                   rec(mat := ImmutableMatrix(field, list)));
+                   rec(mat := ImmutableMatrix(field, List(list, ShallowCopy))));
   SetBaseDomain(mat, field);
 
   return mat;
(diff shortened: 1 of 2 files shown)


james-d-mitchell (approved):
I’m happy with this thanks for the fix! Seems a weird choice to not be allowed to multiply or compare matrices in different representations in GAP, I thought the whole point of representations was to have different implementations of mathematically equivalent objects, but what do I know...

wilfwilson:
I agree!

PR #790, opened 2021-10-17
Avoid using SIZEOF_VOID_P

fingolfin:
This allows future GAP versions to drop it (although 4.12 definitely will support it). As far as I could tell, `SYSTEM_IS_32_BIT` is not used anywhere, so I just removed it. Likewise, `SYS_IS_CYGWIN32` was unused, so I also removed it while at it.


james-d-mitchell:
Thanks for this too!

PR #792, opened 2021-10-17
Stop using obsolete autoconf macros

fingolfin:
This stops newer autoconf versions from complaining about them


james-d-mitchell:
Thanks @fingolfin !

PR #803, opened 2022-02-28
Avoid using `SIZEOF_VOID_P`

wilfwilson:
[This was added to `stable-3.4` in #790 but didn't make its way into 4.0.0].

From @fingolfin's commit:

> This allows future GAP versions to drop it (although 4.12 definitely will support it). As far as I could tell, `SYS_IS_CYGWIN32` was unused, so I removed it while at it.


james-d-mitchell:
@wilfwilson eeek! I did not notice that, again my bad, v4.0.1 here we come!

PR #804, opened 2022-02-28
Remove `doc/*.js` and `doc/*.css` from the repo

wilfwilson:
These files are auto-generated by GAPDoc when compiling the manual. The files are copied from the current version of GAPDoc.

The contents of `doc/manual.js` has recently changed in the latest version of GAPDoc. The new version fixes a bug that existed in the old version. See https://github.com/frankluebeck/GAPDoc/issues/46 for more info.

This therefore means that there may be a mismatch between the installed GAPDoc version, and the version of that file that exists in the repository. For me, using the latest version of GAPDoc, this means that every time I recompile the manual, I have to run `git checkout doc/manual.js` to remove the change.

This is annoying.

To resolve this and similar issues, the easiest thing to do is just to remove these files from the repository.

Almost removes `doc/chooser.html`, which is similarly copied over.

(See also https://github.com/digraphs/Digraphs/pull/511).


james-d-mitchell:
Thanks @wilfwilson makes complete sense to me!

PR #810, opened 2022-03-04
doc: allow manual compile without all deps

james-d-mitchell:


james-d-mitchell:
I think this resolves Issue #810, @wilfwilson can you please confirm?

james-d-mitchell:
The package (against the `master` branch of GAP) is completely broken by some changes in GAP:

[Recent change to `Matrix` causes `Semigroups` package CI to break](https://github.com/gap-system/gap/issues/4814)

makedoc.g
```diff
@@ -26,7 +25,11 @@ UrlEntity := function(name, url)
 end;
 
 PackageEntity := function(name)
-  return UrlEntity(name, PackageInfo(name)[1].PackageWWWHome);
+  if LoadPackage(name, false) <> fail then
```
wilfwilson, line 28: Minor suggestion, you could just check whether the package is (available to be) loaded, rather than trying to actually load it – this might have a marginally higher success rate:
```suggestion
  if TestPackageAvailability(name) <> fail then
```

wilfwilson (approved):
Yes, I can confirm that this still allows the manual to be compiled even when e.g. Digraphs is deleted.

makedoc.g
james-d-mitchell: Good idea, I had `IsPackageMarkedForLoading` before, which was not quite right, but didn't remember/know about `TestPackageAvailability`.

PR #815, opened 2022-03-15
Better congruence lattices

james-d-mitchell:
This PR properly integrates the code for better congruence lattices written in conjunction with @mariatsalakou and @marinaanagno.


james-d-mitchell:
@wilfwilson any idea why the github actions are failing with:

~~~
Run # optionally: clone specific package versions, in case we need to test
fatal: destination path 'digraphs' already exists and is not an empty directory.
Error: Process completed with exit code 128.
~~~
I don't think I changed anything related to this in this PR.

wilfwilson:
I think this is https://github.com/gap-actions/setup-gap/issues/38.

The script isn't successfully deleting an existing package.

Previously this was just annoying, because the installed package was in a directory called like `digraphs-1.5.0` or something, which doesn't stop up us cloning into the `digraphs` directory.

But presumably Digraphs is just called "digraphs" in the package distribution, so it _does_ already exist, and so we can't clone into it. So https://github.com/gap-actions/setup-gap/issues/38 needs to be fixed.

PR #819, opened 2022-03-21
Improve compatibility with GAP's MatrixObj project

fingolfin:
... by using TryNextMethod in Matrix() methods instead of raising an error

Together with https://github.com/gap-system/gap/pull/4833 by @hulpke this should resolve the conflict between GAP master and semigroups. So if we this patch is acceptable for you, then as soon as there is a semigroups release with it, we could merge that GAP PR and finally release GAP 4.12 -- well, after some more testing, obviously! But thanks to the work done on https://github.com/gap-system/PackageDistro we will also (again) get tests of Semigroups (and all other distributed GAP packages) against GAP master, so we can deal with those, too

### DIFF
--- a/gap/elements/semiringmat.gi
+++ b/gap/elements/semiringmat.gi
@@ -189,9 +189,9 @@ function(filter, mat, threshold, period)
   local checker, row;
 
   if not IsRectangularTable(mat) or Length(mat) <> Length(mat[1]) then
-    ErrorNoReturn("the 2nd argument must define a square matrix");
+    TryNextMethod();
   elif filter <> IsNTPMatrix then
-    ErrorNoReturn("cannot create a matrix from the given arguments");
+    TryNextMethod();
   fi;
 
   checker := SEMIGROUPS_MatrixOverSemiringEntryCheckerCons(filter,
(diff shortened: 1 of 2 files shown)


fingolfin:
I've discovered a few more places that need `TryNextMethod`, and adjusted some of the tests. 

But I think I'll also have to disable a few tests, simply because the behaviour in 4.10 / 4.11 will be different from that in 4.12. And arguably some bugs ought to be fixed in GAP master, too... but one step after the other...

fingolfin:
@james-d-mitchell I've updated this again (sorry for taking so long). Assuming this PR passes tests, *and* you are OK with it, I think we'd have to proceed as follows:
1. merge this PR
2. release semigroups 4.0.1 with it
3. merge that update into the GAP package distro (easy :-) )
4. merge https://github.com/gap-system/gap/pull/4833
5. once all that works, we can remove some of the "workarounds" in this PR again (they mostly exist to solve the chicken-and-egg problem between merging this PR first vs. merging that GAP PR first)

james-d-mitchell:
Thanks @fingolfin I'm happy to merge this, and with the rest of the steps too. I'm a wee bit confused why there are 3 commits, when 2 of them are in master already. I tried rebasing and pushing to your branch but it says it's up to date?

fingolfin:
Oh, sorry: this PR is targeting `master`, but perhaps it should be targeting `stable-4.0`? I can change that, let me know

james-d-mitchell:
Either branch should be ok, maybe rebasing on the present master will be sufficient?

fingolfin:
Changing the base branch is possible with a button press from my phone, so I did that now :-)

PR #820, opened 2022-03-25
Remove spurious dependency on libgmp

fingolfin:
There is no reason to include gmp.h here. This makes compiling Semigroups a tad easier for e.g. people who have GMP installed via Homebrew on an M1 mac (e.g. me :-) ) where this is not found without adding the right compiler flags manually to CPPFLAGS


james-d-mitchell:
Thanks @fingolfin, makes complete sense.

PR #821, opened 2022-04-19
Use ccache to speed up CI builds (and don't use `-march=native` by default)

fingolfin:
I noticed that the GitHub Actions CI jobs spend ~50% of their time (e.g. 6 minutes out of 12) compiling C/C++ code. This patch attempts to reduce that by combining the GH Actions caching feature with https://ccache.dev to reduce the compiler overhead to just a few seconds for most builds -- only builds were the C++ sources actually changed would take longer.

However this feature does not mesh well with `-march=native`, which anyway causes other problems e.g. for distro packagers. I've thus also removed this from the default FLAGS. If you think this goes to far, one could also try to modify `.github/workflows/ci.yml` to build w/o `-march=native`.

Note that the performance benefit (if any) of course only shows up once the caches are primed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,12 @@ jobs:
       - name: "Install dependencies"
         if: ${{ runner.os == 'macOS' }}
         run: brew install automake
+      # Setup ccache, to speed up repeated compilation of the same binaries
+      # (i.e., GAP and the packages)
+      - name: "Setup ccache"
+        uses: Chocobo1/setup-ccache-action@v1
+        with:
+          update_packager_index: false
       - name: "Install GAP and clone/compile necessary packages"
         uses: gap-actions/setup-gap@v2
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,7 @@ jobs:
         uses: gap-actions/build-pkg@v1
         with:
           ABI: ${{ matrix.ABI }}
+          CONFIGFLAGS: --disable-hpcombi
       - name: "Run tst/teststandard.g"
         uses: gap-actions/run-pkg-tests@v2
         with:
--- a/configure.ac
+++ b/configure.ac
@@ -82,17 +82,6 @@ AC_MSG_RESULT([$enable_debug])
 
 AM_CONDITIONAL([KERNEL_DEBUG], [test "x$enable_debug" = xyes])
 
-AC_ARG_WITH([march-native],
-            [AS_HELP_STRING([--without-march-native], 
-                            [do not use compile flag -march=native even if available])]
-           )
-
-AS_IF([test "x$with_march_native" == "xno" ],
-      [AC_MSG_NOTICE([the compile flag -march=native will not be used even if available])])
-
-AS_IF([test "x$with_march_native" != "xno"],
-      [AX_CHECK_COMPILE_FLAG(-march=native, AX_APPEND_FLAG(-march=native))])
-
 # Check if HPCombi is enable, and available
 AX_CHECK_HPCOMBI()
 
--- a/m4/ax_check_hpcombi.m4
+++ b/m4/ax_check_hpcombi.m4
@@ -6,11 +6,11 @@ dnl to handle the other checks as to whether or not HPCombi can be used.
 AC_DEFUN([AX_CHECK_HPCOMBI], [
   m4_define([ax_hpcombi_cxxflags_variable],[HPCOMBI_CXXFLAGS])
 
-  dnl # Check if the flags required for HPCombi are supported 
-  AX_CHECK_COMPILE_FLAG(-march=native, 
-                        AX_APPEND_FLAG(-march=native,
+  dnl # Check if the flags required for HPCombi are supported
+  AX_CHECK_COMPILE_FLAG(-march=avx,
+                        AX_APPEND_FLAG(-march=avx,
                                        [ax_hpcombi_cxxflags_variable]),
-                        [AC_MSG_WARN([flag -march=native not supported])
+                        [AC_MSG_WARN([flag -march=avx not supported])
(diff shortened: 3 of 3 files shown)


james-d-mitchell:
Thanks @fingolfin, I'm happy to merge this, except that I'm not completely sure that it actually does what you intended re: `-march=native`. There's also a check in `m4/ax_check_hpcombi.m4` and it appears that this also appends the flag `-march=native` by default (it's required I think for `HPCombi` to work). A quick fix is just to switch using `HPCombi` off by default?

james-d-mitchell:
Looks like the flag is still getting passed:
~~~
/bin/bash ./libtool  --tag=CXX   --mode=compile g++ -DHAVE_CONFIG_H -I. -I./config  -DNDEBUG -DHPCOMBI_CONSTEXPR_FUN_ARGS    -I/home/runner/work/Semigroups/Semigroups/libsemigroups/include -I/home/runner/work/Semigroups/Semigroups/libsemigroups/extern -std=gnu++14 -O3 -Wall -Wextra   -I./extern/eigen-3.3.9/ -march=native -flax-vector-conversions   --coverage -MT src/libsemigroups_la-bipart.lo -MD -MP -MF src/.deps/libsemigroups_la-bipart.Tpo -c -o src/libsemigroups_la-bipart.lo `test -f 'src/bipart.cpp' || echo './'`src/bipart.cpp
~~~

james-d-mitchell:
One fix is to just add the flag `--disable-hpcombi` to the gh-action, this is certainly the quickest fix.

fingolfin:
Indeed. I just see that the `HPCombi` actually adds `-march=native`, and then the same also is done in the `libsemigroups` build system :-(. But I don't understand why `HPCombi` would need these? As I understand, it is about using SSE and AVX, so shouldn't something like `-mavx` or  `-mavx2` or so be sufficient?

fingolfin:
Indeed looking at the HPCombi repository, I see `check_cxx_compiler_flag('-mavx' HPCOMBI_HAVE_FLAG_AVX)` used in https://github.com/hivert/HPCombi/blob/master/CMakeLists.txt and the line with `-march=native` is commented out

fingolfin:
Unfortunately still shows `-march=native` because `libsemigroups` also adds it. I don't know how I to properly pass `--disable-hpcombi` on to it, though. Any hints?

james-d-mitchell:
@fingolfin, I'm pretty sure if you specify `--disable-hpcombi` to the Semigroups package's `configure`, then this gets passed onto libsemigroups `configure` and then HPCombi will be disabled. If that doesn't work, we can look at releasing a new version of libsemigroups which drops `-march=native` too.

fingolfin:
Indeed, it is being passed on... However, the problem is that the `AX_CHECK_HPCOMBI` macro still performs the test for `-march=native` and adds the flag. It would need to be adjusted to first check `enable_hpcombi` and then do nothing if it is set to `no`. The same patch then also should be applied to `libsemigroups`. And ideally both should set `-march=avx` instead.... *sigh*

Anyway, I made https://github.com/libsemigroups/libsemigroups/pull/345 but it only does the  `-march=avx`, but did not yet do anything about handling `enable_hpcombi=no` right

james-d-mitchell:
Thanks for this and sorry for the trouble! I’m happy to help, in the morning!

fingolfin:
OK, this *should* be OK now. Of course if/when libsemigroups gets a release that uses `-march=avx` instead of `-march=native` then we can drop the `--disable-hpcombi` here again.

fingolfin:
Unfortunately it doesn't seem to actually work (the `ccache` bit, I mean; I think the other changes were useful on their own, though): Looking at the most recent GH Actions log for `master`, I see the message [Cache not found...](https://github.com/semigroups/Semigroups/runs/6109782792?check_suite_focus=true#step:4:68). Which is weird, as a basically identical setup works fine for the GAP PackageDistribution repository. Huh.

fingolfin:
OK maybe this was actually the first build with ccache on `master`, which would explain that. And on e.g. https://github.com/semigroups/Semigroups/runs/6109047489?check_suite_focus=true#step:6:401 a cache *is* found. But it has a very low cache hit rate at the end of the job. So something still seems weird.. Hrm

james-d-mitchell:
Thanks for the comments @fingolfin, interestingly in the "Cache not found..." log linked about the "hit rate" is still ~5% in the final stats. Perhaps more relevantly, we've been using ccache to speed up the libsemigroups builds for a while, and what we are seeing here more or less mirrors this (i.e. it doesn't always speed things up, often the cache hit rate is rather lower than might be expected). I never really had the time to investigate what was going on.

Is one possible issue that all 4 of the jobs are trying to save to the same cache, meaning that only one of them is really preserved, and hence the other 3 almost always "miss"? I don't really have much insight...

fingolfin:
That could actually be. In GAP I've for various reasons "split" the caches and it seems to work well / better. Perhaps something to try for you, too? This is the syntax we used:
```
      - name: "Setup ccache"
        if: ${{ runner.os != 'Windows' }}
        uses: Chocobo1/setup-ccache-action@v1
        with:
          update_packager_index: false
          override_cache_key: ${{ runner.os }}-${{ hashFiles('extra.flags') }}-${{ github.ref }}
          override_cache_key_fallback: ${{ runner.os }}-${{ hashFiles('extra.flags') }}

```

Clearly you'll want to adjust this a bit. Maybe this would work:
```
      - name: "Setup ccache"
        if: ${{ runner.os != 'Windows' }}
        uses: Chocobo1/setup-ccache-action@v1
        with:
          update_packager_index: false
          override_cache_key: ${{ runner.os }}-${{ matrix.gap-branch }}-${{ matrix.ABI }}-${{ github.ref }}
          override_cache_key_fallback: ${{ runner.os }}-${{ matrix.gap-branch }}-${{ matrix.ABI }}

```

PR #825, opened 2022-04-21
DO NOT MERGE: experiments with ccache

fingolfin:


fingolfin:
Seems it just needed some priming for the caches; now build times are indeed down.

PR #826, opened 2022-04-26
Fix matrix

james-d-mitchell:
This PR reenables testing against GAP master, and removes the upranking of methods in `elements/semiringmat.gi` introduced in c03fe2f5. Related to the discussion:

https://github.com/gap-system/gap/issues/4878


james-d-mitchell:
Subsumed (?) by #827

Issue #807, opened 2022-03-02
`Matrix(Integer,[[1]])` returns a `IsIntegerMatrix` but does not implement the `MatrixObj` protocol


fingolfin:
The following happens with both Semigroups 3.4.2 and 4.0.0:

Without it:
```gap
gap> Matrix(Integers, [[1,2],[3,4]]);
<2x2-matrix over Integers>
gap> BaseDomain(last);
Integers
```
With semigroups loaded:
```
gap> Matrix(Integers, [[1,2],[3,4]]);
Matrix(IsIntegerMatrix, [[1, 2], [3, 4]])
gap> BaseDomain(last);
Error, no method found!
```

There are two options to resolve this:
1. implement `BaseDomain` and possibly other methods needed to make `IsIntegerMatrix` compliant with the `MatrixObj` interface (to the best of  your ability -- we definitely need to improve the docs, and add conformance tests)
2. don't do this, but then also don't change the behaviour of the `Matrix(Integers, M)` constructor

Observed in https://github.com/gap-system/gap/pull/4805

wilfwilson:
This is related to #512.

wilfwilson:
And I did some unfinished work in this direction in #621.

fingolfin:
In this particular case, though, just adding a `BaseDomain` method for `IsIntegerMatrix` might already help, and shouldn't be hard or cause problems for you, should it? (Granted, I haven't looked at the code at all, so maybe I am mistaken here).

james-d-mitchell:
I’m happy enough to try to resolve this, but would like to point out that I think this is an issue caused by changes in GAP rather than any change in Semigroups. In particular If I recall the code for creating integer matrices using ‘Matrix’ in Semigroups predates the code in GAP. In some sense this is the GAP library changing the behaviour of Semigroups rather than the other way around. This is just for the record, it’s undoubtedly better to have less code for matrices in Semigroups and more in the GAP library, and I’m happy to make whatever changes might be required.

fingolfin:
> I’m happy enough to try to resolve this

Thanks!

> [...], but would like to point out that I think this is an issue caused by changes in GAP rather than any change in Semigroups.

You are 100% right, and I misrepresented this above (not intentionally, but still); in fact earlier today I spent digging through the issue & PR @wilfwilson linked to, re-reading things that I'd forgotten about (including my own comments 😅). It's an annoyingly complex and far too long-unresolved storey 😞 

[...]

PR #827, opened 2022-04-28
Replace IsIntegerMatrix and IsMatrixOverFiniteField by MatrixObj

james-d-mitchell:
This PR so far contains changes that removes the code for integer matrices in the Semigroups package and replaces them with `MatrixObj`. This is WIP, and shouldn't be merged, at least in part because I think there's no released version of GAP that this will work with, and it's a backwards incompatible change.


james-d-mitchell:
I believe that this resolves the original issue raised in #807

james-d-mitchell:
I've adjusted this again to remove the changes to the matrices over finite fields code, and also to try to make this work with GAP master and 4.11 and 4.10. I've tested this with 4.11 and it works ok.

james-d-mitchell:
I have no replaced the code in Semigroups for matrices over finite fields, and I expect that the Semigroups package tests will pass with GAP master, not sure about the other versions of GAP.

james-d-mitchell:
@fingolfin @wilfwilson the CI seems to be broken again with a 404 for 

https://www.gap-system.org/pub/gap/gap4pkgs/packages-stable-4.10.tar.gz

in 

https://github.com/semigroups/Semigroups/runs/7041559173?check_suite_focus=true

wilfwilson:
That link seems to be resolving now.

fingolfin:
Broken CI was due to migration of the GAP website; I forgot to restore on HTTP redirect rule... fixed now

Now I've done the following: I took a GAP master build with the latest package distro, and replace semigroups in there with what is in this branch. I then run parts of the GAP test suite, after first loading Semigroups (I did not yet try "all" combinations).

This revealed a minor test regressions:
```
########> Diff in /Users/mhorn/Projekte/GAP/gap/tst/testinstall/magma.tst:54
# Input is:
AsSemigroup( F );
# Expected output:
<semigroup of size 6, with 2 generators>
# But found:
<semigroup of size 6, 2x2 matrices over GF(2) with 2 generators>
########
```
but that looks like something we can easily adjust in the GAP test suite (e.g. by surprising the print output; we can't just change the expected output, though, as it now differs depending on whether semigroups is loaded or not...)

I am now going to run GAP's `tst/teststandard.g` and `tst/testextra.g`

james-d-mitchell:
Thanks @fingolfin, we are running some GAP tests in the jobs on azure, but these are not being run here (they use GAP 4.10 and 4.11). I think the best solution to that change is to suppress the output, as you say.

james-d-mitchell:
Is there a gh action for running the GAP library tests anywhere?

fingolfin:
> Is there a gh action for running the GAP library tests anywhere?

No, but assuming you already have a running GAP, then normally you can run the (minimal) test suite via `Read(Filename(DirectoriesLibrary("tst"), "testinstall.g"));`. This will exit the current GAP with an exit code set to indicate success or failure.

james-d-mitchell:
Thanks @fingolfin, I think I'm just a little confused about how to do this in the context of the gh actions for GAP, since there's nowhere in the `ci.yml` file that we actually invoke GAP directly, just use one of the GAP gh actions. If you could point me in the right direction that'd be awesome.

I am nearly finished with this, I need to add a few more tests, and rebase the PR, but otherwise everything is looking good!

fingolfin:
@james-d-mitchell it should be possible to run `gap` by just invoking it as such in a bash shell script. So something like this *should* work:
```
      - name: "Run GAP testinstall test suite"
        run: gap -c 'Read(Filename(DirectoriesLibrary("tst"), "testinstall.g"));'
```

fingolfin:
Ah OK, I see you also got it to work by hacking `gap-actions/run-pkg-tests`. Also fine. Perhaps even better, to ensure the right version of `semigroups` is being loaded? Anyway, it seems to work.

james-d-mitchell:
Thanks @fingolfin, yup I figured it out, thanks for letting me know. 

I've tried as far as possible to not require any changes to the GAP tests for this version of Semigroups, but there are some I can't workaround (without doing too much work), so I'll likely submit a PR to GAP for this.

fingolfin:
Great. And of course if there are changes needed on the GAP side, that's fine (and kinda to be expected). Let me know if I can help with anything, or if a chat would help at any point.

james-d-mitchell:
@fingolfin I think the only changes required in GAP `master` are in the PR:

https://github.com/gap-system/gap/pull/4971

PR #834, opened 2022-05-25
semipperm: resolve issue 817

james-d-mitchell:
The issue seems to have been caused by there being mismatch in the
generators used in `SmallerDegreePartialPermRepresentation`, so that
`Factorization` used too large indices

### DIFF
--- a/gap/semigroups/semipperm.gi
+++ b/gap/semigroups/semipperm.gi
@@ -242,7 +242,7 @@ function(S)
   sigmainv, enum, orbits, cosets, stabpp, psi, rho, rhoinv, stab, nrcosets, j,
   reps, gen, offset, rep, box, subbox, T, map, inv, d, k, i, m;
 
-  oldgens := Generators(S);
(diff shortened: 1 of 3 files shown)


gap/semigroups/semipperm.gi
```diff
@@ -242,7 +242,7 @@ function(S)
   sigmainv, enum, orbits, cosets, stabpp, psi, rho, rhoinv, stab, nrcosets, j,
   reps, gen, offset, rep, box, subbox, T, map, inv, d, k, i, m;
 
-  oldgens := Generators(S);
```
wilfwilson, line 245: `Generators`?! Death sentence!

wilfwilson (approved):
Looks good @james-d-mitchell

gap/semigroups/semipperm.gi
james-d-mitchell: My thoughts exactly!

Issue #830, opened 2022-05-20
Improve ccache hit rate in CI


james-d-mitchell:
That could actually be. In GAP I've for various reasons "split" the caches and it seems to work well / better. Perhaps something to try for you, too? This is the syntax we used:
```
      - name: "Setup ccache"
        if: ${{ runner.os != 'Windows' }}
        uses: Chocobo1/setup-ccache-action@v1
        with:
          update_packager_index: false
          override_cache_key: ${{ runner.os }}-${{ hashFiles('extra.flags') }}-${{ github.ref }}
          override_cache_key_fallback: ${{ runner.os }}-${{ hashFiles('extra.flags') }}

```

Clearly you'll want to adjust this a bit. Maybe this would work:
```
      - name: "Setup ccache"
        if: ${{ runner.os != 'Windows' }}
        uses: Chocobo1/setup-ccache-action@v1
        with:
          update_packager_index: false
          override_cache_key: ${{ runner.os }}-${{ matrix.gap-branch }}-${{ matrix.ABI }}-${{ github.ref }}
          override_cache_key_fallback: ${{ runner.os }}-${{ matrix.gap-branch }}-${{ matrix.ABI }}

```

_Originally posted by @fingolfin in https://github.com/semigroups/Semigroups/issues/821#issuecomment-1108334585_

james-d-mitchell:
I'm opening @fingolfin comment in PR #821 as a new issue so it doesn't get buried...

PR #840, opened 2022-06-13
Improve ccache hit rate in CI

james-d-mitchell:
This is just a direct implementation of @fingolfin suggestion in:

https://github.com/semigroups/Semigroups/issues/830


james-d-mitchell:
This seems to be working, the overall times to run the github actions CI for the first and the second run are: 27m 6s and 16m 20s, with the build time for Semigroups going down from about 6 minutes to some tens of seconds.

PR #854, opened 2022-08-07
Fix GC bug in SET_ELM_PLIST2

fingolfin:
The inner plist is modified, so we must invoked `CHANGED_BAG` on it, not the outer one.

Or just use the safe function `ASS_LIST` instead of `SET_ELM_PLIST`...


james-d-mitchell:
Thanks @fingolfin !

PR #855, opened 2022-08-08
Fix gapbind14/demo

fingolfin:
Update `.gitignore` to allow adding the missing `Makefile.in` & `configure`, then update `Makefile.gappkg` to the latest version.

Also remove the added support for `.cpp` from `Makefile.gappkg`, for two reasons:
1. using an identical copy of the original `Makefile.gappkg` makes future updates easier (of course it is perfectly acceptable to submit a PR to GAP to add support for the `.cpp` extension into the master copy of `Makefile.gappkg`!)
2. referencing a relative path like `../src/gapbind14.cpp` in `KEXT_SOURCES` messed up the code which attempts to adapt paths so that all object files end up inside the `gen` directory. It is easier and safer to just copy the source file into the local `src` directory; and if one does that anyway, one may as well rename it on the fly

Of course I am guessing here as to what your `Makefile.in` and `Makefile.gappkg` were saying (I am guessing you had them but they were accidentally not checked in due to the `.gitignore` rule).

I note that the use of a relative path involving `..` messes up the logic for generating the location of the `.o` files inside `gen`, so we end up with `gapbind14/demo/src/gapbind14.o`. This could be fixed by instead

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,7 @@
 .libs
 /libsemigroups/
 Makefile
-Makefile.in
+/Makefile.in
 Transitions*
 _*.xml
 _AutoDocMainFile.xml
@@ -51,7 +51,7 @@ bin
 cnf
 config.log
 config.status
-configure
+/configure
 doc/*.css
 doc/*.js
 doc/main.aux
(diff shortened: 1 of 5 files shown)


james-d-mitchell:
Thanks for this too!

PR #857, opened 2022-08-08
Add a workaround for macOS code signing issues.

fingolfin:
Copying new builds of loadable modules, shared libraries or executable
over an existing older copy confuses macOS' code signing system, leading
to segfaults when trying to load the updated copy. As a workaround,
delete the old copy before copying the new file into place.


james-d-mitchell:
Amazing! Thank you! @fingolfin

PR #863, opened 2022-08-17
make IsDualSemigroupRep not a representation

### LINKED ISSUE #862: Bad implication to a representation
This should not be done:
```gap
InstallTrueMethod(IsDualSemigroupRep,
                  IsSemigroup and IsDualSemigroupElementCollection);
```
A representation should never be implied like this. See also https://github.com/gap-system/gap/pull/3006

flsmith:
Resolve #862


james-d-mitchell:
Nice thanks @flsmith

PR #873, opened 2022-09-26
Add support for Sims1 in libsemigroups

james-d-mitchell:


james-d-mitchell:
This requires some documentation, some more tests (probably), and the release of libsemigroups v2.3

PR #884, opened 2022-10-11
Some updates to improve the performance of `LatticeOfCongruences` and friends

james-d-mitchell:


james-d-mitchell:
I want to do some more benchmarks, then I'll merge this if nothing is highlighted there.

PR #887, opened 2022-11-02
m4/find_gap.m4: update

fingolfin:
Stop referencing unused GAP_LIBS, and balance 'checking for GAP architecture' message: The conclusion to the check was not printed and hence no trailing newline printed. So whatever configure message came next ended up starting on the wrong line.


james-d-mitchell:
Thanks @fingolfin much appreciated!

PR #893, opened 2022-11-16
Fix faulty usage of 0-simple RZMS translations code

flsmith:
Previously the code for computing bitranslations of a RZMS could be incorrectly be triggered by a RZMS defined with a non-regular sandwich matrix.


james-d-mitchell:
Any chance you could add a test that exhibits the incorrect behaviour? @flsmith

PR #906, opened 2023-02-11
Remove invalid MutableCopyMat invocation

fingolfin:
Applying to it a matrix row (= vector) is invalid. It also is not necessary in this case.


fingolfin:
Is there anything blocking merging this?

PR #923, opened 2023-06-14
display: add node labels and edge colors

james-d-mitchell:


james-d-mitchell:
This requires some cleaning up, and the following features:

* node labels argument (or option in a record)
* edge colours argument (or option in record)

james-d-mitchell:
Going to use the forthcoming Graphviz a package instead

PR #937, opened 2023-07-02
build: detect C++ stdlib

james-d-mitchell:
Hopefully this resolves #928. I can't test that it resolves the issue though, it works on my machine (a Mac M1), but it worked previously too. 

@dimpase are you able to check if this resolves the issue?

PR #946, opened 2023-09-10
use install-sh etc from GAP

dimpase:
To fix the problem mentioned in https://github.com/semigroups/Semigroups/pull/937#issuecomment-1616590034, which is present on other platforms, e.g. on Debian linux:
```
$ ./configure            
...
configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."
```

This works as follows:

1) autogen.sh takes the parameter from AC_CONFIG_AUX_DIR and creates the corresnding directiry, named config/

2) ./configure finds GAP location and populates config/ with links to the contents of GAP's cnf/ directory

Note that we set `config/` to be ignored in .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,5 +34,6 @@
 /gen/
 /gh-pages/
 /libsemigroups/
-/src/pkgconfig.h.in
+/src/pkgconfig.h.in*
 /src/semigroups-config.hpp
+/config/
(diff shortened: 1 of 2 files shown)


dimpase:
@fingolfin

fingolfin:
That won't work if GAP was installed using `make install`

dimpase:
Does GAP's `make install` install anything from `cnf/` somewhere?

fingolfin:
No, and it shouldn't install things like `config.guess`, `config.sub`, `install-sh` (distro packagers would complain about it and discard those files anyway).

dimpase:
OK.

PR #947, opened 2023-09-11
Work around autoconf 2.69 bugs

### LINKED ISSUE #943: ./configure unable to find install-sh, etc.
98ed44e84e3e198820a8c0463a5ee22e872b4159 by @fingolfin removed the file `Makefile.am`,  and `AM_INIT_AUTOMAKE` and other `AM_*` macros. As a result, `config/` does not get populated at all, not even created. And one gets
```
$ ./configure                                                                                                                                                   
checking for g++... no
checking for c++... c++
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether c++ accepts -g... yes
checking whether c++ supports C++14 features by default... yes
checking for GAP root directory... /home/dima/tmp/gap
checking for GAP architecture... x86_64-unknown-openbsd7.3-default64-kv9
configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."
```
after

```
$ ./autogen.sh                                      
+ dirname ./autogen.sh
+ autoreconf -vif .
autoreconf-2.69: Entering directory `.'
autoreconf-2.69: configure.ac: not using Gettext
autoreconf-2.69: running: aclocal --force 
autoreconf-2.69: configure.ac: tracing
autoreconf-2.69: configure.ac: adding subdirectory libsemigroups to autoreconf
autoreconf-2.69: Entering directory `libsemigroups'
autoreconf-2.69: running: aclocal --force -I m4
autoreconf-2.69: running: libtoolize --copy --force
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, `config'.
libtoolize: copying file `config/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIR, `m4'.
libtoolize: copying file `m4/libtool.m4'
libtoolize: copying file `m4/ltoptions.m4'
libtoolize: copying file `m4/ltsugar.m4'
libtoolize: copying file `m4/ltversion.m4'
libtoolize: copying file `m4/lt~obsolete.m4'
autoreconf-2.69: running: /usr/local/bin/autoconf-2.69 --force
autoreconf-2.69: running: /usr/local/bin/autoheader-2.69 --force
autoreconf-2.69: running: automake --add-missing --copy --force-missing
autoreconf-2.69: Leaving directory `libsemigroups'
autoreconf-2.69: configure.ac: not using Libtool
autoreconf-2.69: configure.ac: not using Automake
autoreconf-2.69: Leaving directory `.'
```

fingolfin:
Resolves #943. Alternative to #946 and hence closes #946

That issue only affects people working with the git version of Semigroups (because the release tarballs bundle the required files). So it is OK if we play a bit dirty, in this case, by invoking automake and touching install-sh. 

See also the comments inside the changes for further details.

Alternatives considered:
- just require autoconf >= 2.71 and otherwise refuse to run: excludes people who can't easily get a newer autoconf; plus there is no good clean way to check the autoconf version in a script other than to parse the output of `autoconf --version` which is not meant to be human readable
- include bundled copies of those files -- simple (we use it for GAP) but those files need to be regularly updated, and you'll get complaints about why they are missing ;-)
- copy from GAP or `SOMEPREFIX/share/automake*/config.sub` some other location: a bit fragile (none of those places may have a copy; and it is not even necessarily clear what SOMEPREFIX should be).

Granted, this approach also has drawbacks, e.g. it relies on `automake` really installing these files even though it complains ("error: no proper invocation of AM_INIT_AUTOMAKE was found"); also the "fgrep" is a bit of a hack (but the consequences are easy to control: false positive will result in a harmlos "touch", false negative at worst in the user reporting yet another version of autotools affected by the bug).

In the end, any of those solutions could have been made to work "good enough" for regular devs, I just hope this particular one will work for most "regular users who for some reason want to use a git version of Semigroups" (whoever they may be ... 😁 )

CC @dimpase

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -6,3 +6,20 @@
 # Requires GNU autoconf, GNU automake and GNU libtool.
 #
 autoreconf -vif `dirname "$0"`
+
+# autoconf 2.69 has a bug where autoreconf does not copy config.guess
+# and config.sub even though configure.ac needs them, unless automake is
+# being used. We work around this by forcing a call to automake if one
+# or both of config.sub and config.guess are missing.
+if ! test -x config.guess -a -x config.sub ; then
+  automake -acf 2> /dev/null || :
+fi
+
+# There is a second bug in autoconf 2.69 where the generated configure
+# script complains about install-sh not being there (even though it does
+# not actually need it). As a workaround, we just provide an empty file
+# instead. Since newer autoconf versions such as 2.71 are not affected
+# by the bug, we add a test to limit when this workaround is applied
+if fgrep -q ac_aux_dir/install-sh configure ; then
+  touch install-sh
+fi


fingolfin:
You need to use `gmake` not `make`

dimpase (approved):
LGTM. tested on Linux

james-d-mitchell:
> You need to use `gmake` not `make`

I noticed that when trying to compile `GAP` with `make` rather than `gmake` I get a helpful message saying this, is it straightforward to add this mechanism to `Semigroups`?

fingolfin:
In principle yes: you need to rename Makefile.in to (IIRC, just check
gap for the precise name) GNUMakefile.in (and adjust anything referencing it accordingly) and then copy the Makefile from GAP (need to adjust .gitignore, too)

Issue #948, opened 2023-09-11
Possible memory leak


james-d-mitchell:
OK, ran tests number of times, with both this and the old branch. The new branch doesn't really display a different behaviour (crashes don't happen during every run, and anyway it's a memory corruption/leak that finally brings GAP down sometimes).

Leaking/broken test is apparently `semigroups/tst/standard/semigroups/semidp.tst` and `semigroups/tst/standard/semigroups/semiquo.tst`, possibly more, only when run with `acting methods disabled`.

So here what all the logs I made have in common:
```
...
testing: /home/dima/tmp/gap/pkg/semigroups/tst/standard/tools/io.tst
    2438 ms (1711 ms GC) and 7.58MB allocated for tools/io.tst
-----------------------------------
total   1072164 ms (115772 ms GC) and 17.0GB allocated
              0 failures in 72 files

ESC[1m#I  Running tests with acting methods disabled
ESC[0mArchitecture: x86_64-unknown-openbsd7.3-default64-kv9

testing: /home/dima/tmp/gap/pkg/semigroups/tst/standard/attributes/acting.tst
    2281 ms (1625 ms GC) and 47.6MB allocated for attributes/acting.tst
...
testing: /home/dima/tmp/gap/pkg/semigroups/tst/standard/semigroups/semicons.ts\
t
   22820 ms (1734 ms GC) and 105MB allocated for semigroups/semicons.tst
testing: /home/dima/tmp/gap/pkg/semigroups/tst/standard/semigroups/semidp.tst
######## > Diff in:
/home/dima/tmp/gap/pkg/semigroups/tst/standard/semigroups/semidp.tst:319
# Input is:
Size(D) = Size(T) ^ 3;
# Expected output:
ESC[30;42mtrueESC[0m
# But found:
ESC[30;41mError, std::bad_allocESC[0m
########
   10906 ms (1812 ms GC) and 109MB allocated for semigroups/semidp.tst
testing: /home/dima/tmp/gap/pkg/semigroups/tst/standard/semigroups/semieunit.t\
st
...
testing: /home/dima/tmp/gap/pkg/semigroups/tst/standard/semigroups/semiquo.tst
######## > Diff in:
/home/dima/tmp/gap/pkg/semigroups/tst/standard/semigroups/semiquo.tst:121
# Input is:
Size(S / cong);
# Expected output:
ESC[30;42m120ESC[0m
# But found:
ESC[30;41mError, std::bad_allocESC[0m
########
[...]

PR #951, opened 2023-09-12
kernel: trying fix mem leaks

james-d-mitchell:
This is a WIP and shouldn't be merged. It partially addresses #948


james-d-mitchell:
I think this now fully resolves #948, but want to complete some final checking before merging.

james-d-mitchell:
@dimapase I've now run `SemigroupsTestStandard()` multiple times on the openbsd machine, and everything seems to pass. I'll merge this in any case.

PR #954, opened 2023-09-13
set rpath for external libsemigroups

dimpase:
set rpath for external `libsemigroups`, so that `LD_LIBRARY_PATH` is not needed if `libsemigroups` is installed to a non-standard location


dimpase:
I've left `m4/ax_check_libsemigroups.m4` largely as it was, no refactoring. I only split to very long strings into two, and
added the necessary stuff.

james-d-mitchell:
Thanks @dimpase

james-d-mitchell (approved):
This seems to do exactly what is desired, happy to approve. @fingolfin could you please also have a quick look at this if you have the time?

m4/ax_check_libsemigroup.m4
```diff
@@ -15,7 +15,8 @@ AC_DEFUN([AX_CHECK_LIBSEMIGROUPS], [
                           [libsemigroups >= $REQUI_LIBSEMIGROUPS_VERSION],
 			  [need_included_libsemigroups=no],
 			  [need_included_libsemigroups=yes])],
-	[AC_MSG_NOTICE([ignoring flag --with-external-libsemigroups, the Semigroups configure file was created on a system without m4 macros for pkg-config available...])])
+	[AC_MSG_NOTICE([ignoring flag --with-external-libsemigroups, the Semigroups configure file
+                        was created on a system without m4 macros for pkg-config available...])])
```
fingolfin, line 19: I think this should be undone, it causes the error to change from
```
...
checking for awk... awk
configure: using included libsemigroups...
configure: error: libsemigroups is required, clone or download the repo from https://github.com/libsemigroups/libsemigroups into this directory
```
which is sensibly wrapped in my terminal, to
```
checking for awk... awk
configure: using included libsemigroups...
configure: error: libsemigroups is required, clone or download the repo from
                                  https://github.com/libsemigroups/libsemigroups into this directory
```
which is rather weird.

m4/ax_check_libsemigroup.m4
```diff
@@ -27,7 +28,8 @@ AC_DEFUN([AX_CHECK_LIBSEMIGROUPS], [
    		[AS_IF(
    		   [test -f libsemigroups/include/libsemigroups.hpp],
                    [],
-                   [AC_MSG_ERROR([libsemigroups is required, clone or download the repo from https://github.com/libsemigroups/libsemigroups into this directory])])])
+                   [AC_MSG_ERROR([libsemigroups is required, clone or download the repo from
+                                  https://github.com/libsemigroups/libsemigroups into this directory])])])
```
fingolfin, line 32: Should also be undone, IMHO

m4/ax_check_libsemigroup.m4
```diff
@@ -53,6 +55,8 @@ AC_DEFUN([AX_CHECK_LIBSEMIGROUPS], [
   else 
         LIBSEMIGROUPS_VERSION="$(pkg-config --modversion libsemigroups)"
 	AC_MSG_NOTICE([using external libsemigroups $LIBSEMIGROUPS_VERSION])
```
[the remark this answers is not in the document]
fingolfin: (Indeed, this whole file mixes tabs and spaces liberally. Probably best to ignore this for this PR)

m4/ax_check_libsemigroup.m4
```diff
@@ -53,6 +55,8 @@ AC_DEFUN([AX_CHECK_LIBSEMIGROUPS], [
...
+        PKG_CHECK_VAR([LIBSEMIGROUPS_RPATH], [libsemigroups], [libdir],
+              [AC_SUBST([LIBSEMIGROUPS_RPATH],[-Wl,-rpath,${LIBSEMIGROUPS_RPATH}])])
```
fingolfin, line 59: That's the core of the change. Looks OK to me.

Makefile.gappkg
```diff
@@ -138,7 +138,7 @@ gen/%.$(GAP_OBJEXT): %.s Makefile
 # build rule for linking all object files together into a kernel extension
 $(KEXT_SO): $(KEXT_OBJS)
 	@mkdir -p $(@D)
-	$(QUIET_GAC)$(GAP_CXX) -o $@ $(GAP_LDFLAGS) $(GAC_LDFLAGS) $(KEXT_OBJS) $(KEXT_LDFLAGS)
+	$(QUIET_GAC)$(GAP_CXX) -o $@ $(GAP_LDFLAGS) $(GAC_LDFLAGS) $(KEXT_OBJS) $(KEXT_LDFLAGS) $(LIBSEMIGROUPS_RPATH)
```
fingolfin, line 141: I'd say this should rather be added to `KEXT_LDFLAGS`, that'll be better in the long run. But for now, either works *shrug*

fingolfin (commented):
Thanks! The principle seems fine to me, but I have some minor remarks on details.

james-d-mitchell:
Thanks @fingolfin, @dimpase I just rebased onto `main` I can make Max's suggested changes if you want?

m4/ax_check_libsemigroup.m4
james-d-mitchell: Agreed, I think this change should be undone.

dimpase: There are no two equal terminals, besides, such long lines of code aren't helping.
I can try to make sure the output is formatted better.

dimpase:
> Thanks @fingolfin, @dimpase I just rebased onto `main` I can make Max's suggested changes if you want?

It's up to you. I'd rather move the 2nd lines of these long error messages a bit to the left,
to reduce the unfortunate indentation.

dimpase:
You can also wrap the messages in `m4_normalize()`, this will remove any new lines and extra spaces/tabs, I think - but keep short lines in the source.

fingolfin:
which terminals do you know that do not wrap long too lines?

In any case, I am sure the message can be improved, but the current form in this PR isn't an improvement anywhere: on terminals that do wrap around it is worse, on terminals that don't it would also be worse, as they still couldn't see most of the second line.

m4/ax_check_libsemigroup.m4
fingolfin: Have you tested the message? It was perfectly legible for me before your changes but not anymore afterwards. Was the experience different for you?

dimpase:
I think with the change I just pushed they will be again printed as one line - but split in the source.

m4/ax_check_libsemigroup.m4
dimpase: I changed it to avoid extra-long lines in the source, most of all.

fingolfin (approved):
Looks good to me now, thanks!

PR #970, opened 2024-01-05
Release candidate for v5.3.3

james-d-mitchell:


james-d-mitchell:
The GAP tests fail here because of some bad tests in:

https://github.com/gap-system/gap/issues/5561

Think I'll make the release of Semigroups anyway

Issue #964, opened 2023-10-18
Implement `FreeSemilattice` constructor


james-d-mitchell:
In the same vein as `RectangularBand` etc.

PR #971, opened 2024-01-10
Free semilattice

james-d-mitchell:
This is the beginning of a resolution of #964, nothing much is actually implemented in this PR, and it shouldn't be merged.


james-d-mitchell:
Superseded by #995

PR #976, opened 2024-01-12
buildsys: unify rpath handling

fingolfin:
... between using a bundled and an external libsemigroups


james-d-mitchell:
Thanks @fingolfin !

PR #978, opened 2024-01-22
Fix for `HPCombi`

james-d-mitchell:


james-d-mitchell:
@fingolfin @ChrisJefferson is seems that a few days ago the cygwin job started to fail with:

~~~
Error, LOAD_DYN: failed to load kernel module /tmp/gaproot/pkg/Semigroups//bin/x86_64\
-pc-cygwin-default64-kv9/semigroups.so, No such file or directory in
  LOAD_DYN( filename ) at /home/runneradmin/gap/lib/files.gd:619 called from 
~~~

see, for example:

https://github.com/semigroups/Semigroups/actions/runs/7625232891/job/20769065765

Earlier in the log it can be seen that the executable `/tmp/gaproot/pkg/Semigroups//bin/x86_64\
-pc-cygwin-default64-kv9/semigroups.so` is built. This is not caused by anything in this PR (I don't think) since it appears to have started yesterday in the cron job:

https://github.com/semigroups/Semigroups/actions/runs/7606076562

If either of you have any ideas how to resolve this I'd be most grateful to hear about it.

ChrisJefferson:
We always had to do some slightly special things to get files in the right place. I will investigate and get back to you.

PR #981, opened 2024-01-24
Set up Dependabot to update GitHub Actions

fingolfin:
I've noticed warnings at e.g. https://github.com/semigroups/Semigroups/actions/runs/7628515256 node12 versus nod16 etc. etc. due to the use of the outdated `codecov/codecov-action@v2`. Using Dependabot helps avoid this by ensuring GH Actions are always used in their most up-to-date version.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


james-d-mitchell:
Thanks, I also tried to add this for Digraphs I think but simply adding the file didn't seem to enable anything (certainly no PR's have been opened by dependabot), do I have to enable something else too?

fingolfin:
I don't recall doing anything special for GAP, but perhaps something needs to be enabled in https://github.com/semigroups/Semigroups/settings/security_analysis ?

fingolfin:
Also nothing will happen before this PR is merged

james-d-mitchell:
Thanks @fingolfin, I can't see anything to turn at the link you mention. Hopefully this'll just work!

james-d-mitchell:
Seems to have worked, I wonder if the difference is that this PR targets `main` and I merge the dependabot file into `stable-1.6` of Digraphs.

fingolfin:
Ah, quite likely

PR #999, opened 2024-02-13
Fix bug in `IrredundantGeneratingSubset`

reiniscirpons:
Running `IrredundantGeneratingSubset` with a set of 2 generators of a cyclic semigroup currently causes issues, e.g.
```
gap> IrredundantGeneratingSubset([Transformation([2, 1]), IdentityTransformation]);
Error, Usage: cannot create a semigroup with no generators ...
gap> IrredundantGeneratingSubset([Transformation([ 1, 1, 3, 1 ]), Transformation([3, 3, 1, 3 ])]);
Error, Usage: cannot create a semigroup with no generators ...
```
What seems to happen is that the generating set gets reduced to only have 1 generator, but for some reason the check for irredundancy still gets carried out, meaning that we attempt to generate a semigroup with empty generating set, hence the error.

My fix adds a condition in the if statement that causes the error. Not sure this addresses the underlying issue with the stopping logic of the function, but it does fix the error for the two examples above, happy for any further suggestions for fixes.

### DIFF
--- a/gap/attributes/attr.gi
+++ b/gap/attributes/attr.gi
@@ -243,7 +243,7 @@ function(coll)
     fi;
 
     if not x in redund and not x in out then
-      if x in Semigroup(Difference(gens, [x])) then
+      if Length(gens) > 1 and x in Semigroup(Difference(gens, [x])) then
         AddSet(redund, x);
         gens := Difference(gens, [x]);
       else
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Nice one thanks @reiniscirpons

PR #1004, opened 2024-03-08
Workaround for GAP 4.12 test failure due to new GAPDoc version

fingolfin:


fingolfin:
Hmm, I think there are more places that need to be adjusted...

Perhaps a better approach is to modify `ci/run-gap-testinstall.g` to delete the offending .tst file

fingolfin:
OK that should be better, I think.

fingolfin:
@james-d-mitchell seems to work

james-d-mitchell:
Thanks @fingolfin sorry I didn't have time to do it myself. I'll merge and make a new release as soon as I can

PR #1015, opened 2024-05-23
Fix stable-4.12

ChrisJefferson:
This will (hopefully) fix stable-4.12 on macs (by using 'brew' to install gmp instead of using the internal one because it's too old).

I also pulled in #1012 , as I need that for CI to pass, and updated my website at the same time (as St Andrews shut all the cs.host websites)


ChrisJefferson:
OK, this proved harder to fix, because GAP 4.12 won't go look for homebrew gmp (later versions do).

At this point, we could write a quick special script, just for stable-4.12 macos, to get the right GMP, but before doing that, I wanted to check if you'd be happy instead just switching to checking master and stable-4.13, and dropping testing 4.12, or only testing 4.12 on ubuntu?s

james-d-mitchell:
Thanks @ChrisJefferson, I think just not testing GAP 4.12.2 on macosx is the best solution, thanks so much for the work you did on this!

james-d-mitchell:
I merged #1012, and then cherry-picked your commits onto stable-5.3, will auto merge this. Thanks @ChrisJefferson

james-d-mitchell (approved):
Looks good.

PR #1037, opened 2024-08-31
Update for recent gaplint

james-d-mitchell:


james-d-mitchell:
No idea why the CI is failing when it passed before, and nothing in this PR changes anything related to the failure. Seems to be that in GAP 4.12 is Digraphs 1.6.1 but Semigroups requires Digraphs 1.6.2.

PR #1045, opened 2024-11-07
Use GAP kernel helper IS_STRING_REP to test for strings

fingolfin:
There are actually six TNUMs indicating strings.

I don't know which branch you want this to land in. I based this branch
on tag v5.4.0 so you should be able to re-target this PR if it should
go into `stable-5.4` instead of `main`.


james-d-mitchell:
Thanks @fingolfin much appreciated!

james-d-mitchell (approved):

PR #1054, opened 2025-03-26
PermuteMultiplicationTable, refactor OnMultiplicationTable, add docs

pramothragavan:
This PR
- introduces a new function ``PermuteMultiplicationTable(temp, table, p)``, implemented in C, which permutes a multiplication table ``table`` and stores the result by changing ``temp`` in-place.
- refactors ``OnMultiplicationTable`` to call ``PermuteMultiplicationTable``.

I wasn't sure where to put the function, so I added a new file.

### DIFF
--- a/doc/isomorph.xml
+++ b/doc/isomorph.xml
@@ -11,16 +11,20 @@
 <#GAPDoc Label="OnMultiplicationTable">
 <ManSection>
   <Oper Name = "OnMultiplicationTable" Arg = "table, p"/>
-  <Returns>
-    A multiplication table.
-  </Returns>
+  <Oper Name = "PermuteMultiplicationTable" Arg = "temp, table, p"/>
   <Description>
     If <A>table</A> is a multiplication table of a semigroup and the second
     argument <A>p</A> is a permutation of
-    <C>[1 .. Length(<A>table</A>)]</C>, then this operation returns a
+    <C>[1 .. Length(<A>table</A>)]</C>, then <C>OnMultiplicationTable</C> returns a
     multiplication table of a semigroup isomorphic to that defined by
     <A>table</A> where the elements <C>[1 .. Length(<A>table</A>)]</C> are
-    relabelled according to <A>p</A>.
+    relabelled according to <A>p</A>.<P/>
+
+    <C>PermuteMultiplicationTable</C> works similarly but takes an additional argument <A>temp</A>, 
+    which should be a list of mutable lists of the same dimensions as <A>table</A>.
+    The argument <A>temp</A> is modified in-place to store the multiplication table that would be given by 
+    <C>OnMultiplicationTable(<A>table</A>, <A>p</A>)</C>, while the argument <A>table</A> remains unchanged.
+    This approach is preferred over <C>OnMultiplicationTable</C> when the user does not need to create a separate result variable.
 
     <Example><![CDATA[
 gap> table := [[1, 1, 3, 4, 5, 6, 7, 8],
@@ -36,7 +40,12 @@ gap> OnMultiplicationTable(table, p);
 [ [ 1, 2, 4, 4, 5, 6, 7, 8 ], [ 1, 2, 2, 4, 5, 6, 7, 8 ],
   [ 1, 2, 2, 4, 5, 6, 7, 8 ], [ 1, 2, 2, 4, 5, 6, 7, 8 ],
   [ 7, 5, 5, 6, 5, 6, 7, 8 ], [ 7, 5, 5, 6, 5, 6, 7, 8 ],
-  [ 7, 5, 6, 6, 5, 6, 7, 8 ], [ 7, 5, 6, 6, 5, 6, 7, 8 ] ]]]></Example>
+  [ 7, 5, 6, 6, 5, 6, 7, 8 ], [ 7, 5, 6, 6, 5, 6, 7, 8 ] ]
+gap> temp := List([1 .. Size(table)], x -> [1 .. Size(table)]);;
+gap> PermuteMultiplicationTable(temp, table, p);;
+gap> temp = OnMultiplicationTable(table, p);
+true]]>
+  </Example>
   </Description>
 </ManSection>
 <#/GAPDoc>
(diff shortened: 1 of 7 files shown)


src/isomorph.cpp
```diff
@@ -0,0 +1,61 @@
+#include "gap_all.h"
+#include "isomorph.hpp"
+
+
+Obj PermuteMultiplicationTable(Obj self, Obj temp, Obj M, Obj p) {
+    int n = LEN_LIST(M);
```
james-d-mitchell, line 6: What if `temp` isn't a list? I think you need to check `IS_LIST` and give an error if not.

src/isomorph.cpp
```diff
@@ -0,0 +1,61 @@
...
+    }
+  
+    int p_type = TNUM_OBJ(p);
+    if (p_type == T_PERM2) {
+        int deg_p = DEG_PERM2(p);
+        Obj q = STOREDINV_PERM(p);
+        if (q == 0) {
+            q = INV(p);
+            SET_STOREDINV_PERM(p, q);
+          }
```
james-d-mitchell, line 28: Weird formatting

src/isomorph.cpp
```diff
@@ -0,0 +1,61 @@
...
+        }
+      } else {
+        int deg_p = DEG_PERM4(p);
+        Obj q = STOREDINV_PERM(p);
+        if (q == 0) {
+            q = INV(p);
+            SET_STOREDINV_PERM(p, q);
+          }
+        int deg_q = DEG_PERM4(q);
```
james-d-mitchell, line 47: Can you maybe add an assertion here that checks that `q` is a `PERM4`?

src/isomorph.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "isomorph.hpp"
+
+
+Obj PermuteMultiplicationTable(Obj self, Obj temp, Obj M, Obj p) {
+    int n = LEN_LIST(M);
+
+    if (LEN_LIST(temp) != n) {
+        ErrorQuit("the arguments <temp> and <table> must have the same dimensions", 0L, 0L);
+    }
+    for (int i = 1; i <= n; i++) {
```
james-d-mitchell, line 11: Also need to check that each item in `temp` and `M` are satisfy `IS_LIST`

src/isomorph.cpp
```diff
@@ -0,0 +1,61 @@
...
+        if (q == 0) {
+            q = INV(p);
+            SET_STOREDINV_PERM(p, q);
+          }
+        int deg_q = DEG_PERM2(q);
+        for (int i = 1; i <= n; i++) {
+          Obj row = ELM_LIST(temp, i);
+          int ii = IMAGE(i - 1, CONST_ADDR_PERM2(q), deg_q) + 1;
+      
+          for (int j = 1; j <= n; j++) {
```
james-d-mitchell, line 34: You probably need to check that the elements in the lists are integers too

src/isomorph.cpp
```diff
@@ -0,0 +1,105 @@
...
+  if (p_type == T_PERM2) {
+    int deg_p = DEG_PERM2(p);
+    Obj q     = STOREDINV_PERM(p);
+    if (q == 0) {
+      q = INV(p);
+      SET_STOREDINV_PERM(p, q);
+    }
+    SEMIGROUPS_ASSERT(TNUM_OBJ(q) == T_PERM2);
+    int deg_q = DEG_PERM2(q);
+    for (int i = 1; i <= n; i++) {
```
james-d-mitchell, line 71: I'm a bit uncomfortable with the use of `int`s here, a quick look in the source code of GAP `permutat.cc` indicates that they are using `UInt` instead for similar variables. Perhaps these resolve to the same thing, but for consistency, can you please change `int` to `UInt` everywhere in this function?

src/isomorph.cpp
```diff
@@ -0,0 +1,105 @@
...
+      Obj row = ELM_LIST(temp, i);
+      int ii  = IMAGE(i - 1, CONST_ADDR_PERM2(q), deg_q) + 1;
+
+      for (int j = 1; j <= n; j++) {
+        int home    = Int_ObjInt(ELM_LIST(
```
james-d-mitchell, line 76:
```suggestion
        int home    = INT_INTOBJ(ELM_LIST(
```
and the same change elsewhere, it seems like there's no point in using `Int_ObjInt` but there is some overhead, so let's not.

src/isomorph.hpp
```diff
@@ -0,0 +1,35 @@
+//
+// Semigroups package for GAP
+// Copyright (C) 2016 James D. Mitchell
```
james-d-mitchell, line 3: Same change as above

tst/standard/attributes/isomorph.tst
```diff
@@ -430,6 +430,48 @@ gap> AutomorphismGroup(S);
...
+Error, the arguments <temp> and <table> must be square tables of the same dime\
+nsions
+gap> S := MultiplicationTable(SymmetricInverseMonoid(2));;
+gap> M := List([1 .. 7], x -> [1 .. 7]);;
+gap> p := (1, 4, 2, 5, 7);;
+gap> OnMultiplicationTable(S, p) = M;
+false
+gap> PermuteMultiplicationTable(M, S, p);
+gap> OnMultiplicationTable(S, p) = M;
+true
```
james-d-mitchell, line 475:
```suggestion
true

#
```

doc/isomorph.xml
```diff
@@ -11,16 +11,20 @@
 <#GAPDoc Label="OnMultiplicationTable">
 <ManSection>
   <Oper Name = "OnMultiplicationTable" Arg = "table, p"/>
-  <Returns>
-    A multiplication table.
-  </Returns>
+  <Oper Name = "PermuteMultiplicationTable" Arg = "temp, table, p"/>
```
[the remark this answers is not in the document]
pramothragavan: How about `output` and `table` so that it's more consistent with `OnMultiplicationTable`? Feels odd to also change `OnMultiplicationTable` to have variable name `input`

pramothragavan:
Not sure I've done exactly what you wanted with the assertions @james-d-mitchell but let me know

james-d-mitchell:
> Not sure I've done exactly what you wanted with the assertions @james-d-mitchell but let me know

Probably easiest for me to just do it, when you are done, please let me know and I'll add the assertions.

james-d-mitchell:
Ah you did exactly what I was suggesting already, thanks!

pramothragavan:
Assuming the CI failure is unrelated?

PR #1067, opened 2025-06-20
Replace custom type `GVarFunc` with GAP's `ObjFunc`

fingolfin:
This will improve compatibility with GCC 15 and future GAP versions.

For background see <https://github.com/gap-system/gap/pull/6010> and <https://github.com/gap-system/gap/issues/5857>

This is targeting `main` right now but I am happy to retarget it to your preferred branch.


fingolfin:
This is currently blocking https://github.com/gap-system/gap/pull/6010 -- it would be awesome to get this into a release "soon". Bonus points if it also contains  https://github.com/semigroups/Semigroups/pull/1068  -- but of course I fully understand that you just made a release and have other work to do as well.

PR #1087, opened 2025-09-02
Update to use `libsemigroups` v3

Joseph-Edwards:
This PR updates the package to use `libsemigroups` v3.

Presently, this is a draft and there is a lot more work to be done.


gap/congruences/congpart.gi
```diff
@@ -43,7 +43,7 @@ end);
 InstallMethod(SemigroupCongruenceByGeneratingPairs,
 "for a semigroup and a list",
 [IsSemigroup, IsList],
-19,  # to beat the library method for IsList and IsEmpty
+23,  # to beat the library method for IsList and IsEmpty
```
Joseph-Edwards: This now actually beats the library method. I just picked the first value that beats the library method, so it's possible there is a more sensible choice for this value.

This also raises the question: are we actually beating the library methods we claim to. This issue would suggest some of the old rankings might've changed. This is probably an issue for another time.

PR #1104, opened 2025-09-15
Add a Valgrind GitHub Actions workflow

wilfwilson:
This is basically a copy of the file in Digraphs: thank you whoever made that!

To see that the workflow functions as expected, see here: https://github.com/wilfwilson/Semigroups/actions/runs/17740854488/job/50414239468

Note that the job fails, because Valgrind does indeed find some issues...

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -23,7 +23,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       NO_COVERAGE: true
-      VALGRIND: valgrind --trace-children=yes --leak-check=full --show-leak-kinds=all --error-exitcode=1
+      VALGRIND: valgrind --trace-children=yes --leak-check=full --error-exitcode=1
     steps:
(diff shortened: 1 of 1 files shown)


james-d-mitchell:
I'm not seeing the valgrind job actually running, is that just me?

james-d-mitchell:
> Looks like it failed after running testinstall with acting methods off... with 2834 errors 😱 

Zoiks that is not a small number

wilfwilson:
Hopefully it's the same one error happening 2834 times :crossed_fingers:

james-d-mitchell:
I'll have a look on Wednesday

wilfwilson:
Your preferred option sounds good Joe

james-d-mitchell:
I was going to say, maybe it's only worthwhile running the tests in `tst/standard/libsemigroups` in this action, but those are absent from your list. @Joseph-Edwards

james-d-mitchell:
Thanks @wilfwilson and @Joseph-Edwards great work!

PR #1106, opened 2025-09-15
CI: use the token to upload coverage data to Codecov, and fail if the `codecov/codecov-action` action doesn't work

wilfwilson:
The code coverage hasn't been working properly for a while, but we didn't notice. This should fix things and make it so that we notice any future problems more quickly.


fingolfin (approved):
Looks good to me!

PR #1110, opened 2025-09-15
Update the CI workflows to use the latest versions of the `gap-actions` tools

wilfwilson:
~I've added pull requests #1113, #1114, #1115, and #1116, all of which contain commits from this pull request.~

~Once those are merged, I'll rebase this pull request and mark it as ready for review, at which point it should have a rather smaller diff.~

The main thing to note is that `gap-actions/setup-gap@v3`, unlike `v2`, no longer clones or builds the packages that we want to be cloned or built. So we have to do it ourselves now.

To see that the updated "Extended tests" workflow works properly (it won't run here since it isn't set to run on pull requests), see https://github.com/wilfwilson/Semigroups/actions/runs/17761920802

### DIFF
--- a/.github/workflows/config-options.yml
+++ b/.github/workflows/config-options.yml
@@ -22,16 +22,18 @@ jobs:
           update_packager_index: false
           override_cache_key: ${{ runner.os }}-$GAPBRANCH-64-${{ github.ref }}
           override_cache_key_fallback: ${{ runner.os }}-$GAPBRANCH-64
-      - name: "Install GAP and clone/compile necessary packages"
-        uses: gap-actions/setup-gap@v2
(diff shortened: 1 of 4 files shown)


wilfwilson:
I've added pull requests #1113, #1114, #1115, and #1116, all of which contain commits from this pull request.

Once those are merged, I'll rebase this pull request and mark it as ready for review, at which point it should have a rather smaller diff.

wilfwilson:
Actually, I'll mark this ready for review now. The only remaining pull request that I took out from this one is #1113, and the changes from that don't really get in the way of reviewing this one. (Thanks for merging the other pull requests @james-d-mitchell).

PR #1111, opened 2025-09-15
Merge the code for the Windows GitHub Actions job with that of the UNIX jobs

wilfwilson:
As of #1110, the code for the Windows job no longer really has any special cases, so it doesn't have to be kept separate any more.

~This PR requires #1110. So I won't mark this as ready for review until #1110 is merged, at which point I'll rebase and there will only be the final commit remaining.~

### DIFF
--- a/.github/workflows/standard.yml
+++ b/.github/workflows/standard.yml
@@ -33,9 +33,18 @@ jobs:
           - v4.14
           - v4.13
           - v4.12
+        configflags:
+          - "--disable-hpcombi"
+
+        include:
+          - os: Windows
+            gap-version: latest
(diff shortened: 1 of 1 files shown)


wilfwilson:
I've rebased this and it's now a single commit that is ready for review, and hopefully merging. There's a failing test but I think that's spurious; I'll restart it when I can.

james-d-mitchell:
Thanks @wilfwilson , any chance you could also merge the changes into main at some point, I tried to do this earlier today, but it seemed non-trivial

wilfwilson:
Yeah the conflicts arose because dependabot has been making changes to the CI in `main` while I've been doing my tinkering on `stable-5.5`. I'm working on it.

PR #1112, opened 2025-09-15
macOS CI: reinstate testing Semigroups on GAP 4.12

wilfwilson:
Whatever was the cause of us wanting to exclude this, doesn't seem to be an issue any longer.


wilfwilson:
Okay, that's not true. It must only be fixed with the latest GitHub Actions stuff, i.e. in #1110. So I'll keep this change in there.

PR #1166, opened 2026-04-01
Codex doc improvements

james-d-mitchell:


doc/attr.xml
```diff
@@ -1499,12 +1498,12 @@ gap> MinimalFaithfulTransformationDegree(S);
     minimal degree that is isomorphic to <A>S</A>. This is currently only
     implemented for a very small number of types of semigroups. <P/>
 
-    See also <Ref Attr="MinimalFaithfulTransformationRepresentation"/>.
+    See also <Ref Attr="MinimalFaithfulTransformationDegree"/>.
 
     <Example><![CDATA[
 gap> S := RightZeroSemigroup(10);
 gap> MinimalFaithfulTransformationRepresentation(S);
-7
```
Joseph-Edwards, line 1507: I've just checked this locally and I also get 7. Is there a reason for changing this to a TODO?

doc/conginv.xml
```diff
@@ -45,7 +45,7 @@ gap> TraceOfSemigroupCongruence(cong);
       An inverse semigroup.
     </Returns>
     <Description>
-      If <A>cong</A> is a congruence over a semigroup with inverse op,
+      If <A>cong</A> is a congruence over a semigroup with inverse operation,
```
Joseph-Edwards:
```suggestion
      If <A>cong</A> is a congruence over a semigroup with inverses,
```
or

```suggestion
      If <A>cong</A> is a congruence over a semigroup with an inverse operation,
```

doc/congrms.xml
```diff
@@ -140,7 +140,7 @@ gap> cong := RZMSCongruenceByLinkedTriple(S, N, colBlocks, rowBlocks);;
       <C>colBlocks</C>, and <A>rowClass</A> must be a positive integer
       corresponding to a position in the list <C>rowBlocks</C>.<P/>
 
-      If the arguments are valid, an <C>IsRMSCongruenceClassByLinkedTriple</C>
+      If the arguments are valid, a <C>IsRMSCongruenceClassByLinkedTriple</C>
```
Joseph-Edwards:
```suggestion
      If the arguments are valid, an <C>IsRMSCongruenceClassByLinkedTriple</C>
```

doc/congsemigraph.xml
```diff
@@ -55,11 +55,11 @@ false
       This function returns a semigroup congruence over the graph inverse
       semigroup <A>S</A> in the form of a Wang pair.<P/>
 
-      If <A>S</A> is a finite graph inverse semigroup <A>H</A> and <A>W</A> are
+      If <A>S</A> is a finite graph inverse semigroup, <A>H</A> and <A>W</A> are
```
Joseph-Edwards:
```suggestion
      If <A>S</A> is a finite graph inverse semigroup, and <A>H</A> and <A>W</A> are
```

doc/congsemigraph.xml
```diff
@@ -84,7 +84,7 @@ gap> cong := CongruenceByWangPair(S, [3, 4], []);
     <Description>
       This operation takes <A>cong</A>, a finite graph inverse semigroup
       congruence, and returns an object representing the same congruence, but
-      described as a congruence by Wang pairs: a pair of sets <A>H</A> and
+      described as a congruence by Wang pair: a pair of sets <A>H</A> and
```
Joseph-Edwards, line 87: I'm not sure that this one was an issue in the first instance, but I also don't think it has been changed to something incorrect.

doc/dual.xml
```diff
@@ -95,12 +95,12 @@ false]]></Example> </Description> </ManSection>
...
 <#GAPDoc Label="IsDualSemigroupRep">
   <ManSection>
-    <Filt Name = "IsDualSemigroupRep" Type = "Category" Arg="sgrp"/>
+    <Filt Name = "IsDualSemigroupRep" Type = "category" Arg="sgrp"/>
     <Returns>Returns <K>true</K> if <A>sgrp</A> lies in the category of
       dual semigroups.</Returns>
     <Description>
       Semigroups created using <Ref Func="DualSemigroup"/>
-      normally lie in this category. The exception is semigroups
+      normally lie in this category. The exception is for semigroups
```
Joseph-Edwards:
```suggestion
      normally lie in this category, with the exception of semigroups
```

doc/freeband.xml
```diff
@@ -152,13 +152,13 @@ true]]></Example>
       appearing in the word representing the element <A>x</A> of the free band.
       <P/>
 
-      The function <C>ContentOfFreeBandElement</C> returns the content of free
+      The function <C>ContentOfFreeBandElement</C> returns the content of a free
```
Joseph-Edwards:
```suggestion
      The function <C>ContentOfFreeBandElement</C> returns the content of the free
```

doc/semicons.xml
```diff
@@ -277,7 +277,7 @@ true]]></Example>
           <Item><C>IsTropicalMaxPlusMatrixSemigroup</C>,</Item>
           <Item><C>IsTropicalMinPlusMatrixSemigroup</C>,</Item>
           <Item><C>IsProjectiveMaxPlusMatrixSemigroup</C>,</Item>
-          <Item><C>IsIntegerMatrixSemigroup.</C></Item>
+          <Item><C>IsIntegerMatrixSemigroup</C>.</Item>
```
Joseph-Edwards, line 280: I can't see what is actually being changed here. I've checked and these two full stops are the same unicode character. Quite suspicious to me.

doc/semifp.xml
```diff
@@ -105,7 +105,7 @@ true]]></Example>
   <Func Name="FreeSemigroupAndAssignGeneratorVars" Arg="arg..."/>
   <Returns>A free semigroup or monoid.</Returns>
   <Description>
-    <C>FreeMonoidAndAssignGeneratorVars</C> is synonym with:
+    <C>FreeMonoidAndAssignGeneratorVars</C> is synonymous with:
```
Joseph-Edwards, line 108: I'm not sure if either synonymous or synonym are right here? Something like
``` 
A potential implementation of <C>FreeMonoidAndAssignGeneratorVars</C> is:
```
or something like that.

doc/utils.xml
```diff
@@ -47,10 +50,10 @@
...
     <Returns><K>true</K> or <K>false</K>.</Returns>
     <Description>
-      This function should be called with no argument to test your installation
-      of &SEMIGROUPS; is working correctly. These tests should take no more
-      than a few seconds to complete.  To more comprehensively test
-      that &SEMIGROUPS; is installed correctly use
+      This function should be called with no argument to test that your
+      installation of &SEMIGROUPS; is working correctly. These tests should
+      take no more than a few seconds to complete.  To more comprehensively
+      test that &SEMIGROUPS; is installed correctly use
```
Joseph-Edwards:
```suggestion
      test that &SEMIGROUPS; is installed correctly use
```
If the trailing whitespace can be removed here, then it should be.

doc/utils.xml
```diff
@@ -98,10 +102,10 @@
     <InfoClass Name = "InfoSemigroups"/>
     <Description>
     <C>InfoSemigroups</C> is the info class of the &SEMIGROUPS; package.  The
-    info level is initially set to 0 and no info messages are displayed.  To
-    increase the amount of information displayed during a computation increase
-    the info level to 2 or 3. To stop all info messages from being displayed,
-    set the info level to 0.  See also
+    info level is initially set to <C>0</C> and no info messages are displayed.
+    To increase the amount of information displayed during a computation increase
```
Joseph-Edwards:
```suggestion
    To increase the amount of information displayed during a computation increase
```

Joseph-Edwards (commented):
On the whole, I'm very impressed by Codex's performance here. I have a few suggestions that I have outlined below, but most of these are personal preferences rather than Codex being wrong.

doc/attr.xml
james-d-mitchell: There's no function `MinimalFaithfulTransformationRepresentation`, so you can't have reproduced this locally :) That's why.

doc/congsemigraph.xml
james-d-mitchell: It is represented by a single Wang pair, so I think this change is good.

doc/semicons.xml
james-d-mitchell: I guess it's just the punctuation `.` not being inside the `<C>` tags, which is consistent with the earlier lines.

doc/semifp.xml
james-d-mitchell: Good point will update.

james-d-mitchell:
> On the whole, I'm very impressed by Codex's performance here. I have a few suggestions that I have outlined below, but most of these are personal preferences rather than Codex being wrong.

Thanks @Joseph-Edwards. Just to mention this was Codex with me reviewing every suggested change one file at a time, not just letting it rip :)

PR #1136, opened 2026-01-21
Generalized Conjugacy Class

JosephDayngerRuiz:
This pull request will include the finite field case for computing Cartan matrices for finite monoids over finite fields.

PR #1175, opened 2026-04-22
`GeneralizedConjugacyClasses`

JosephDayngerRuiz:
Continuation of pull request #1136. To implement the generalized conjugacy class functionality of monoids in a similar way to conjugacy classes of groups in gap.


JosephDayngerRuiz:
Ready for review!

PR #1176, opened 2026-04-28
CI: use configflags instead of CONFIGFLAGS

fingolfin:
See https://github.com/gap-actions/build-pkg/pull/32

### DIFF
--- a/.github/workflows/config-options.yml
+++ b/.github/workflows/config-options.yml
@@ -38,7 +38,7 @@ jobs:
       - name: "Build Semigroups"
         uses: gap-actions/build-pkg@v3
         with:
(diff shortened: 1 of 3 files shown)


fingolfin:
@james-d-mitchell it seems most of the test failures are [also on `main`](https://github.com/semigroups/Semigroups/actions/runs/25037024514)? (And one is due to `Could not resolve host: github.com`)

james-d-mitchell:
Thanks for the PR @fingolfin, I'm happy to merge this as is, it looks like the test failures are occurring due to changes in GAP stable-4.15 and master branches, I'll investigate shortly.

PR #1179, opened 2026-05-09
Remove moduleState init from gapbind14 kernel extension

fingolfin:
Global variables are zero-initialized anyway, so this makes no
difference.

Note that module state only matters for HPC-GAP. In the future,
I'd like to hide the moduleStateSize and moduleStateOffsetPtr
for "regular" GAP builds.


fingolfin:
I can also retarget this at the stable-5.6 branch if you prefer

james-d-mitchell:
Thanks happy to have this in main, the ci failures are related to the issue reported by Reinis, and not this pr probably

PR #1182, opened 2026-05-16
Release candidate for v5.6.3

james-d-mitchell:
@fingolfin can you please confirm that if this release is made and merged into `main`, then this will allow GAP 4.16.0 to be released (at least as far as the Semigroups package is concerned of course). Thanks!


fingolfin:
@james-d-mitchell I tested this and to the best of my ability am convinced it will pass all PackageDistro tests and be fine. So let's say I am 99.9% confident.

Anyway: even if I am wrong and there is some further obscure issue, then this is not necessarily a blocker for the release; we could agree to ignore it and postpone a package distro fix for later; I can make a 4.16.1 release quickly as needed. My main concern is that I'd like to avoid a release of GAP that breaks a major package like semigroups. In that sense, I guess even Semigroups 5.6.2 really is fine, because the "breakage" mostly is one test of the test suite that is not important for actual usage of the package. It's just that I'd rather avoid having "semigroups" always fail in package distro tests, because it means that *actual* failures can slip in without being noticed.

tl;dr: releasing this as-is seems perfect, thank you so much, I'll take responsibility for dealing with any further issues

james-d-mitchell:
Thanks @fingolfin happy to make a release, and don't want to be a blocker for releasing GAP 4.16!

PR #1185, opened 2026-05-17
bipart: error when trying to multiply bipartitions of diff. degree

### LINKED ISSUE #1183: Multiplying bipartitons of distinct degree causes wrong result
Bonjour, earlier today I tried multiplying two bipartitions of distinct degree, just to see what happens:

```gap
gap> LoadPackage("Semigroups");;
gap> a := Bipartition([[1, -1, -2], [2, 5], [3, 4, -3], [-4, -5], [6, -6]]);
<bipartition: [ 1, -1, -2 ], [ 2, 5 ], [ 3, 4, -3 ], [ 6, -6 ], [ -4, -5 ]>
gap> b := Bipartition([[1, -4], [2], [3, -1], [4], [5, -3], [-2], [-5]]);
<bipartition: [ 1, -4 ], [ 2 ], [ 3, -1 ], [ 4 ], [ 5, -3 ], [ -2 ], [ -5 ]>
gap> a * b;
<bipartition: [ 1, -3, -5, -6 ], [ 2, 5 ], [ 3, 4, 6 ], [ -1 ], [ -2 ], [ -4 ]>
```

I was expecting one of two results:

1. An error gets thrown informing me that I am attempting something bad, or
2. The domain of the second bipartition `b` gets silently expanded to contain two singletons `[6]` and `[-6]`, and the multiplication proceeds as usual, resulting in `a * b = <bipartition: [ 1, -4 ], [ 2, 5 ], [ 3, 4, -1 ], [ -2 ], [ -3 ], [ -5 ], [6], [-6]>` or even just `<bipartition: [ 1, -4 ], [ 2, 5 ], [ 3, 4, -1 ], [ -2 ], [ -3 ], [ -5 ]>` (if e.g. Semigroups automatically trimmed singleton pairs `[n], [-n]`).
3. The domain of the second bipartition `b` gets silently expanded to contain the transversal `[6, -6]`, and the multiplication proceeds as usual, resulting in `a * b = <bipartition: [ 1, -4 ], [ 2, 5 ], [ 3, 4, -1 ], [ -2 ], [ -3 ], [ -5 ], [6, -6]>`.

But the actual answer `<bipartition: [ 1, -3, -5, -6 ], [ 2, 5 ], [ 3, 4, 6 ], [ -1 ], [ -2 ], [ -4 ]>` is definitely wrong - `b` does not touch the point `6` at all, so it should be impossible for `a * b` to have a non-trivial block involving `6`!

---

I'm not sure which of the behaviors 1., 2. or 3. is better for bipartitions. On one hand, the way GAP handles transformations is by automatically expanding their domains, which means that transformations of distinct degrees can be multiplied together. On the other hand, it is not clear how to expand the domain of a bipartition, we have a choice: either expand it with two singletons `[-n], [n]` or to expand with a transversal `[-n, n]`. I think adding the singleton `[-n], [n]` is a bit more logical than adding the transversal. So I think its between 1. and 2.

Likely the simplest thing to do would be to just do 1. and throw an error.

james-d-mitchell:
Fixes #1183

### DIFF
--- a/src/bipart.cpp
+++ b/src/bipart.cpp
@@ -273,6 +273,15 @@ Obj BIPART_PROD(Obj x, Obj y) {
   Bipartition* xx = bipart_get_cpp(x);
   Bipartition* yy = bipart_get_cpp(y);
 
+  if (xx->degree() != yy->degree()) {
+    // TODO implement a product_inplace checks version for bipartitions in
+    // Libsemigroups
+    ErrorMayQuit(
+        "the arguments <x> and <y> must have equal degrees but found %d <> %d",
+        xx->degree(),
+        yy->degree());
+  }
+
   Bipartition* z = new Bipartition(xx->degree());
   z->product_inplace_no_checks(*xx, *yy);
 
--- a/tst/testinstall.tst
+++ b/tst/testinstall.tst
@@ -12,7 +12,7 @@
 #@local bug, c, cong, contain, ec, es, f, file, g, gens, gns, hom, i, id, inv
 #@local iso, iter, latt, log, looking, lookingfor, map, mat, max, n, number, o
 #@local pair, pairs, r, regular, rel, rels, s, s1, s2, sgns, slist, small, t
-#@local tab, tuples, u, x, y, z
+#@local tab, tuples, u, x, y, z, a, b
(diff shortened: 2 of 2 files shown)


reiniscirpons (approved):
LGTM, thanks for the fix!

Issue #1191, opened 2026-05-18
libsemigroups requires C++17 standard


jswent:
recent versions of libsemigroups implement the C++17 standard, however the current readme (and presumably other documentation) says C++14

james-d-mitchell:
Thanks! @jswent

PR #1192, opened 2026-05-18
require c++17 standard for libsemigroups

jswent:
this pr aims to address issue #1191, however there are likely places I've missed since I'm not very familiar with this repo


james-d-mitchell:
Thanks @jswent much appreciated!

PR #1193, opened 2026-05-18
Adjust kernel extension code setting up StructInitInfo

fingolfin:
Currently GAP's `StructInitInfo` struct type contains several
obsolete fields which cannot be removed or replaced because of
kernel extensions relying on the specific layout of this struct.

This patch removes this reliance for this package and thus gets
us one step closer to eventually allowing such a refactoring of
the struct in the GAP kernel.

The technique employed here used to work only in C, but at least
locally it works for me in this C++ code base, too. Not sure if
that is due to enhancements in newer C++ standards, or a non-portable
compiler extension, or something else...

### DIFF
--- a/gapbind14/demo/src/gapbind_demo.cc
+++ b/gapbind14/demo/src/gapbind_demo.cc
@@ -119,18 +119,10 @@ static Int InitLibrary(StructInitInfo* module) {
 }
 
 static StructInitInfo module = {
-    /* type        = */ MODULE_DYNAMIC,
-    /* name        = */ "gapbind_demo",
-    /* revision_c  = */ 0,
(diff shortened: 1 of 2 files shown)


james-d-mitchell:
Seemed to pass the CI here too, so hopefully safe to merge this? Took the liberty of reformatting, thanks @fingolfin

fingolfin:
Happy about formatting changes, do as you like! It still failed, though. I've run `clang-format` locally again and pushed the result, but I don't think that made it happy -- perhaps due to different `clang-format` versions formatting different (mine is 22.1.3).

After installing `llvm@15` and using its `clang-format`, hopefully all is well now...

PR #1194, opened 2026-05-19
Update PackageInfo.g

alex-a-levine:
Updated Luna's name

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -142,7 +142,7 @@ Persons := [
 
   rec(
     LastName      := "Elliott",
-    FirstNames    := "Luke",
+    FirstNames    := "Luna",
     IsAuthor      := true,
     IsMaintainer  := false,
     Email         := "le27@st-andrews.ac.uk",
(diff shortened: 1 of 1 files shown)


PackageInfo.g
```diff
@@ -142,7 +142,7 @@ Persons := [
 
   rec(
     LastName      := "Elliott",
-    FirstNames    := "Luke",
+    FirstNames    := "Luna",
     IsAuthor      := true,
     IsMaintainer  := false,
     Email         := "le27@st-andrews.ac.uk",
```
james-d-mitchell: Any chance we can update Luna's email address too, pretty sure this one is no longer valid @alex-a-levine

fingolfin: and WWWHome below

reiniscirpons:
@le27 could you check if the proposed affiliation/email etc details for you are correct? You should be able to see the proposed changes in the "Files changed" tab above.

james-d-mitchell:
Thanks @reiniscirpons et al

PR #1198, opened 2026-06-16
Some minor Graph Inverse Semigroup related fixes/additions

joe-dw:
Set `IsPositionalObjectRep` filter on graph inverse semigroup elements, 

Added attribute `EdgesOfGraphInverseSemigroup`, which returns the generators of the semigroup which are not vertices,

Added `IsWholeFamily` test to subsemigroups, which returns true if the subsemigroup is equal to the graph inverse semigroup of the graph it is defined on.


james-d-mitchell:
Thanks @joe-dw !

PR #1200, opened 2026-06-19
Fix of MinimalHereditarySubsetsVertex filter

joe-dw:
`MinimalHereditarySubsetsVertex` should take an input of a digraph and integer:

`  if IsMultiDigraph(D) then
    ErrorNoReturn("the 1st argument (a digraph) must not have multiple edges"); `

But it was previously defined with the filter

`DeclareOperation("MinimalHereditarySubsetsVertex",
                 [IsGraphInverseSemigroup, IsPosInt]); `

I have changed this to require a digraph instead of a GIS

### DIFF
--- a/gap/congruences/congsemigraph.gd
+++ b/gap/congruences/congsemigraph.gd
@@ -23,6 +23,6 @@ DeclareOperation("CongruenceByWangPair",
 DeclareOperation("AsCongruenceByWangPair", [IsSemigroupCongruence]);
 
 DeclareOperation("MinimalHereditarySubsetsVertex",
-                 [IsGraphInverseSemigroup, IsPosInt]);
+                 [IsDigraph, IsPosInt]);
 
 DeclareAttribute("GeneratingCongruencesOfLattice", IsGraphInverseSemigroup);


james-d-mitchell:
As discussed, this is probably not the correct place to make this change. I suggest doing the following:

1. Move the declaration of`MinimalHereditarySubsetsVertex` to Digraphs
2. Copy the function `SEMIGROUPS_MinimalHereditarySubsetsVertex` as the installed method for `MinimalHereditarySubsetsVertex` in Digraphs
3. Modify the Semigroups gi file to look (something) like the following (but that actually works in case this does not):

~~~
if not IsBoundGlobal("MinimalHereditarySubsetsVertex") then
    BindGlobal("SEMIGROUPS_MinimalHereditarySubsetsVertex",
    function(D, v)
    local subsets, hereditary, u, out, s, a;
    if IsMultiDigraph(D) then
        ErrorNoReturn("the 1st argument (a digraph) must not have multiple edges");
    elif not (v in DigraphVertices(D)) then
        ErrorNoReturn("the 2nd argument (a pos. int.) is not a vertex of ",
                    "the 1st argument (a digraph)");
    fi;
    out := Set(OutNeighboursMutableCopy(D)[v]);
    subsets := [];
    for u in [1 .. Length(out)] do
        a := out[u];
        RemoveSet(out, a);
        hereditary := ShallowCopy(out);
        for s in out do
        UniteSet(hereditary, VerticesReachableFrom(D, s));
        od;
        if not (a in hereditary) and not (hereditary in subsets) then
        AddSet(subsets, hereditary);
        fi;
        AddSet(out, a);
    od;
    return AsSortedList(subsets);
    end);
else
    BindGlobal("SEMIGROUPS_MinimalHereditarySubsetsVertex",
    MinimalHereditarySubsetsVertex);
fi;
~~~

PR #1206, opened 2026-07-30
Make smallgrp an explicit dependency

fingolfin:
SEMIGROUPS_ProcessRandomArgsCons for IsReesMatrixSemigroup picks a random group when the caller does not supply one, using NumberSmallGroups and SmallGroup. Without the SmallGrp package this fails, e.g. in a GAP started with `--bare':

    gap> RandomSemigroup(IsReesMatrixSemigroup);;
    Error, Variable: 'NumberSmallGroups' must have an assigned value

This is the only place in Semigroups using the small groups library.

See https://github.com/gap-system/gap/issues/2434

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---

Of course you could also rework the code to not use SmallGrp, that'd be perfectly cromulent :-)

### DIFF
--- a/PackageInfo.g
+++ b/PackageInfo.g
@@ -424,7 +424,8 @@ Dependencies := rec(
                           ["genss", ">=1.6.5"],
                           ["images", ">=1.3.1"],
                           ["IO", ">=4.5.1"],
-                          ["orb", ">=4.8.2"]],  # TODO bump to v5.1.0
+                          ["orb", ">=4.8.2"],  # TODO bump to v5.1.0
+                          ["smallgrp", ">=1.0"]],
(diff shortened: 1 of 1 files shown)


fingolfin:
BTW the current function would also run into an error if it happened to produce `order` equal to one of 1024, 2016, 2024, 2025, 2040 as for those SmallGrp does not provide the groups. So the function is a chance of about 1 in 400 to run into an error.

Also, if the goal was to select small groups uniformly, that is of course also not the case. But I assume you are aware, just wanted to mention it for completeness sake :-).

james-d-mitchell:
Thanks @fingolfin !

PR #1213, opened 2026-09-02
Added `LeftOne` and `RightOne` for graph inverse semigroups

joe-dw:
These functions exist for transformations and partial perms, this adds the same functionality for graph inverse semigroups

### DIFF
--- a/doc/semigraph.xml
+++ b/doc/semigraph.xml
@@ -303,7 +303,7 @@ true
   <Description>
     <C>RightOne(<A>x</A>)</C> returns the unique idempotent
     <C>e</C> such that <C><A>x</A> * e = <A>x</A></C> and
-    <C>e</C> and <A>x</A> are related by Green's L-relation.
+    <C>e</C> and <A>x</A> are related by Green's &L;-relation.
     <Example><![CDATA[
 gap> S := GraphInverseSemigroup(ChainDigraph(3));
 <finite graph inverse semigroup with 3 vertices, 2 edges>
--- a/doc/semigraph.xml
+++ b/doc/semigraph.xml
@@ -281,7 +281,7 @@ gap> IndexOfVertexOfGraphInverseSemigroup(v_3);
   <Description>
     <C>LeftOne(<A>x</A>)</C> returns the unique idempotent
     <C>e</C> such that <C>e * <A>x</A> = <A>x</A></C> and
-    <C>e</C> and <A>x</A> are related by Green's R-relation.
+    <C>e</C> and <A>x</A> are related by Green's &R;-relation.
     <Example><![CDATA[
 gap> S := GraphInverseSemigroup(ChainDigraph(3));
 <finite graph inverse semigroup with 3 vertices, 2 edges>
--- a/doc/semigraph.xml
+++ b/doc/semigraph.xml
@@ -273,3 +273,49 @@ gap> IndexOfVertexOfGraphInverseSemigroup(v_3);
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="LeftOne">
+<ManSection>
+  <Oper Name = "LeftOne" Arg = "x" Label="for an element with an inverse"/>
+  <Returns>A semigroup element.</Returns>
+  <Description>
+    if <A>x</A> has an inverse element <C>x ^ -1</C>, then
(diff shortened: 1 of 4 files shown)


joe-dw:
@james-d-mitchell assuming the tests complete this looks ready to go

gap/semigroups/semigraph.gi
```diff
@@ -302,3 +302,13 @@ InstallMethod(IsWholeFamily,
 "for a subsemigroup of a graph inverse semigroup",
 [IsGraphInverseSubsemigroup],
 S -> Size(ElementsFamily(FamilyObj(S))!.semigroup) = Size(S));
+
+InstallMethod(LeftOne,
```
james-d-mitchell, line 306: This method is valid for all inverse semigroups, so why not implement it for IsInverseSemigroup, instead of just for graph inverse semigroups explicitly?

james-d-mitchell (changes requested):
Please implement for inverse semigroups in general, rather than just of graph inverse semigroups.

gap/semigroups/semigraph.gi
joe-dw: The issue is that there isn't an equivalent filter `IsInverseSemigroupElement` so I would have to write the function for a general semigroup element and then test if its inverse exists. There exist functions such as `HasInverse` but they don't appear to be universal for `IsInverseSemigroup`, for example `HasInverse(GraphInverseSemigroup(Digraph([ [] ])).1)` is false

joe-dw: I had previously implemented this and Michael suggested that it would be cleaner to only make the function for GIS

joe-dw: `Inverse` also doesn't exist for GIS, so maybe the solution is to implement these functions?

joe-dw:
@james-d-mitchell does the first commit look good? I can roll back to that from before we changed to GIS only (commit 1b7e83c446a8e5ac80a80b1052581df3041cce72)

gap/semigroups/semigraph.gi
james-d-mitchell: Good point! I feel like we could still implement a more general version for `IsMultiplicativeElementWithInverse` which would be something like:
~~~
inv := Inverse(x);
if inv != fail then
return x * inv;
fi;
~~~

`Inverse` should certainly work for gis elements, this is an oversight. Maybe it'd be better to fix that first, then make the change proposed here after? Happy to discuss further

doc/semigraph.xml
```diff
@@ -273,3 +273,47 @@ gap> IndexOfVertexOfGraphInverseSemigroup(v_3);
...
 <#/GAPDoc>
+
+<#GAPDoc Label="LeftOne">
+<ManSection>
+  <Oper Name = "LeftOne" Arg = "x" Label="for a graph inverse semigroup element"/>
+  <Returns>A graph inverse semigroup element.</Returns>
+  <Description>
+    <C>LeftOne(<A>x</A>)</C> returns the unique idempotent
+    <C>e</C> such that <C>e * <A>x</A> = <A>x</A></C> and
+    <C>e</C> and <A>x</A> are related by Green's R-relation.
```
james-d-mitchell:
```suggestion
    <C>e</C> and <A>x</A> are related by Green's &R;-relation.
```

doc/semigraph.xml
```diff
@@ -273,3 +273,47 @@ gap> IndexOfVertexOfGraphInverseSemigroup(v_3);
...
+<#/GAPDoc>
+
+<#GAPDoc Label="RightOne">
+<ManSection>
+  <Oper Name = "RightOne" Arg = "x" Label="for a graph inverse semigroup element"/>
+  <Returns>A graph inverse semigroup element.</Returns>
+  <Description>
+    <C>RightOne(<A>x</A>)</C> returns the unique idempotent
+    <C>e</C> such that <C><A>x</A> * e = <A>x</A></C> and
+    <C>e</C> and <A>x</A> are related by Green's L-relation.
```
james-d-mitchell:
```suggestion
    <C>e</C> and <A>x</A> are related by Green's &L;-relation.
```

james-d-mitchell (approved):
Thanks @joe-dw I think I've come full circle and am now happy to merge this PR subject to the two minor doc changes indicated (the `&L;` and `&R;` expand to nicer looking L and R in some formats of the doc).

Issue #1211, opened 2026-08-31
Graph inverse semigroups can be defined to be missing their zero element.


joe-dw:
All graph inverse semigroups have an adjoined zero element (https://doi.org/10.1016/j.jalgebra.2025.08.041, page 311), however if D has only one vertex, the current method does not include the multiplicative zero as an element.

```gap> G := GraphInverseSemigroup(Digraph([ [] ]));
<finite graph inverse semigroup with 1 vertex, 0 edges>
gap> Elements(G);
[ v_1 ]
gap> MultiplicativeZero(G);
0
gap> MultiplicativeZero(G) in Elements(G);
false
gap> MultiplicativeZero(G) in G;
true```

To fix this, I think the zero just needs adding to the generating set, and the `GraphInverseSemigroup` adjusting accordingly. I'll work on this once I finish my current GIS related PR

joe-dw:
Fixed by PR #1216

PR #1216, opened 2026-09-03
Adding the zero element to the generating set for graph inverse semigroups with at most one vertex.

joe-dw:
Fixing issue #1211


joe-dw:
There are some tests that currently fail as they expect the old behavior, where
`GraphInverseSemigroup(Digraph([ [] ]))`
is isomorphic to the trivial group

joe-dw:
@james-d-mitchell this now passes the standard tests on my end, AssignGeneratorVariables still works and warns the user when the generating set includes zero

gap/semigroups/semigraph.gi
```diff
@@ -66,15 +70,25 @@ function(graph)
...
+                                         [MultiplicativeZero(S)]));
+    Add(gens, MultiplicativeZero(S));
+  else
+    SetGeneratorsOfSemigroup(S,
                            Concatenation(gens,
                                          List([1 .. DigraphNrEdges(graph)],
                                               x -> gens[x] ^ -1)));
+  fi;
   SetGeneratorsOfInverseSemigroup(S, gens);
-  SetGraphOfGraphInverseSemigroup(S, graph);
```
james-d-mitchell, line 77: This'd be more compact as (a properly formatted, and working version of):

~~~
  SetGraphOfGraphInverseSemigroup(S, graph);
  inv_gens := [];
  for i in [1 .. DigraphNrVertices(graph) + DigraphNrEdges(graph)] do
    Add(inv_gens, Objectify(fam!.type, [[i], graph]));
  od;
  gens := Concatenation(inv_gens, List([1 .. DigraphNrEdges(graph)], x -> gens[x] ^ -1));
  if DigraphNrVertices(graph) < 2 then
     Add(gens, MultiplicativeZero(S));
     Add(inv_gens, MultiplicativeZero(S));
  fi;
  SetGeneratorsOfInverseSemigroup(inv_gens);
  SetGeneratorsOfSemigroup(gens);
~~~

if the order of the generators is not important, then you could conditionally add the zero to `inv_gens` before populating `gens`. I haven't checked that this actually works.

joe-dw: I've added it at the end as there are test cases that use the fact that, for any i less than the number of edges, 

`S.i = e_i`

And this way it shouldn't break anything

james-d-mitchell (changes requested):
Thanks @joe-dw this looks mostly good, a little refactoring required please. Also can you please add a comment to the code indicating why these changes are being made (something like "if the number of vertices is 1 or 0, then the multiplicative zero element of a gis is not a product of any vertices or edges, and so must be added to both generating sets).

gap/semigroups/semigraph.gi
james-d-mitchell: I think that makes sense

joe-dw: Will implement your changes and add some comments

james-d-mitchell:
Nice one thanks @joe-dw

PR #1204, opened 2026-07-24
Added some functions for working with graph inverse semigroups and their congruences

joe-dw:
Functions have been added to find edges and paths with range or source of a given vertex in a graph inverse semigroup, as well as finding the positive and negative paths that form any given element. There is also a new definition for the trace of a congruence by Wang pair, and graph inverse semigroup specific implementations of CongruenceTestMembershipNC, ImagesElm and EquivalenceRelationPartition for both traces and congruences.

PR #1217, opened 2026-09-03
Add `EdgesWithRange/Source`, `PathsWithRange/Source`, and `Positive/NegativePath` for graph inverse semigroups

joe-dw:
This is a smaller version of PR #1204 containing only a few functions and the related documentation and tests.


doc/semigraph.xml
```diff
@@ -317,3 +317,94 @@ true
...
+Expand commentComment on line R318Resolved
+<ManSection>
+  <Attr Name="EdgesWithSource" Arg="x"/>
+  <Returns>A list of graph inverse semigroup elements.</Returns>
+  <Description>
+    If <A>x</A> is a vertex of a graph inverse semigroup (i.e.
+    <C>IsVertex(<A>x</A>) = true</C>), then this attribute returns a list of
+    all edges <C>e</C> such that <C>Source(e)=<A>x</A></C>. If <A>x</A> is not
+    a vertex, then it returns a list of all edges with source
+    <C>Range(<A>x</A>)</C>.
```
james-d-mitchell: `EdgesWithSource` returns a list of all edges with source equal to `Range(x)` while `EdgesWithRange` returns a list of all edges with range equal to `Source(x)`, is this really what was intended? There might be a good reason for this that I'm not thinking of, but it'd be easier to remember if `EdgesWithSource` returned a list of all edges with source equal to `Source(x)` (source-source-source), and range-range-range for `EdgesWithRange`?

doc/z-chap07.xml
```diff
@@ -250,6 +250,10 @@
     <#Include Label = "IndexOfVertexOfGraphInverseSemigroup">
     <#Include Label = "LeftOne">
     <#Include Label = "RightOne">
+    <#Include Label = "EdgesWithRange">
+    <#Include Label = "EdgesWithSource">
+    <#Include Label = "PathsWithRange">
+    <#Include Label = "PathsWithSource">
```
james-d-mitchell, line 256: Where's the doc for `Positive/Negative` path?

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+InstallMethod(PositivePath,
+"for a graph inverse semigroup element",
+[IsGraphInverseSemigroupElement],
+function(elt)
+  local pos, S;
+
+  pos := PositionProperty(elt![1], IsNegInt);
+  if pos = fail then
+    return elt;
+  elif pos = 1 then
```
james-d-mitchell, line 344: Can you please add a comment about what this case corresponds to?

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+
+InstallMethod(NegativePath,
+"for a graph inverse semigroup element",
+[IsGraphInverseSemigroupElement],
+elt -> PositivePath(elt ^ -1) ^ -1);
+
+InstallMethod(IsIsomorphicSemigroup,
+"for two graph inverse semigroups",
+[IsGraphInverseSemigroup, IsGraphInverseSemigroup],
+43,
```
james-d-mitchell: I don't think this rank increment should be needed, if it is, then there should be a comment here saying why.

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+  for e in EdgesWithSource(Range(x)) do
+    Append(outPaths, List(PathsWithSource(Range(e)), p -> x * e * p));
+  od;
+  return outPaths;
+end);
+
+InstallMethod(Idempotents,
+"for a graph inverse semigroup",
+[IsGraphInverseSemigroup],
+43,
```
james-d-mitchell: Same comment, either not needed or comment explaining why.

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+function(G)
+  local idemps, v;
+  if not IsFinite(G) then
+    Error("the argument (a semigroup) is not finite");
```
james-d-mitchell:
```suggestion
    Error("the argument (a graph inverse semigroup) is not finite");
```

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+  GraphOfGraphInverseSemigroup(G2)));
+
+InstallMethod(EdgesWithRange,
+"for a graph inverse semigroup element",
+[IsGraphInverseSemigroupElement],
+function(x)
+  local G;
+  G := FamilyObj(x)!.semigroup;
+  if not IsFinite(G) then
+    ErrorNoReturn("the semigroup containing the argument must be finite");
```
james-d-mitchell:
```suggestion
    ErrorNoReturn("the graph inverse semigroup containing the argument (element) must be finite");
```

james-d-mitchell: Similar comments for your other error messages, be as specific as possible, and follow the pattern (from `CONTRIBUTING.md`:

> * error messages should give enough information for the user to know what they need
  to change. For example, if a function has received the wrong input type, an
  appropriate error to return is: `ErrorNoReturn("expected [something] as 2nd
  argument, found [something else]")`.

doc/semigraph.xml
```diff
@@ -317,3 +317,94 @@ true
   </Description>
 </ManSection>
 <#/GAPDoc>
+
+<#GAPDoc Label="EdgesWithRange">
+<ManSection>
+  <Attr Name="EdgesWithRange" Arg="x"/>
+  <Returns>A list of graph inverse semigroup elements.</Returns>
+  <Description>
```
james-d-mitchell, line 325: The doc doesn't describe what happens when `x` is the zero element? It looks like there's an error thrown by something that your methods call that says something like "the zero isn't allowed", this should be explicitly documented, and commented in the code. This goes for all the functions here.

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+elt -> PositivePath(elt ^ -1) ^ -1);
+
+InstallMethod(IsIsomorphicSemigroup,
+"for two graph inverse semigroups",
+[IsGraphInverseSemigroup, IsGraphInverseSemigroup],
+43,
+  # two graph inverse semigroups are isomorphic if and only if their underlying
+  # graphs are isomorphic
+{G1, G2} -> IsIsomorphicDigraph(GraphOfGraphInverseSemigroup(G1),
+  GraphOfGraphInverseSemigroup(G2)));
```
james-d-mitchell, line 365: For future PRs, it'd be best to subdivide things a bit further, i.e. including the `EdgesWithRange/Source`, `PathsWithRange/Source` and `Positive/NegativePath` in one PR is okay, but `IsIsomorphicSemigroup`
would fit better in a separate PR. I'm not saying change this PR now it's made, just for the future. Also what about `IsomorphismSemigroup`? If I want an explicit map from G1 to G2? How do I get that?

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+  for v in VerticesOfGraphInverseSemigroup(G) do
+    Append(idemps, List(PathsWithRange(v), p -> p ^ -1));
+  od;
+  return idemps;
+end);
+
+InstallMethod(IdempotentGeneratedSubsemigroup,
+"for a graph inverse semigroup",
+[IsGraphInverseSemigroup],
+43,
```
james-d-mitchell: Again this shouldn't be necessary, or if it is, then there should be a comment explaining it.

gap/semigroups/semigraph.gi
```diff
@@ -312,3 +312,136 @@ InstallMethod(RightOne,
...
+function(G)
+  local counter, v;
+  if not IsFinite(G) then
+    Error("the argument (a semigroup) is not finite");
+  fi;
+  counter := 1;
+  for v in VerticesOfGraphInverseSemigroup(G) do
+    counter := counter + Length(PathsWithRange(v));
+  od;
+  return counter;
```
james-d-mitchell:
```suggestion
return Sum(VerticesOfGraphInverseSemigroup(G), v -> Length(PathsWithRange(v)));
```

tst/standard/semigroups/semigraph.tst
```diff
@@ -59,6 +58,68 @@ gap> List(x, Range);
 [ v_4, v_4, v_2, v_1, v_2, v_1, v_1, v_2, v_1, v_3, v_4, v_2, v_1, v_5, v_4, 
   v_2, v_1, v_3, v_4, v_2, v_1, v_5, v_4, v_2, v_1, v_3, v_4, v_2, v_1, v_5, 
   v_4, v_2, v_1, v_1, v_2, v_3, v_4, v_5 ]
+gap> List(x, PositivePath);
```
james-d-mitchell, line 61: There should be tests for the errors also, and for the case when the argument is the zero element.

james-d-mitchell (changes requested):
Thanks @joe-dw, this is better, but still needs some improvements, and corrections. Please also try to have meaningful names for PRs in future, I edited this PR title to be more accurate, and possibly meaningful for future readers.

doc/semigraph.xml
joe-dw: This is because I want all edges `e` such that `ex <> 0`, which holds if `Range(e) = Source(x)`, so this is intended.

If you think it would be clearer I can make it so it returns nothing for non-vertex elements? I don't think I ever use it for anything other than vertices, and just thought it was worth adding at some point.

gap/semigroups/semigraph.gi
joe-dw: The semigroup isomorphism may be more complicated, but I can look into writing that. The main issue would be converting the isomorphism between digraphs to convert edges but this is almost certainly doable

gap/semigroups/semigraph.gi
joe-dw: Needs one added to the sum for the zero element but otherwise I've changed this as suggested

doc/semigraph.xml
joe-dw: I've implemented this change such that if not `IsVertex(x)` then the returned list is empty. I think the simplicity of " `EdgesWithRange(x)` returns all edges `e` such that `Range(e) = x` " is the clearest way to implement the function

gap/semigroups/semigraph.gi
joe-dw: Having looked further into this, these methods don't really make sense until PR #1216 is merged, since as long as issue #1211 remains, they would imply

```
G := GraphInverseSemigroup(Digraph([ [] ]));;
NrIdempotents(G) > Size(G);
true
```

I'll remove them from this PR for now, and once that's been merged I'll add them to another one, which also means this PR is just the paths/edges functions and addresses your other comment

joe-dw: (This is the reason the tests failed on this PR)

gap/semigroups/semigraph.gi
james-d-mitchell: The isomorphism of the graphs will give you a mapping of the vertices to the vertices, if that can be lifted to the edges in the generating set, then you could use `SemigroupIsomorphismByImages` to create the isomorphism itself.

gap/attributes/homomorph.gi
```diff
@@ -425,6 +425,20 @@ function(hom, x)
   return preim;
 end);
 
+# GAP >= 4.17 has a library method for PreImagesSet that returns the preimage
```
james-d-mitchell, line 428: The changes in this file shouldn't be here, they're already merged into main.

gap/semigroups/semigraph.gi
joe-dw: Would you prefer I add IsomorphismSemigroup to this PR? Working on it now but I can make a later PR with it too

james-d-mitchell: Later is probably better, thanks @joe-dw

joe-dw:
@james-d-mitchell

joe-dw:
I'm going to change how EdgesWithRange/Source works actually, because there's no reason it can't work with infinite but finitely generated semigroups. Only PathsWithRange/Source
 would need to fail. Will edit this and write some test cases

doc/semigraph.xml
```diff
@@ -317,3 +317,132 @@ true
...
+<#GAPDoc Label="PathsWithRange">
+<ManSection>
+  <Attr Name="PathsWithRange" Arg="x"/>
+  <Returns>A list of graph inverse semigroup elements.</Returns>
+  <Description>
+    If <A>x</A> is an element of a finite graph inverse semigroup (i.e.
+    it satisfies <C>IsGraphInverseSemigroupElement</C>), then this function
+    returns a list of all paths (products of edges) <C>p</C> such that 
+    <C>Range(p)=<A>x</A></C>, including <A>x</A> if it is a vertex.
+    Otherwise it returns an empty list.
```
james-d-mitchell: This still doesn't look right to me, it currently reads as "If x is an element of a finite graph inverse semigroup, then .... If x is not an element of finite graph inverse semigroup, then an empty list is returned." So, if I do `PathsWithRange(fail)` I should get the empty list? Same comment applies elsewhere too.

gap/semigroups/semigraph.gi
```diff
@@ -327,3 +327,109 @@ InstallMethod(RightOne,
...
+function(x)
+  local G;
+  G := FamilyObj(x)!.semigroup;
+  if not IsVertex(x) then
+    return [];
+  elif not IsFinite(EdgesOfGraphInverseSemigroup(G)) then
+    ErrorNoReturn("the graph inverse semigroup containing the argument",
+    "(element) must have a finite number of edges!");
+  fi;
+  return Filtered(EdgesOfGraphInverseSemigroup(G), e -> Range(e) = x);
```
james-d-mitchell: I think you can probably do this with constant complexity rather than with complexity equal to the number of edges in the graph, by using the underlying graph and it's `OutNeighbours` and `InNeighbours` which might be preferable to this.

gap/semigroups/semigraph.gi
```diff
@@ -327,3 +327,109 @@ InstallMethod(RightOne,
...
+  G := FamilyObj(x)!.semigroup;
+  if not IsFinite(G) then
+    ErrorNoReturn("the graph inverse semigroup containing the argument",
+    "(element) must be finite");
+  elif not IsVertex(x) then
+    return [];
+  fi;
+  outPaths := [x];
+  for e in EdgesWithSource(x) do
+    Append(outPaths, List(PathsWithSource(Range(e)), p -> x * e * p));
```
james-d-mitchell: This looks good, although if there are paths of length longer than 5000 this will likely trigger a recursion depth trap error. I'm not sure that it's worth updating this to avoid it, just commenting.

gap/semigroups/semigraph.gi
joe-dw: The issue is that each edge of a digraph is only stored as its source and range, so there isn't a convenient way to go from an edge in the digraph to an edge in the GIS. I think the best way to do this might be to put the generators into lists with the same source and same range, and then assign the attributes all at once

doc/semigraph.xml
joe-dw: I've changed this to remove the mention of the empty list, as if x is not a vertex then there are no paths with Range(p) = x and therefore the statement is trivial. There is no mention of what happens when the argument is not a GIS element, which is in line with other definitions, for example:

```
If  E is a digraph (i.e. it satisfies IsDigraph (Digraphs: IsDigraph)), then 
GraphInverseSemigroup returns the graph inverse semigroup G(E) where, 
roughly speaking, elements correspond to paths in the graph E.
...
```

gap/semigroups/semigraph.gi
joe-dw: It is possible to construct examples that could run in reasonable time but fail due to the recursion, most obviously the chain digraph. It seems like it might not matter but I'll rewrite it to use iteration instead of recursion.

PR #1219, opened 2026-09-06
Add PreImagesSet method for semigroup homs by images

fingolfin:
GAP 4.17 changes PreImagesSet: a generic method now checks that the argument is a subset of the range and returns the preimage of its intersection with the image. With that method, PreImagesSet no longer signals an error for elements without preimages, which the tests expect, and the error for elements outside the range changes.

Install a method for semigroup homomorphisms by images that performs the range check with the package's own error message and otherwise delegates to PreImagesElm, so behaviour is the same with all supported GAP versions.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>

See also https://github.com/gap-system/PackageDistro/issues/1550

### DIFF
--- a/gap/attributes/homomorph.gi
+++ b/gap/attributes/homomorph.gi
@@ -425,6 +425,20 @@ function(hom, x)
   return preim;
 end);
 
+# GAP >= 4.17 has a library method for PreImagesSet that returns the preimage
+# of the intersection with the image; this method keeps erroring for elements
+# without preimages, consistent with the PreImagesElm method above.
+InstallMethod(PreImagesSet,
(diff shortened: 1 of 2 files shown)


james-d-mitchell (approved):

james-d-mitchell:
Thanks @fingolfin

fingolfin:
Thanks for the quick merge, @james-d-mitchell . Would be lovely to have this in a release in September, so that we can gear up to GAP 4.17.0 in early October (right after GAP Days)

james-d-mitchell:
Thanks @fingolfin I'll try to make a release later this week, it has been on my list of things to do for a while now! Thanks again!

